Issue #111, opened 2020-08-09
Click Machine lacks a recipie


TheAdamizer:
**Modpack version**
0.2.5 + Rats

**Issue**
Click machine has no recipe

Maybe this is intentional (but I couldn't see any commits indicating as such) but I noticed Click Machine was added but there isn't currently a recipe for it.  After some investigation, it looks like there is a vanilla recipe that isn't working correctly.

NillerMedDild:
We've found the cause of this issue and are working to resolve it, thanks for letting us know anyhow :)

TheAdamizer:
Not sure where you are on your investigation, but I tried to debug the missing recipe a bit:  I'm able to reproduce the issue simply installing kubejs alongside click machine in an otherwise vanilla forge instance.  Debug log doesn't seem to indicate anything wrong with how ClickMachine is starting up outside of that.

PR #112, opened 2020-08-09
Adds back default recipe for auto clicker

TheAdamizer:
As per #111, this just adds the default recipe back for the ClickMachine Auto Clicker (which to me seems appropriate considering the rather end-game nature of click automation).

### DIFF
--- a/kubejs/data/enigmatica/kubejs/base/recipes/shaped.js
+++ b/kubejs/data/enigmatica/kubejs/base/recipes/shaped.js
@@ -41,7 +41,12 @@ events.listen('recipes', function (event) {
             L: 'mekanism:ultimate_bin',
             C: '#forge:ingots/compressed_iron',
             B: '#forge:glass'
-        })
+        }),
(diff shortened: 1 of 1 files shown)


NillerMedDild:
I would rather fix the issue causing the recipe not to load than re-add it, but thank you :) I'll merge if the solution isn't found within a reasonable time period.

TheAdamizer:
@NillerMedDild Sounds good thanks for the quick look!

ben-mkiv:
if it's related to their static recipe, i've made an PR for JSON recipe
https://github.com/Shadows-of-Fire/ClickMachine/pull/22

TheAdamizer:
It seems to be indeed caused by the static recipe.  I built your branch from the linked PR and ran it alongside KubeJS and the default recipe is now seen without needing to explicitly add it back using kubejs.

NillerMedDild:
The mod dev fixed the issue, by making a change in his library mod :) Thanks anyway!

PR #167, opened 2020-08-17
Linux scripts

brckts:
I took the liberty of translating the windows specific scripts to POSIX shell. 

They _might_ have a bug or two but should be generally ok.


NillerMedDild:
Thanks for your contribution! Powershell is no longer Windows specific, Powershell Core works on Mac and Linux :)

brckts:
Oh true, well, at least I learned something. Closing this then.

PR #340, opened 2020-10-29
Added tags for all the different tiers of powah items to be able to use them in quests

Merlo17:
I hope the filepath is right, i took what we had in E5 since E6 didn't have tags yet apparently, if it isn't I can change that later


NillerMedDild:
KubeJS switched folder structure recently, tags are now in `kubejs\server_scripts\enigmatica\kubejs\base\tags` :)

PR #348, opened 2020-11-02
Updated Modded for Dummies with new chapters and updated info

Merlo17:
Added a new FTB Chunks Map chapter, removed the Xaero's Map chapter, added an Ultimine entry and edited the quests entry. I wasn't actually able to test this on E6, just on a test instance, but I'm pretty sure this will work on E6 too. I was having a weird FTBQ issue when trying to create a new world to test it out in E6, but I'll let you do the testing


NillerMedDild (approved):

NillerMedDild:
Thanks Merlo! :D Great work.

PR #349, opened 2020-11-02
Nature's Aura Ore Generator Cleanup

MuteTiefling:
Removed mod specific ores so only EE ores spawn.

Added tags to nether gold to allow the nether version of ore generator to spawn it.

Added nether gold ore to the nether drop table.
Added Ancient Debris to the drop table (very rare)

For reference with the spawn weights, by default coal is 5000, diamond is 50, and emerald is 30. Putting Debris at 1 therefore makes it every rare.  Nether Gold at 1000 makes it relatively rare compared to quartz (default 3000)


NillerMedDild:
You rock, gj :D 
I'm not entirely sure on the tag `'forge:ores/nether/gold'`, but we can always change that down the line.

MuteTiefling:
I took the tag layout from pre-existing tags in Nature's Aura. Dunno if forge has settled on something but I figured I'd at least try to be consistent with NA itself :D

PR #360, opened 2020-11-07
Missed ores for IF and NA ore generators.

MuteTiefling:
Missed a few with the last update.
1) Arcane ore added to both. Similar rarity to Gold.
2) Bitumen added to both. Similar to silver rarity.
3) Fluorite added to both. Similar to diamond rarity.
4) Potassium Nitrate added to both. Similar to gold rarity.

Also re-pushing the prior changes that got merged to Master instead. Double checking this is going to develop! :D


NillerMedDild:
Awesome Motoko, thanks :D

PR #365, opened 2020-11-08
Optimise unification scripts for Rhino

MaxNeedsSnacks:
This mostly gets rid of the Java stream calls in `getPreferredItemInTag` and slightly optimises the `compareIndices` method so it doesn't loop through the list twice like it currently does.

Should be fine but may need additional testing (worked in my worlds without errors)

Also a general TODO: Start using ES6 syntax where possible, minimise pollution of the global scope by wrapping things in wrapper objects where possible, and stop using Java code or utils.listOf if we can avoid it (since we don't NEED to use it in most places anymore)

(Edit: force-pushed because the commit wasn't verified for some reason, plus I messed up some string formatting)

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/constants.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/constants.js
@@ -113,7 +113,8 @@ var modPriorities = [
     'botania',
     'quark',
     'pedestals',
-    'refinedstorage'
+    'refinedstorage',
+    'mapperbase'
 ];
 
 var vanillaWoodTypes = ['oak', 'birch', 'spruce', 'jungle', 'acacia', 'dark_oak'];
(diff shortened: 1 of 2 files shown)


NillerMedDild:
Thanks Max! Did you meassure the performance difference for this?

> minimise pollution of the global scope by wrapping things in wrapper objects where possible

Why? :D

MaxNeedsSnacks:
Realistically, the impact should be minimal, but it should be a bit easier to maintain and use in practice in case we need to change anything (also runtime for compareIndices is theoretically halved but that obviously only amounts to fractions of a second usually)

As for polluting the global scope, that’s really just a best practice we don’t have to do, it’s really just useful if you somehow reuse variable names across files and want to avoid accidentally overwriting things in other scripts ^^

NillerMedDild:
Ah that makes sense, I didn't know what global scope pollution was :P

MaxNeedsSnacks:
If you don't mind, I'd like to just go over the other scripts as well once / if this is merged, to get them all over to ES6 / Rhino standards ^^

NillerMedDild:
Yes please, I'd like to see what the best practice for various things actually is :)

PR #370, opened 2020-11-10
Axe and Hoe updates

MuteTiefling:
All axes and hoes should now be added to their respective tags.

I also took the liberty of adding all AIOTs and Paxels to the Axe tag.
AIOTs also function as hoes, so I've added them to the Hoe tag as well.

If either of these are undesired, all of the items are grouped up at the bottom of the list


NillerMedDild (approved):

NillerMedDild:
Thanks :D I'll go ahead and add AIOTs to the hoes list as well.

PR #374, opened 2020-11-13
Initialize RandomCobbleGen Config

MuteTiefling:
default config has ores and things generating from cobble gen. Neat mechanic, but maybe not what is being sought here.

Defaulted cobble/stone/basalt back to vanilla standards.
Added custom generators for stones, cobblestones, and sandstones from the following mods:

minecraft
embellishcraft
byg
biomesoplenty
good_nights_sleep
quark


MuteTiefling:
It would probably be easier to maintain as data packs or kubeJS scripts, but I couldn't find any way to actually do the custom generators that way and I wanted to get this up there.

NillerMedDild:
Don't worry, I wasn't planning on keeping the default config :D But thank you very much for doing all of the work!

PR #377, opened 2020-11-14
Tag Additions

MuteTiefling:
Alright, I admit it may be a bit much. But here's my thought process here... I've been playing modded for a long time and I still have trouble remembering the various names people give to stuff. Bins, chests, dsu, quantum... etc. Something that I've always wanted was an easy way to search for these things with a common term and now Tags show up. Perfect!

Tag all the things!

First pass here is armor by slot, the rest of the tools with a special multitool category, weapons (ranged, melee), and storage with multiple categories (basic, bulk, liquids, gases, bags, and energy).

I don't mind doing more and I feel these line up nicely with the quests sections. I thought that we could make mention of the tags in the various sections so that a player can easily search in JEI, for instance $lighting now brings up a nice list of useful light sources.  Need a new weapon? $melee


Darkere:
From what I see in the tags the naming is forge:<plural> like forge:coals but for specific things like forge:ores/coal it's singular for those. 

Additionally, I would prefer forge:lights over forge:lighting. Not my call though

MuteTiefling:
We already have forge:axes and forge:hoes so I just tried to keep with that structure. Easy to change though. 

I was a little conflicted on lights and basic storage tags, honestly. Forge:lights would be fine as well.

NillerMedDild:
I think this is a good addition, but I fear the overhead of having to maintain these lists to be honest.

NillerMedDild:
Three notes before I merge:
- You can add arrays of strings directly to tags, no need to iterate
- Files that add to a tag like `forge:armor/boots` should preferably by in the file `tags/forge/armor/boots.js`, unless it's just 2-5 entries.
- Great work as always :D

MuteTiefling:
alright, lemme re-work them a bit

NillerMedDild (approved):

NillerMedDild (approved):

PR #382, opened 2020-11-17
PNC config refresh

MuteTiefling:
Added  initial vacuum trap blacklist to prevent bees from being easily cloned with the new mob spawner.

Also reset the oil world-gen blacklist as it appears to be using an old default that didn't have the new nether and end biomes in it.

I know BYG adds some biomes to both of these too, but I couldn't find a list right now.


NillerMedDild (approved):

NillerMedDild:
Thanks!

PR #388, opened 2020-11-19
More Insolator trees

MuteTiefling:
Didn't check BoP and BYG since there seemed to be auto compat already. Adding the few missing trees I can find.


MuteTiefling:
Sorry for the spam! That should be it now

NillerMedDild (approved):

NillerMedDild:
I like the spam 😄 Thanks!

PR #397, opened 2020-11-21
JEI Hide - first pass

MuteTiefling:
Took a little different approach here, rather than a big long array.

I noticed that the JEI hide list completely fails if any of the entries that are put in are invalid, whether because they are mispelled or are simply non-existent. So I'm checking if each entry exists to avoid breakages if mods change in the future. That also meant I could take a lazier approach to most of the exclusions, building a list of materials and a list of types that get put together and pushed into the array.

It is slightly slower to load this way (13 ms as opposed to 2-3 ms) but it should hopefully be easier to maintain as well. I've still got a list of all of these items in array form as well if you'd prefer to revert.


NillerMedDild:
Thanks! Pretty nice approach :D I'm thinking we may be able to make it a bit more generic, albeit slower too. Glad you figured out how to check if an item exists, I've needed that for some time.

MuteTiefling:
Thank Max for the check, he's the one that helps me :D

PR #399, opened 2020-11-22
Tetra Replacements

Ridanisaurus:
- Updated current files
- Added Bows
- Added Shields (Lapis doesn't work as it is not a valid material for shield at the moment)


NillerMedDild:
Good work, thanks!

PR #403, opened 2020-11-23
Remove carriage returns from setup.sh

Infinoid:
I get these errors when I follow the Linux installation instructions:

```
$ bash setup.sh
setup.sh: line 2: $'\r': command not found
setup.sh: line 3: cd: $'..\r': No such file or directory
setup.sh: line 4: $'\r': command not found
setup.sh: line 7: $'\r': command not found
```

This patch fixes those by changing CRLF line endings to LF.

### DIFF
--- a/development/setup.sh
+++ b/development/setup.sh
@@ -8,4 +8,4 @@ echo "java -jar InstanceSync.jar" >> .git/hooks/post-merge
 echo "Done setting up hooks"
 echo "Running InstanceSync"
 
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Thanks for the PR, but it seems to hardcode the java path, so I'll just set EOL to LF myself :)

Infinoid:
You're right, I meant to put that in a second (local) commit but it snuck through.  Thanks for fixing it.

PR #410, opened 2020-11-23
EE Arcane Gem for Enchanting Apparatus

Ridanisaurus:
Added additional recipes for all the items/blocks which required Ars Nouveau's Mana Gem using EE Arcane Gem.
Sadly, these are hardcoded, so it didn't pick it up automatically, and they don't have their recipes as part of their API so we gonna have duplicated 10 recipes, but at least players can now progress through it.


NillerMedDild (approved):

NillerMedDild:
Thanks :D

Issue #429, opened 2020-11-27
Simple Farming seeds don't work with IE machines


Lykrast:
**Modpack version**
0.3.0

**Issue**
The Industrial Squeezer and Garden Cloche from IE do not have recipes with the Simple Farming seeds (Plant Oil and farming). The Farmer's Delight seeds are properly supported by both. The similar PnC Thermopneumatic Processing Plant and Botany Pots seem to support the seeds.

NillerMedDild:
Thank you for the report, I will investigate the issue.

Figulus88:
Simple Farming crops, seeds, and foods also lack recipes for Mekanism's Crusher for making Bio Fuel.

Same with Farmer's Delight crops, seeds, and foods.

They all seem to have recipes for Mek's Nutritional Liquifier though.

MuteTiefling:
I'll take a look at the garden cloche tonight. Looks like it supports datapacks now, so that's a major plus.

PR #467, opened 2020-12-03
Added Compat for Bio Fuel

MuteTiefling:
Added tagged inputs to Crusher for Biofuel

Tagged a few items that were missing tags so they're auto picked up by these and other recipes.

Added a bunch of misc items that I didn't feel could be easily tagged, or their tags were too broad to fit with the general theming of the input/output ratios.

#429


MuteTiefling:
Intentionally left out food. In general, the mekanism crushing recipes are a net loss if you craft up. For instance, cake gives you 8 bio fuel, but uses 3 wheat (5 bio fuel each) to make. Since they're mostly pointless/recycling recipes, I felt it better to stick to crops/seeds.

NillerMedDild (approved):
Awesome work!

PR #475, opened 2020-12-04
Cloche Compat

MuteTiefling:
Addition of Simple Farming crops to the cloche

Also added mushrooms from byg and quark. Have not gone through the rest of byg to see what else makes sense to add. Possibly the glow canes/vines. This is a good start though.

#429


NillerMedDild:
Awesome Motoko, thanks.

PR #510, opened 2020-12-12
Ore drill update

MuteTiefling:
Osmium and Arcane Ore were still using Ore Blocks instead of chunks.

Also updated the script to use a master array for the nether/end biomes, for easier maintenance.


NillerMedDild (approved):

NillerMedDild:
Good stuff! Small note, Arcane may be switching name to mana in the future, can't remember exactly.

PR #513, opened 2020-12-12
Tetra: Missing Improvement

Ridanisaurus:
Some Improvements are missing during the Replacement.
The ones I was able to see are:
- Bronze
- Sky Ingot
- Infused Iron


NillerMedDild:
Ah I hadn't thought of that, ty!

Issue #518, opened 2020-12-13
Blood Magic Alchemy Table


MuteTiefling:
InDev

Need to change the Plant Oil recipe to use 3x any forge:crops instead of hard coded crops.

Gravel to Flint Recipe is broken as well, using 1 gravel and 1 flint to generate 1 flint. 

I'll work on these soon.

MuteTiefling:
Hrm, something is breaking somewhere. I see the scripts are still in place, but they're not loading in the dev instance.

MuteTiefling:
same with the ore processing for Astral... hrm, probably the unification script

NillerMedDild:
I've found the cause, I think :) fixing now

MuteTiefling:
Yeah, that fixes the issue with them being missing, but re-introduces one I was trying to avoid. Namely, the alchemy table now crushes gem ores like lapis and fluorite to mostly useless dust.

NillerMedDild:
Ah my bad, didn't think about that :)

PR #522, opened 2020-12-13
Alchemy and Infuser extras

MuteTiefling:
Just a few cross mod compatibilities that I thought would make these two crafters a bit more useful.  Also completes #518


NillerMedDild:
good ideas!

PR #524, opened 2020-12-14
Pretty Pipes Recipe Re-write

MuteTiefling:
Aimed to bring the overall cost down, making it more accessible early on, as well as to get it using something other than just a metric ton of iron and gold.

Heavily inspired by Thermal recipes, with some Thermal crafting components thrown in.

All basic items have had their cost drastically reduced, while keeping them accessible. Upgrades typically require some sort of alloy, meaning the player would need to mess around with hammering ores to dust, or get into some light machine processing to simplify the process.


NillerMedDild (approved):

NillerMedDild:
Great stuff!

PR #527, opened 2020-12-14
Recipe Updates

MuteTiefling:
First pass of converting to new tags

For whatever reason, this method will not work on things that retain NBT, so RFTools Storage stuff, and Mekanism blocks.

It also doesn't work with any Industrial Foregoing stuff... and a few other odd items that I've marked out in the google spreadsheet (new tab).

Not sure of the best way to handle these.


NillerMedDild:
As discussed on Discord, we found the best approach for the recipes that cannot be replaced is to override them datapack-style.

MuteTiefling:
Should be ready to go with that last commit.

NillerMedDild (approved):

NillerMedDild:
Excellent work :D

PR #542, opened 2020-12-16
Update additions.js

MuteTiefling:
#518 Maybe we go about it this way instead? That resolves the gem dust bit.


NillerMedDild (approved):

NillerMedDild:
Good idea!

PR #550, opened 2020-12-17
Rods and Wires - Just for Fun

MuteTiefling:
I thought it would be nice to have an alternate means of creating these items. With this, it's possible to make the rod/wire press itself directly with Thermal, as well as to make the wires and rods in the multiservo press.


MuteTiefling:
removed the leftover comment section

NillerMedDild (approved):

PR #552, opened 2020-12-18
Thermal Dynamos Rebalance Attempt

MuteTiefling:
Dynamos proved to be pretty underwhelming with default values. In game, I had 4 arboreal extractors going and couldn't supply enough fuel to keep a single compression dynamo running. Needless to say, it wasn't even sufficient to keep a small RS system online.

I have yet to find any config that would allow us to change the base rf/t generation, so dynamos are still capped at 640/rf/t at 73% fuel efficiency, or or 160rf/t at 133% efficiency.

These changes let a single fully upgraded arboreal extractor run approximately 11 dynamos at high efficiency, generating about 1.7k rf/t. It's not going to beat out Ethylene, but I think it makes these much more viable for early game instead of relying on solar or wind. Mind, the starter dynamo with no upgrades only manages 40rf/t per dynamo, and fuel generation would be much slower at that point as well. Generating this tree oil also required phyto-grow in the tree extractor. 

Similarly, lapidary generators get a massive boost, as do numismatics, considering they burn rare items for fuel. Still low output, but much more efficient in their burn rate.

Magmatic and Stirling remain at their default rates as that's really quite sufficient to get started and encourages upgrading to something a little more involved to run.

As a bit of a test, I set up a compression dynamo with 1 bucket of tree oil in it. This ended up giving the expected roughly 4 million RF total, but at a slow rate of 160rf/t. At the max rate of 640 rf/t, I'd expect somewhere around 2.9 million rf.

Lastly, since sapphire and ruby are disabled throughout the pack, this also removes them as possible fuels for the lapidary dynamo.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/compression.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/compression.js
@@ -1,55 +1,67 @@
 events.listen('recipes', (event) => {
     event.remove({ type: 'thermal:compression_fuel' });
-    var multiplier = 1;
+    var multiplier = 10;
     var data = {
         recipes: [
             {
                 fluid: 'pneumaticcraft:diesel',
-                energy: 125000
+                energy: 125000,
+                amount: 250
             },
             {
                 fluid: 'pneumaticcraft:biodiesel',
-                energy: 125000
+                energy: 125000,
+                amount: 250
             },
             {
                 fluid: 'immersiveengineering:biodiesel',
-                energy: 125000
+                energy: 125000,
+                amount: 250
             },
             {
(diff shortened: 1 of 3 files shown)


NillerMedDild (approved):

NillerMedDild:
As always, great work, and good idea to buff the Dynamos.

> As a bit of a test, I set up a compression dynamo with 1 bucket of tree oil in it. This ended up giving the expected roughly 4 million RF total, but at a slow rate of 160rf/t. At the max rate of 640 rf/t, I'd expect somewhere around 2.9 million rf.

I think that's too much power from tree oil, but when the rate is so slow it's acceptable. I hope we get the ability to ramp up throughput at some point!

MuteTiefling:
Yeah, ideally max rf/t can be increased at some point, then some of these can be dropped to a lower total energy density. Tree oil in  particular.

PR #554, opened 2020-12-18
PNC Fuels

MuteTiefling:
With the last update to PNC, we can now add fuels through datapacks.

Added Mekanims Bioethanol (equivalent to current IF and PNC ethanol)
Added IF Biofuel (equivalent to above ethanol)
Added Thermal Tree Oil which give moderate air at a very slow burn rate. Slightly more air than ethanol, but worse than biodiesel.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/pneumaticcraft/fuels.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/pneumaticcraft/fuels.js
@@ -1,0 +1,34 @@
+events.listen('recipes', (event) => {
+    event.remove({ type: 'thermal:compression_fuel' });
+    var data = {
+        recipes: [
+            {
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/pneumaticcraft/fuels.js
```diff
@@ -0,0 +1,34 @@
+events.listen('recipes', (event) => {
+    event.remove({ type: 'thermal:compression_fuel' });
```
NillerMedDild, line 2: Errr..?

NillerMedDild: error during copy, will fix after merging.

NillerMedDild (approved):

PR #556, opened 2020-12-19
Crushing Update

MuteTiefling:
Add BYG and BoP flowers > Dye recipes that were missing on machines
Add Thermal elemental crushing recipes where appropriate.


NillerMedDild (commented):
Could you please change the input dye to `forge:dyes/{color}`? :)

The byg.js file was deleted, I assume you simply renamed it?

MuteTiefling:
Yep. Renamed the byg file while adding missing entries

PR #562, opened 2020-12-19
Saw update

MuteTiefling:
Pedestals and Thermal to come soon.


NillerMedDild:
Good stuff! 

If you use the same variable across multiple scripts, please place it in `constants.js` so they can share it :) that script has a very high load priority, so it will be available.

MuteTiefling:
There's slight differences, but I'll see what I can do. It might require doing some form of blacklist in each script to avoid duplicate recipes since some of these things are covered natively

NillerMedDild:
If it's more than 30 seconds of work don't bother :D I thought they were identical

MuteTiefling:
Should be all set now :D

NillerMedDild (approved):

NillerMedDild:
Nice!!

PR #563, opened 2020-12-19
Essence Update

MuteTiefling:
Using the dissolution chamber for essence conversion turns out to be a little wonky. It requires item input/output. So.. glass panes act as a re-usable catalyst here.


NillerMedDild (approved):

NillerMedDild:
Makes perfect sense :P thanks!

PR #570, opened 2020-12-20
Wood rework

MuteTiefling:
As promised, here's a cleaned up version since I had the time.

This creates a registry of all log/wood types, their associated planks, and their stripped variants (where applicable) in the global constants. It then re-creates the previous saw recipes to use this registry instead.

saw_undergarden was renamed to sawmill as this is now much more re-usable for new additions.

I also took the opportunity to move the axe/crafting table stripping onto this registry.

Lastly, each sawing mechanic has a simple blacklist to prevent re-adding recipes for the mods it supports by default.


NillerMedDild (approved):

NillerMedDild:
Excellent work! :D

PR #584, opened 2020-12-21
missed saw recipes and a typo

MuteTiefling:
Simple Farming Insolator recipes: typo :(

Added Immersive Engineering Saw Mill recipes, but not happy with them. For whatever reason, no matter how I try to add the secondaries (for sawdust output) it just fails to load the recipe entirely.

Good enough for now, I guess. I hope we don't need to do this with datapacks.

undergarden added stripped varieties, so removed exception handling for that.


NillerMedDild (approved):

NillerMedDild:
Thanks!

PR #608, opened 2020-12-22
Update compression.js

### LINKED ISSUE #604: Compression dynamo bugged
**Modpack version**
3.2

**Issue**
Immersive engineering's gasoline does not work in the compression dynamo, neither does LPG, nor liquid ethylene 

**Logs**
no logs

MuteTiefling:
Problem went away entirely when I removed the partial bucket and re-set the energy amount to the appropriate values.

Tree Oil re-balanced back to same levels as ethanol as they're about equally 'complicated' to make in terms of machine infrastructure.

added immersive petroleum's gasoline and diesel

re-tested every fuel. they all work in test.

fixes #604


NillerMedDild:
Thank you!

PR #683, opened 2020-12-29
Removal of some chests from forge:chests

MuteTiefling:
Because it sucks having RS suck up your Smart Chest or Personal chest for something silly like making a drawer....


NillerMedDild:
Looks good. I would like them to be tagged as chests, but the chance of consumption in recipes outweighs that benefit.

Issue #687, opened 2020-12-29
Check if IF Ore Laser is missing ores


NillerMedDild:
Apatite, Cinnabar...?
https://github.com/NillerMedDild/Enigmatica6/issues/684#issuecomment-752263732

MuteTiefling:
Could've sworn we had both of those...

PR #691, opened 2020-12-30
ore-check

MuteTiefling:
update to IF's ore drill. Apatite was missing. Switched to dimensional chunks instead of ore. All other ore chunks appear to be present except Undergarden. That's going to require getting a list of Biomes from the dimension so it can be limited to miners running there. 

Also added a small patch for astral's ore transmutation. It doesn't accept tags that I can see, so I just stuck it to the base EE Iron and Diamond ores. Should help reduce some confusion.


MuteTiefling:
should complete #687

NillerMedDild (approved):

PR #692, opened 2020-12-30
Pedestals cobble gen update

MuteTiefling:
Pedestals apparently now has expanded cobble gen functionality like the random cobblegen. It functions by placing a catalyst block under the pedestal to tell it what type to make.

This adds all the same stones as we've done for cobblegenrandomizer.


NillerMedDild (approved):

NillerMedDild:
Awesome from both you and them \o/

PR #695, opened 2020-12-30
updates to ore related configs

MuteTiefling:
added all ores at similar weights to Nature's Aura.

Also removed a few nonsense enchants from the enchant pendant, like pedestals enchants or things that don't really apply to the player per se.

Note: This is not a global config... it's just a default. So any players with old worlds will need to reset their config in instance\saves\world\serverconfig, unless we have some trickery to force that update?

Further note, despite everything looking like it should work, it does not. I'm PRing these in the expectation that Astral fixes itself and these will simply work in the future.

The entire mineralis ritual itself appears to keep running in an area as well, even when the ritual is dismantled (can still see the particles).


NillerMedDild (approved):

NillerMedDild:
Nice, thanks!

MuteTiefling:
Ah, should have probably hidden RFtools dimensional shards/ores in JEI too

NillerMedDild:
Ah right. 

> Note: This is not a global config... it's just a default. So any players with old worlds will need to reset their config in instance\saves\world\serverconfig, unless we have some trickery to force that update?

We do have some trickery, the mod serverconfigupdater :D

MuteTiefling:
noice

Issue #725, opened 2021-01-03
[Suggestion] Create Ore Buff


MuteTiefling:
Currently, Create ore processing feels like it's quite a bit behind everything else. Doubling with just about any other mod is quite simple and available pretty early on. The initial Create ore processing, however, yields a roughly 1.5x bonus to start and 2.2x bonus at the high end, which involves a fair amount of progression.

I'd propose a slight boost to the primary output and an addition of secondary outputs similar to Thermal.

First, every ore would be paired with a secondary ore following the table below:

| Ore      | Primary  | Secondary |
|----------|----------|-----------|
| Iron     | Iron     | Nickel    |
| Nickel   | Nickel   | Iron      |
| Gold     | Gold     | Zinc      |
| Copper   | Copper   | Gold      |
| Aluminum | Aluminum | Osmium    |
| Silver   | Silver   | Lead      |
| Uranium  | Uranium  | Lead      |
| Osmium   | Osmium   | Aluminum  |
| Zinc     | Zinc     | Gold      |

From here, we boost the numbers of the outputs.

The Millstone currently gives 1 crushed ore per ore. This would be boosted to 1 crushed ore + 25% chance of 2 more crushed ores, resulting in an expected average yield of 2.08x ingots per ore.

In addition, it would have a 5% chance to produce 2 crushed ores of the secondary output. This results in an expected average yield of 0.14 ingots per ore. This is just a bit higher than the IE crusher goes (10% extra) and the Thermal pulverizer is capable of getting up to about 40% bonus material by comparison (requires a basalz farm, effectively).

This all assumes the crushed ores are washed (no changes to the crushed > nugget ratios).

So that gets us a little bit over 2x ore processing, with a fair amount of logistics involved.

The Crusher would see a similar boost to bring it up to 3x processing. This would entail the primary being raised from it's current state (1 guaranteed + 30% chance of 2 more) to 1 guaranteed plus a 60% chance of 2 more crushed ores. This results in an expected average yield of 3.06 x ingots per ore.

Lastly, this too would be granted secondary outputs similar to the Millstone, but at a better rate. Here it would be a 10% chance of 2 secondary crushed ores, resulting in an expected average yield of 0.28 x ingots per ore.

I'm perfectly happy to write the scripts for this, but figured this is a big enough change to warrant discussion. 

[...]

PR #731, opened 2021-01-03
Create Ore Processing Revamp

MuteTiefling:
Jaopca Config - blacklist aluminum from Thermal as it was creating a duplicate processing recipe.

Tagged Create's ores for easy checks

re-did all create ore processing to match #725

added gem processing to crusher (was missing for some reason, even prior to my blanket removal of ore? create has default recipes for diamond and emerald)

Fixing of Secondaries for other mods can be done at a later date.

### DIFF
--- a/config/jaopca/modules/thermal_expansion.toml
+++ b/config/jaopca/modules/thermal_expansion.toml
@@ -1,5 +1,5 @@
 
 [general]
 	#The material blacklist of this module.
-	materialBlacklist = []
+	materialBlacklist = [aluminum]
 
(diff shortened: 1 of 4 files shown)


NillerMedDild (approved):
Looks good!

One note:
```js
  if (tagIsEmpty('#forge:gems/' + material)) {
      return;
  }

  var gemTag = ingredient.of('#forge:gems/' + material);
  var gem = getPreferredItemInTag(gemTag).id;
```
When you're going to be working with the preferredItemInTag method for the tag you're checking isEmpty on, it's better performance wise just to check if gem == air, as this is essentially two tag lookups.

MuteTiefling:
Ah, yeah good point.

PR #744, opened 2021-01-04
Disabled Certus Quartz from Spawning

stefment:
trying to prevent certuz quartz and charged certuz quarts from spawning since it has no use apparantly.


NillerMedDild:
Thank you for your contribution! :)

PR #745, opened 2021-01-04
IE and TS Ore Updates (add missing secondaries)

MuteTiefling:
This brings the Immersive Crusher and Thermal Pulverizer in line with the changes to Create in #725

In the case of the Pulverizer, it already has Cinnabar and Apatite as secondaries for Gold and Tin respectively, so I've opted to keep those as they're generally more important for the mod.

Jaopca is adding duplicates, however, and frankly it confuses me. I can't remove them with KJS and my attempts to work with the config either remove the recipes I'm adding, or otherwise do nothing at all. Any guidance would be appreciated :D

I've tried adding `["aluminum", "uranium", "zinc"]` to the various blacklists and just can't make them go away.


MuteTiefling:
hrm, still getting jaopca recipes in the Create stuff too:

image

NillerMedDild (approved):

PR #770, opened 2021-01-05
Update Blaze.json potential fix #768

### LINKED ISSUE #768: Blaze Bees doesn't use any flower, fungus, or netherwart
**Modpack version**
0.3.4

**Issue**
JEI says blaze bees use netherwarts but they don't. Tried all possible flowers and fungi and nothing worked so the blaze bees can't produce anything atm. They work previously with netherwarts though.

**Logs**
N/A

stefment:
potential fix for #768

### DIFF
--- a/config/resourcefulbees/bees/Blaze.json
+++ b/config/resourcefulbees/bees/Blaze.json
@@ -1,5 +1,5 @@
 {
-  "flower": "tag:forge:crops/nether_wart",
+  "flower": "minecraft:nether_wart",
   "maxTimeInHive": 3200,
   "sizeModifier": 1.0,
   "traits": ["blaze", "nether"],


Ridanisaurus:
Just checked it, and it seems to work. No idea why it's refusing the Forge Tag entry, every though it does recognize it. Will escalate it to the Bee devs.
This is with Forge Tag flower.
image

Ridanisaurus:
Well, tags allow the usage of other flowers that share the same tag. Let's see some mod adds their own version of Nether Warts, this way it would be compatible to be used as well as the Vanilla Nether Wart.
Tags are used as a unification method for inputs and processing.

NillerMedDild (approved):

NillerMedDild:
Thanks :D

Issue #788, opened 2021-01-06
Mining feels like a breeze


stefment:
**Modpack version**
0.3.4 Beta

**Issue**
Because of veinminer and because of ore doubling mining is a breeze. Ores also spawn relatively few blocks apart. So I was thinking about making ores a little bit rarer to slightly extend the early game.

Before: 
![2021-01-06_07 56 14](https://user-images.githubusercontent.com/8888420/103738678-b6a7d900-4ff4-11eb-9ceb-5597a2d022a3.png)

After:
![2021-01-06_07 56 20](https://user-images.githubusercontent.com/8888420/103738685-b90a3300-4ff4-11eb-855d-ab15be19000d.png)

Before; 
![2021-01-06_07 47 47](https://user-images.githubusercontent.com/8888420/103738243-f1f5d800-4ff3-11eb-9047-ace3120a7c98.png)
After:
![2021-01-06_07 47 41](https://user-images.githubusercontent.com/8888420/103738256-f621f580-4ff3-11eb-9428-18f0e1074987.png)

hvdklauw:
I think redstone can use a nerf too

NillerMedDild:
I agree, decreased it a little.

PR #789, opened 2021-01-06
slightly rarer ore

stefment:
for #788


NillerMedDild:
I can get behind making ores slightly rarer, but some of these changes reduce ore count by half, that's overkill imo.

NillerMedDild:
I hope the new values are to your liking, and thank you for both bringing this to my attention and for your well-made PR & issue :)

stefment:
I started with a 20% reduction and went through the caves but nothing seemed to change much. Then i went 50% reduction and honestly it still seems like there is plenty of ore and closely stacked together. Could the config be misleading?

PR #832, opened 2021-01-09
Bee Update - maxTimeInHive

Ridanisaurus:
As discussed, I increased the `maxTimeInHive` by 50%, and increased all the outputs to compensate for this increase. The Output was increased by a double than 50% as most of the time it's an output of 1, and it will also help compensate for the reduction of Block Mutation.


NillerMedDild:
> The Output was increased by a double than 50%

What? :D

Issue #779, opened 2021-01-05
Create Recipe Conflict


KangyKSLG:
**Modpack version**
0.3.4

**Issue**
To create Otherstone from the Occultism mod you need to put 4 stone in an Automated Packing setup.
4 stone for 1 Otherstone.
4 stone also creates stone brick

![image](https://user-images.githubusercontent.com/42292756/103711724-eb457180-4faf-11eb-9d40-94a9ac42eb9c.png)

MuteTiefling:
https://github.com/Creators-of-Create/Create/issues/674

Not seeing a good way to fix this with configs/scripts either. Can't target the recipe because it gets the same ID as the base mod, and Create doesn't appear to have a blacklist for these, it's either all on or all off.

theboo:
the otherstone recipe should not exist, for some reason create is seeing it as a crafting recipe, the recipes for this machine take any 2x2 crafting recipe where all 4 are the same ingredients... the real recipe is trading 4 stone for 1 otherstone with the spirit trader...

https://i.imgur.com/rDwQHPG.png

KangyKSLG:
Ahhh, I had a feeling the otherstone recipe wasn't intended.

Now I need to travel far enough from my base to cause the otherworld stuff to spawn when using the third eye

So far I'm getting nothing so either I haven't travelled far enough, I'm not understanding where to find it, or there is a bug haha

PR #835, opened 2021-01-09
Occultism progression fixes

### LINKED ISSUE #829: Occultism is hard to get started with
**Modpack version**
0.3.5

The new Strata seems to break Occultism's groves and other sources of otherstone, which normally spawn as Andesite (that we remove) but that can be turned into otherstone and mined after eating the demon fruit. Hit reported this on discord, so we should look into it, as it reportedly makes the mod near unusable because of the lack of resources.
Unfortunately I'm not too familiar with the mod so that's the best I can do, but they wouldn't post to github so I had to do it for them

MuteTiefling:
#829: Spirit Fire recipe for converting stone to otherstone
Unstable saplings available in Market for 1 Spirit Attuned Gem

#779: Otherstone trader recipe set to 1 stone for 1 otherstone. This removes the Create recipe conflict. Prior change pretty much removes the need for this recipe at all, however.

### DIFF
--- a/kubejs/data/farmingforblockheads/farmingforblockheads_compat/occultism_saplings.json
+++ b/kubejs/data/farmingforblockheads/farmingforblockheads_compat/occultism_saplings.json
@@ -1,0 +1,18 @@
+{
+    "modId": "quark",
+    "group": {
+        "name": "Occultism Saplings",
+        "enabledByDefault": true,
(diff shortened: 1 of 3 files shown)


NillerMedDild (approved):

NillerMedDild:
Excellent PR, thanks!

Issue #833, opened 2021-01-09
Progression Skip/Vanilla Cobble Gen randomized


DocHuzz:
**Modpack version**
0.3.5 Beta

**Issue**
When creating the bog standard vanilla cobble gen (Water and Lava), the created cobblestone is randomized.  It creates ANYTHING that is tagged cobblestone.  Undergarden, nether stuff, even blackstone.

Blackstone is a requirement for progression in Tetra and is meant to be taken after going to the nether for a T4 hammer and tools.  See video

https://www.youtube.com/watch?v=blw7T9gUCNo

NillerMedDild:
Thanks for the report.

MuteTiefling removed the Blackstone from the list of stones that can be generated, and changed the block that needs to be under the generated block for the randomizing effect to occur.

PR #836, opened 2021-01-09
Change Cobbleworks so it doens't interfere with basic cobble generators.

MuteTiefling:
Returns cobble/stone/basalt generation back to generic vanilla
re-introduces cobbleworks tags on new catalyst blocks. Iron Block for cobble, and diamond block for stone.

removed blackstone from the generic cobbleworks as it is easily found in the nether and is used for progression. #833

Added Purpur Stone and Myalite to both cobblegens.


NillerMedDild (approved):

NillerMedDild:
Hm yeah, this is better, thanks :)

PR #839, opened 2021-01-09
Valid Apiary Block updates

MuteTiefling:
Soul sand. Comes up on discord regularly how to plant wart in the apiary. Currently you'd place soulsand inside the structure, then wart. However, this is contrary to what the book explains to do with flowers/grass. In the book, it says to make the floor out of grass and place the flowers on that.

This allows soul sand to be part of the structure now, with just the wart inside. Makes things a bit roomier.

Also added botania glass for decoration purposes as requested on discord


NillerMedDild (approved):

TemperedFool:
I would also request that the glass from the mod that adds both Etherial and Glowing glass be added; seems on theme.

PR #922, opened 2021-01-17
Crushing Recipes Update

MuteTiefling:
Unified inventory ore crushing on the new EE hammer. Removed IE hammer crushing recipes. EE recipe now uses `forge:ores` as input.

Removed now redundant ingot/gem to dust recipes. Jaopca handles them when it's config isn't broken! It does not handle them all, however, so the rest can stay. Yay!

Added missing gem > dust recipes for inventory earth charges.

renamed milling_rods to just milling, made generic to deal with any future recipes.

added byg and bop sandstone to sand crushing for a variety of machines to match recipes for vanilla sandstone crushing. also added to unpacking through thermal, matching vanilla sandstone.


MuteTiefling:
Caught an issue with the press recipe. fix incoming

Issue #817, opened 2021-01-08
[Enhancement] Akashic Tome and MorphOTool


Merlo17:
https://www.curseforge.com/minecraft/mc-mods/akashic-tome
https://www.curseforge.com/minecraft/mc-mods/morph-o-tool

Good QoL stuff to have for sure

NillerMedDild:
Will add :)

Merlo17:
Ah if you want I can spare you some of the dirty work and PR you the default books and wrenches to put in the config for them when they're in (haven't looked at it but I assume it's just some boring finding them and putting the ids in?)

NillerMedDild:
That'd be nice!

Yeah that's pretty much it - It'd also be nice if you could create "filled" morph-o-tool and akashic tomes and paste the first output of `/kubejs hand` here, so I can make recipes for them :D

Merlo17:
Alrighty, will look into that

Merlo17:
I have them mostly done but I'm stuck because of this issue I reported:
https://github.com/Vazkii/AkashicTome/issues/58

Whenever it gets fixed I can get the last few things in and put the recipes in place

Merlo17:
Akashic Tome:
[...]

PR #926, opened 2021-01-17
Dummy Recipe set up for Akashic Tome

MuteTiefling:
@Merlo17 managed to get the tome in. Needed to escape out some quotes in nested nbt (cooking for blockheads). Just change the map/ingredients at the end to whatever you like. 

#817


Merlo17:
ty, I moved it to shapeless though, so this commit is probably wrong :P thanks anyways, it was too late for me to figure that out yesterday, you saved me a headache

PR #970, opened 2021-01-20
Buffed Tetra custom materials

Merlo17:
This hopefully feels a bit better, I tested each material with a stock tetra pick against either its vanilla counterpart or an equivalent of sorts. I only touched the efficiency stat, so we might want to go through them again and tweak other values if necessary. 

I would also like to know what you guys think about them, because I'll be honest, after testing a lot of them they all started to feel kinda the same and it was hard to gauge how good they felt, so outsider feedback would be helpful :P


NillerMedDild:
I think the values look good for what it's worth.

PR #996, opened 2021-01-22
Bee Update (Biomes, Secondary Outputs, and Custom Honey)

Ridanisaurus:


NillerMedDild:
I must admit, I have not read through all 123 files :P thank you for yet another huge contribution Rid!

Ridanisaurus:
I did a terrible job at adding description, clearly. So just for tracking purposes I'll see if I can add them here.
- Add BYG Biomes 
- Disabled the Forge Default Biome from the config so we can tab into the custom biomes
- Added 3 biomes to each of the main Bees to help make them more targetable
- General Bees like Gravel, Sand, Water ..etc. are default biomes only
- Added custom honey to Metals, Alloys, and some of the General Bees
- Added Vanilla potion effects to the custom honeys (mostly level 1, and couple level 2)
- Updated the Secondary and Bottle output to almost all the bees
- Added BooBee as an Easter Egg because Boo was a good boi for 48 hours :D 

Might have missed something here or there, but this should cover at least 95% of the changes.

NillerMedDild:
Moved to the top for visibility.

PR #999, opened 2021-01-22
Creosote Unification

### LINKED ISSUE #988: Suggestion: Add Use to Thermal Creosote
**Modpack version**
0.3.7

**Describe the bug(?)**
Thermal added Pyrolyzer, which allows you to make Creosote. It doesn't have any use, unlike IE Creosote, which can be used to make Treated Wood

MuteTiefling:
Thermal Machines now make Immersive Creosote. Thermal Creosote should no longer be obtainable, though it can still be burned as fuel to get rid of it.

Added Immersive Creosote to Compression Dynamo, as it was previously missing.

Added In World Crafting as alternate for making treated wood just for fun. It will, on average, use the same amount of creosote, but since it's percentage based, promotes large batches since a small batch may very well end up producing less.

Resolves #988

### DIFF
--- a/kubejs/data/thermal/recipes/machine/pyrolyzer/pyrolyzer_coal.json
+++ b/kubejs/data/thermal/recipes/machine/pyrolyzer/pyrolyzer_coal.json
@@ -12,7 +12,7 @@
             "chance": 0.25
         },
         {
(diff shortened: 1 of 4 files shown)


NillerMedDild:
Nice solution!

PR #1002, opened 2021-01-22
BM Compatibility Review

MuteTiefling:
Jaopca now supports BM. Adjusting our recipes.

Alchemy Table Metal Ore recipes removed. Leaving Gem Ore recipes for parity with other crushers

Revert Gold/Iron processing changes back to BM's materials.

Unhide hidden BM fragments/gravels/sands

Remove our  ore processing from ARC to default to Jaopcas as they're identical to what we had, except using fragments/gravels/sands.

Remove ARC Ingot > Dust. Already covered by Jaopca.


MuteTiefling:
Not merging yet since I don't think JAOPCA is updated

NillerMedDild:
Updated JAOPCA and a few others, do not use F3+T it will crash atm :P

PR #1013, opened 2021-01-23
Update occultism-server.toml

theboo:
Doubled the tier 1 and tier 2 stabilizer slots and quadrupled tier 3 and tier 4. This should put it more in line with both the difficulty in crafting these(the tier 3 and tier 4 each require a nether star and dragon egg respectively, and you can add 6 to each storage so it gets expensive pretty quick) and to make it a viable storage option for someone looking to stay on the magic side of things rather than just plonking down a big RS system.


NillerMedDild:
I don't even remotely understand Occultism's storage system, but I'll trust you on this :P

PR #1016, opened 2021-01-23
Sulfer Nerf

MuteTiefling:
This -should- nerf sulfur down to reasonable levels. I need to set it up to run for a few hours to see though.


NillerMedDild:
You don't need to remove the existing one?

MuteTiefling:
Not sure. After looking at it the only removals we had were for ores we don't want. I'm hoping this will simply override the base value.

PR #1021, opened 2021-01-23
Nature's Aura Ore Restructuring

MuteTiefling:
Think we're going to need to make a request to NA for overrides here. I should be seeing a lot less sulfur with this setting and I'm not. My original though of removing/re-adding is no good either because if I blacklist all sulfure types, it simply isn't going to generate at all.

Have blacklisted many other ores, however, so there's that.


NillerMedDild:
10/10 title :D

PR #1101, opened 2021-01-27
Add Tags to facilitate Iron Chest quests

MuteTiefling:
Add tags to the upgrades and chests so that the quests can accept either.


MuteTiefling:
Took this naming convention as it's probably best to keep them out of the mod's own name when it's strictly related to our quests.

MuteTiefling:
Will update quests once dev server can be updated

PR #1109, opened 2021-01-28
Blacklist Mimicubes from spawners

MuteTiefling:
Mimicream is used to dupe items legitimately.

8 mimicream around any item with durability in a crafting grid will create a 1 durability copy of that item with many/most of the enchants intact. So it makes a pretty interesting way to duplicate enchants or even rare items.

However, it's kinda too much to be able to spawn them on a whim.


NillerMedDild:
That is one weird mechanic :P

Issue #861, opened 2021-01-12
[Sugestion] Balance Power Generation


ManiacRobert1:
Yes, i know the pack it's a kitchen sink pack, but whole power system can be balanced.

Let's take as example powah reactors, a basic one can generate stupid amount of FE(6.5k/t) for small amounts of iron, redstone, coal, clay and lava.

If you don't use mekanism machines in crazy-mode as possible, you don't need another power-gen in your game progression.

Another example is gas-burning ethylene.

![image](https://user-images.githubusercontent.com/39351347/104329393-84671180-54f5-11eb-99b8-4795e26a356e.png)

That is my example of gas-burning generator, output from melon farm goes into thermal sawmill, 1 melon = 9 melon slice, 1 melon slice = 4 bio fuel => 1 melon block = 45 bio fuel(+ output of seeds), with that output you can generate 100krf/t with a small amount of effort, so, powah reactors / generators are 100% useless, fission is useless in basic shape for FE gen.

My sugestion about powah reactors is to buff with 30%-70% energy generation (10%-15% to nitro stage).
About mekanism gas-burning generators, should have almost same amount of energy production how was it in e2e.

And now, about Extreme / Bigger reactors.

Yes, i know, we have Powah, Mekanism, IF etc, but, there are some pros about that mods.

- Can be added Custom Liquids(i see on direwolf gameplay astral liquid).
- Cool CC: Tweaked community-made apps for reactors.
- In case of Extreme Reactors, the textures has been changed.
- Power output it's not that powerfull as mekanism or powah.
- Did I mention cool looking apps by community on the monitor?

Darkere:
Agree that we do need to take a look to at least somewhat balance out power generation. 

That said I don't think big reactors fullfills any roles that aren't already filled by mekanism and powah apart from the variable internals. Those might be covered by another mod we are currently considering.

MuteTiefling:
Just to put a few numbers out there, here are some of the bigger generators in the pack. 

| Energy Producer                       | Fuel Input                                 | Energy Output (MFE) | Rate (kFE/t) | Rate (FE/t) |
|---------------------------------------|--------------------------------------------|---------------------|--------------|-------------|
| Basic Fission + Basic Turbine (quest) | 8000 mB Fissile Fuel (8 uranium ingots)    | 228.28              | 570.91       | 570910      |
[...]

PR #1125, opened 2021-01-29
Mekanism Power Rebalancing - Ethylene is more expensive and slower to make. Fission/Fusion reduced in efficiency

MuteTiefling:
More info in the commits themselves. 
#861

### DIFF
--- a/config/Mekanism/general.toml
+++ b/config/Mekanism/general.toml
@@ -70,7 +70,7 @@
 		#Disables Forge Energy (FE,RF,IF,uF,CF) power integration. Requires world restart (server-side option in SMP).
 		blacklistForge = false
 		#Maximum Joules per mB of Steam. Also affects Thermoelectric Boiler.
-		maxEnergyPerSteam = "10"
+		maxEnergyPerSteam = "2"
 		#Conversion multiplier from Forge Energy to Joules (FE * JoulePerForgeEnergy = Joules)
 		JoulePerForgeEnergy = "2.5000"
 		#Conversion multiplier from EU to Joules (EU * JoulePerEU = Joules)
@@ -142,7 +142,7 @@
 		radioactiveWasteBarrelProcessTicks = 1200
 		#Number of mB of gas that decay every radioactiveWasteBarrelProcessTicks ticks when stored in a Radioactive Waste Barrel. Set to zero to disable decay all together. (Gases in the mekanism:waste_barrel_decay_blacklist tag will not decay).
 		#Range: 0 ~ 9223372036854775807
-		radioactiveWasteBarrelDecayAmount = 1
+		radioactiveWasteBarrelDecayAmount = 100
 		#Radiation sources are multiplied by this constant roughly once per second to represent their emission decay. At the default rate, it takes roughly 10 days to remove a 1,000 Sv/h (crazy high) source.
 		sourceDecayRate = 0.9995
 		#Defines the minimum severity radiation dosage severity (scale of 0 to 1) for which negative effects can take place. Set to 1 to disable negative effects completely.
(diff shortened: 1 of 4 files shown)


NillerMedDild:
Excellent changes, and nice documentation/research on it. I would like to hold off on merging this until 0.4.0 though, to prevent disasters :D

NillerMedDild:
I would like create a Wiki on this repo to document changes like this going further (this is mostly a note to self)

MuteTiefling:
> Excellent changes, and nice documentation/research on it. I would like to hold off on merging this until 0.4.0 though, to prevent disasters :D

I thought as much. Would be pretty horrible to log in and have your reactor blow up.

TemperedFool:
Would be a good test of player's backup systems...

Merlo17:
Just as a kinda obvious note for us later, it might be good to mention the changed values and the wiki link in game, maybe in jei tabs or quests

Kataiser9:
Here's my unsolicited $.02. Apologies for the long post... :'D

I disagree with the idea of nerfing fusion by two orders of magnitude in a non-expert pack, partly because _me like big numbers_ but mostly because I am not sure that it will achieve anything meaningful.  There is virtually no functional difference between 1M RF/t and 100M RF/t in a non-expert pack; even with truly outlandish factory setups you're not gonna get energy costs anywhere near that high (unless you intentionally avoid using any energy upgrades, which would make no sense considering how trivial they are to mass-produce). My SSP world with fairly extensive factory setups usually sits at 10-30k depending on use, and I doubt that most players will ever break 100k.

So this seems like a somewhat arbitrary nerf, but I realize I haven't provided much reason _not_ to nerf it. There's only one thing in this pack that uses fusion-tier energy levels, and of course that's the SPS, which is balanced around that level of energy. With the proposed 3.18M RF/t water-cooled reactor, it would take over four hours to produce a single pellet of antimatter, or about 14 hours with the air-cooled version; this doesn't really matter if you're on a server and your stuff's always running, but for SSP by kitchen sink standards that's a little nuts.

Thus, if you're set on nerfing fusion this hard, I strongly recommend revising the SPS energy costs to match.

On a less important note, is there a way to change the energy storage of the Induction Matrix? I didn't see one from a quick glance through the configs, but IMO if you nerf fusion, it would make sense to lower those values as well if possible. With the strongest power generation in the pack reduced to 1% of its normal power, it'd become quite pointless to make anything beyond a small, low-tier Induction Matrix, which would be a bit of a shame. (Of course, this isn't exactly critical since the same is _mostly_ true with default configs, but I thought I'd bring it up.)

MuteTiefling:
> 3.18M RF/t water-cooled reactor

Do note, this is with a _very_ sub optimal setup. 

My aim in bringing basic fusion down this far was to incentivize using d-t as the booster it is meant to be. As mentioned, just a single Chemical Infuser making D-T fuel maxes out the turbine I was testing with and would end up producing a lot more energy.  5 million more rf/t in fact... and that capped out the turbine. So clearly another turbine could be useful there. 

If you're going for something as extreme as antimatter, you now have a reason to actually attempt to optimize your turbine.. As it stands, several people on the official servers have made antimatter with ethylene, and not slowly I should add. For my part, I stored all the energy from processing fissile fuel and made quite a bit of antimatter that way. 

Some changes have been implemented in Mek that will allow us to revisit this, however. My hope was to reduce the passive power gen as much as possible and force the use of the turbine for all energy. With current configs that's not possible, but the new configs will allow it. So I'll for sure be revisiting this again soon. 

If it turns out that antimatter is really out of reach with these changes, even with a better turbine than mine, then we can certainly revist the SPS.

NillerMedDild:
@MuteTiefling Would you like this merged now, or should I wait till you've had a look at the new options? :)

MuteTiefling:
May as well merge now. It'll make it easier to tweak when Mek finally does update

NillerMedDild:
Documented through your dev notes here: https://github.com/NillerMedDild/Enigmatica6/wiki/Mekanism
Feel free to add/edit that as you see fit.

PR #1135, opened 2021-01-29
Unify candles under forge:candles tag

theboo:


NillerMedDild:
Thanks Boo, good stuff!

PR #1195, opened 2021-02-02
Fix industrial foregoing sludge refiner cost 1000FE/t

shengjie:
See issue https://github.com/InnovativeOnlineIndustries/Industrial-Foregoing/issues/873

Machine will eat your RF but do nothing until you provide 1000FE/t

To reproduce this bug: simply put an sludge refiner with some sludge plus a stirling engine. you will see machine eats all RF an do nothing.

It has been fixed in code, however the config still uses wrong value.


NillerMedDild:
Thank you for your contribution! Would it be possible for you to change the target of this PR to the develop branch? :)

NillerMedDild:
Perfect, thanks!

PR #1209, opened 2021-02-04
Oil and Bitumen Unification

MuteTiefling:
Thermal and IP crude oil should no longer be obtainable. 

The IP Pumpjack will now pump up PNC oil. 

Thermal oil, previously obtained by unobtainable bitumous sands, has had its recipe replaced with a similar EE ore/chunk to oil recipe in the centrifuge.

PNC oil now works as the input for any oil recipe.


MuteTiefling:
Would like to unify lubricant, gasoline, and diesel, but I can't quite do that yet. The auto lubricant thing from IP doesn't want to accept PNC lubricant and I can't change it yet.

https://github.com/Flaxbeard/ImmersivePetroleum/issues/223

Suppose we could go the other way around though, and unify on IP's fluids.

Issue #1206, opened 2021-02-03
IF Laser Lens


Dracos99:
**Modpack version**
0.3.13
**Describe the bug**
Magenta / Pink / Cyan / Green Lenses from IF dont have a targeted ore or fluid...making them useless. maybe they could have undergarden ores and/or other fluids added to them.

Merlo17:
I'll look into this tomorrow ^^ surprised green doesn't have emeralds tbh

PR #1211, opened 2021-02-04
Overhauled IF laser drills

Merlo17:
As suggested by https://github.com/NillerMedDild/Enigmatica6/issues/1206

-Split up some ores to fill up all lenses (for example, lime had uranium and emeralds and green was useless, I gave emeralds to green and left uranium in lime)
-Added undergarden ores restricted to undergarden biomes
-Added some fluids for fun (might want to look over there as they might be seen as progression breaking so I'm not sure if you want to keep them. I added them anyways, your call)


NillerMedDild:
Looks good, nice work 👍

Issue #1219, opened 2021-02-05
Can't use EE steel plate in multiservo press mold recipes


Rodg88:
**Modpack version**
0.3.13

**Describe the bug**
Making the "metal press mold: wire" in the thermal multiservo press requires the IE steel plates, it doesn't accept the emendatus enigmatica ones. Looks like the same for the rod mold, but haven't checked that one in-game.

Merlo17:
Oh yeah, something definitively went wrong there, some of the other molds don't even have recipes. I'll look into it

PR #1222, opened 2021-02-05
Fixed IE mold recipes in TE press

Merlo17:
Fixes https://github.com/NillerMedDild/Enigmatica6/issues/1219

Also added some recipes that were missing. @MuteTiefling was that on purpose or just an oversight?


MuteTiefling:
It was intentional but adding these is fine I think. 

These recipes were meant for supporting the creation of wires and rods in the multi press without needing to go into IE to get treated wood and all that. 

It already natively supports gears and plates, so I had left those out.

Merlo17:
Ah I see now. They will probably never get used but just for the sake of completion I'd keep them, it feels weird to only have a couple craftable to me

MuteTiefling:
Sure. I think they're technically more expensive than the ie route anyway. But yeah, options!

NillerMedDild (approved):

PR #1240, opened 2021-02-06
Proposed new Flux Networks recipes

Merlo17:
Here's the reasoning behind these: Flux Networks is mostly in the pack to be a direct endgame extension of Powah, since everything it does pretty much invalidates the stuff powah does. As such, it's positioned as a sort of "final upgrade tier" to Powah's items, requiring progression through the whole mod.
Hopefully this doesn't feel too gated for a non expert pack, but I would say it's fair to have them as an upgrade of sorts and treat them basically as part of powah itself instead of a standalone mod, because of how close the features offered are. 
Here are the recipes:

Replaced ender eyes with powah ender cores just for fun (flux core recipe now makes twice as much to make up for that)

Flux Point is basically a max tier ender gate, their functionalities are basically identical and as such it's positioned as an upgrade to the nitro tier gate.

The same can be said for the controller, since it allows for wireless charging, and thus is positioned as an upgrade to the nitro player transmitter

Other recipes (flux plugs in particular) are untouched, since they're basically useless without their counterparts and don't really need more balancing.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
@@ -60,6 +60,9 @@ events.listen('recipes', function (event) {
         'eidolon:tallow',
         'engineersdecor:dependent/slag_brick_block_recipe',
 
+        'fluxnetworks:fluxcontroller',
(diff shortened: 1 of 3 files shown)


NillerMedDild:
Looks good, nicely done :)

PR #1250, opened 2021-02-07
Recipe Cleanup and expanding Create's Fan support for more blocks.

MuteTiefling:
Allow Create Fan to blow through more blocks
Add missing recipe for Embur Hyphae
Hide Additional Bars using BOP materials


kubejs/server_scripts/enigmatica/kubejs/base/recipes/shaped.js
```diff
@@ -376,6 +376,9 @@ events.listen('recipes', function (event) {
             A: '#forge:wax',
             B: '#forge:string'
         }),
+        shapedRecipe(item.of('byg:embur_hyphae', 3), ['AA ', 'AA '], {
```
NillerMedDild, line 379: This pattern will result in the recipe only working on the left side of the crafting grid `['AA', 'AA']` would be better imo :)

MuteTiefling: Right! Will change that

Issue #1251, opened 2021-02-07
Compact Powah Recipes


Merlo17:
Suggested by someone on discord, create recipes for Powah's orb with blocks of items, to allow for faster automation.

PR #1258, opened 2021-02-08
Added compact recipes for Powah energizing

Merlo17:
Refactored the script a little bit, should be easier to add new recipes now. Array solution still isn't that good but js won't let me use math :(

Recipes are the same as default but with blocks and *9 the energy cost, for easier bulk crafting. Someone suggested making them cost less power but I don't think it makes too much sense gameplay wise, the numbers are easy to tweak anyways if we want to.

Also along the way I noticed IE uranium blocks weren't hidden in JEI so I fixed that too

From https://github.com/NillerMedDild/Enigmatica6/issues/1251

### DIFF
--- a/kubejs/client_scripts/constants.js
+++ b/kubejs/client_scripts/constants.js
@@ -172,6 +172,8 @@ var regexHide = [
     /minecraft:\w+_ore/,
     /immersiveengineering:plate_/,
     /immersiveengineering:stick_/,
+    /immersiveengineering:storage_/,
+    /immersiveengineering:ore_/,
     /create:\w+_sheet/,
     /create:\w+_nugget/,
     /create:\w+_ingot/,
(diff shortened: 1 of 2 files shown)


MuteTiefling:
Also, as a point against making them cheaper... the reactors make a huge return on this amount.

kubejs/client_scripts/constants.js
```diff
@@ -172,6 +172,8 @@ var regexHide = [
     /minecraft:\w+_ore/,
     /immersiveengineering:plate_/,
     /immersiveengineering:stick_/,
+    /immersiveengineering:storage_/,
```
NillerMedDild, line 175: This should not be hidden, as the blocks can be used for building and are obtainable.

NillerMedDild: Very very very minor thing, I'll nuke it myself :)

Merlo17: but they can only be obtained via the stonecutter, so it works like every other block/ore like that? personally I would hide and maybe put an info tab/tooltip telling you those blocks are avaiable

NillerMedDild: Hmm, good point.

Issue #1257, opened 2021-02-08
Silicon


Toonlink264:
**Description**
add silicon from sand recipe's in crushers i.e. create/mek/thermal

**Why would you like this added/changed?**
material stonework factory gets stuck if it gets full wouldn't mind a different method of silicon that isn't place block down, get silicon

Merlo17:
I'm not super sure what to make of this: the stoneworks factory is the only place in the pack where silicon comes from cobble, while everywhere else it comes from quartz.
I wonder if we should keep the stoneworks recipe at all, it feels a bit out of place doesn't it?

Otherwise, my idea is to make silicon available via crushing create limesand

Toonlink264:
if you take out the silicon from the factory it loses literally all purpose for the work that goes into getting one

Merlo17:
Does it? It's an all in one compact machine to make cobble, gravel, sand, glass and derivate products. Still very useful if you ask me. Compared to the other recipes in the pack it doesn't really make much sense that it can do silicon does it?

MuteTiefling:
Yeah, that seems a little extreme. There's plenty of reason to still make the stonework factory.

PR #1264, opened 2021-02-08
Added processing recipes for Limesand to Silicon

Merlo17:
As suggested by https://github.com/NillerMedDild/Enigmatica6/issues/1257

I wasn't sure at first (and I'm still not entirely), but it does make sense to add a way to process silicon with create cobbleworks if the material stoneworks factory can do it too.
The automation chain is pretty long anyways, and it only has a chance to actually produce, so the smelting quartz way is still the preferrable route, but there is at least a create alternative that doesn't involve quartz.
Still up for discussion though

### DIFF
--- a/kubejs/client_scripts/constants.js
+++ b/kubejs/client_scripts/constants.js
@@ -126,6 +126,7 @@ var itemsToHide = [
     'quark:biotite',
     'quark:biotite_ore',
     'quark:tallow',
+    'refinedstorage:silicon',
     'thermal:coal_coke',
     'thermal:coal_coke_block',
     'thermal:ender_pearl_dust',
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/splashing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/splashing.js
@@ -2,12 +2,16 @@ events.listen('recipes', (event) => {
     var data = {
         recipes: [
             {
-                output: 'buildinggadgets:construction_block_dense',
+                outputs: ['buildinggadgets:construction_block_dense'],
                 input: 'buildinggadgets:construction_block_powder'
+            },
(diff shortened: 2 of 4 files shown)


NillerMedDild:
I like the Limesand recipe, I think it makes sense and I think it should be applied to Mekanism and Thermal as well. Will you do the honors? :)

Merlo17:
Honored to do the honors! What concerns me slightly is that they probably don't have a way to make the output chance only... or can they? Will investigate

MuteTiefling:
Thermal can do chance output. Mekanism cannot.

Merlo17:
Just found out too, to implement that I increased the input count for those who can't do chance. Mek is 4 -> 1 for example

OneLemonyBoi:
Maybe add an in-world crafting option via Interactio, which is more inefficient.

Merlo17:
> Maybe add an in-world crafting option via Interactio, which is more inefficient.

how would you set up the craft? limesand into what?

Merlo17:
Also I added mekanism and thermal recipes. I wanted to add IE and pedestals crushing recipes too, but unfortunately they don't support chance output or multiple inputs, so my only option would've been 1:1 and I still want to keep these recipes more expensive than the quartz ones, so no silicon for them :(

OneLemonyBoi:
> > Maybe add an in-world crafting option via Interactio, which is more inefficient.
> 
> 
> 
> how would you set up the craft? limesand into what?

3:1 ratio for limesand to silicon.

MuteTiefling:
Create Washing is basically that, but more interesting.

Merlo17:
Agreed, I don't particularly think that it does anything either

NillerMedDild:
I think 4 -> 1 is a bit harsh in the Mek Crusher, would you mind if I change it to 3?

PR #1273, opened 2021-02-09
Better End Integration Pass - Trees and Sawing

MuteTiefling:
Added Wood types. Saws and Log Stripping appear to have worked without issue.

Will need to do more testing to add plants/trees to those registries. Could not get any saplings to grow in my initial test. Will work on that tomorrow.


theboo:
each sapling has its own grass/moss/mycelium that it can grow on, so might want to use that for cloches/pots?

MuteTiefling:
Yeah, I'll probably do something like that. Just need to actually find all the pieces.

Issue #1276, opened 2021-02-09
Better End Compat


Merlo17:
TODO list:

- [x] Add alloy recipes to better end furnace
- [x] Add better end alloys to other alloying machines
- [x] Hide duplicate items (ender dust) and replace input with the EE one (annoyingly it's not tagged as forge:dusts/ender so we gotta replace)
- [x] sapling compat with phyto/cloche/botany pots using proper grass/moss/mycelium for each sapling
- [x] Tetra Material
- [x] Tetra Replacements
- [x] Unify their sulfur (Remove it and it's recipes)
- [x] Add missing tags

MuteTiefling:
I'm working on the pots/cloche/insolator integration. Should have it done tonight.

MuteTiefling:
#1282

MuteTiefling:
Leaving in their sulfur as it is actually rather unique. It grows on brimstone soaked in water and is harvested as a crystal with silk touch. I did add it to the sulfur gem tag, however, so that it can be ground down to sulfur dust in various recipes. 

There are more tags that need doing, I think.

PR #1278, opened 2021-02-09
Betterend compat

Merlo17:
Big batch of compat, mentioned here https://github.com/NillerMedDild/Enigmatica6/issues/1276

only note is I couldn't add every alloy recipe to the smelter, since the recipes only accept a single item as input, so for example bronze which is 3+1 couldn't be added.


NillerMedDild:
Great stuff! The majority of `betterend/alloying.js` would be better suited as part of the `unify_materials.js` script - The one that adds gears, ore processing and the like. Would you like to take a crack at it? :)

Merlo17:
sure, I will have a look later tonight ^^

Merlo17:
Should be all done! Along the way I also noticed IE ores weren't hidden in JEI so I fixed that and pushed to the wrong branch so now it's in the PR and you're gonna like it :P

NillerMedDild:
> I also noticed IE ores weren't hidden in JEI so I fixed that

They're hidden with regex :P

Merlo17:
Are you sure? I couldn't find it in the regex array and they were unhidden on the first launch without reloads

NillerMedDild:
Did you forget to pull perhaps? They're in the regex array, line 177

Merlo17:
Huh, I knew I had to go get new glasses, better go tomorrow :P

PR #1283, opened 2021-02-10
ID Menril integration

Merlo17:
Added to our tree registry and the market, I also fixed thermal's sawmill compat because I noticed it was broken and relayed on their discord. It's a bit of a hacky fix imo but it works (there were duplicate recipes, one with plugins and one without in the ID, and both had their input broken).


MuteTiefling:
In `trees.js` there are two sections, one for `tree` and one for `tree_shroom`. You'll want to add this to the `tree` section.

Also, for sawing, you just need to add the wood type to `wood_variants,js` that'll get it added to all the various sawing machines.
Unless their wood types follow some weird format, you can just add `'integrateddynamics:menril'` to the `woodVariantsToConstruct` variable. 

Last thing, but I'm not sure it's good to use a `const` there, as that makes it globally available to anything running after. data is generally reinitialized and re populated each time it's needed, but there's no need for it to be global.

Merlo17:
Alright, gotta learn our unify scripts better I guess, should've expected we had one for wood. I'm away rn but I'll fix it up when I'm home

Merlo17:
Moved it over and edited our scripts to account for the fact that wood or stripped variants don't exist. It's all contained in our scripts so if one day ID does add them we can just remove the extra checks. Ugly fix, but it works

NillerMedDild:
Good work, although not gonna lie, that if block physically hurts to look at :P

Merlo17:
Tell me about it :P

PR #1284, opened 2021-02-10
Cobbleworks Migration and Expansion

MuteTiefling:
Migrated cobbleworks recipe into unification script as it's now managing both Pedestals and Industrial Foregoing.

cobblegenrandomizer is supposed to support datapacks too, so if we can figure that syntax out, we could migrate that config to this script too. Looking into that.


MuteTiefling:
Hrm, actually that looks like it's done through loot tables, rather than a recipe system So probably not possible to migrate into scripts, right? Those would just be plain data pack overrides.

NillerMedDild:
KJS can do loot tables now I believe, but there's no need to migrate the CobblegenRandomizer stuff as it already relies on the same constants that the new cobbleworks unification relies on :)

PR #1289, opened 2021-02-11
structure boosts

theboo:
moderate boosts to botw towers, and large boosts to some valhesia structures


theboo:
nope

NillerMedDild:
What happened here? 😄

theboo:
Shhh we don't talk about this pr... I made the changes to an older version of the config lol, was easier to just re-do everything as I'm not at home

Issue #1277, opened 2021-02-09
Abnormal Mods


NillerMedDild:
Alright, alright. I cave.

We're adding the following Abnormal mods:

- [x] [Upgrade Aquatic]( https://www.curseforge.com/minecraft/mc-mods/upgrade-aquatic)
- [x] [Atmospheric](https://www.curseforge.com/minecraft/mc-mods/atmospheric)
- [x] [Autumnity](https://www.curseforge.com/minecraft/mc-mods/autumnity)
- [ ] [Environmental (Not yet released)](https://github.com/team-abnormals/environmental/issues)

Which means we need another large compat pass. Sorry guys :P

Merlo17:
dumb compat dirty work? don't mind if I do :P

Merlo17:
Also as far as I can tell swamp and bloomful are on 1.15 still but are being ported and combined into one mod at https://github.com/team-abnormals/environmental. something to keep an eye on I guess, not on curse yet

CletusMAngo:
YAYYYYYYYY MY EFFORTS WERE SUCCESSFUL 
Thanks so much guys!

MuteTiefling:
Gonna start collecting data for compat

- [ ]  Trees
- [ ]  Wood Types
- [ ]  Growables
- [ ]  Dyes
- [ ]  Other materials (already noticed new sandstone types that need crushing recipes)

CletusMAngo:
Environmental just released for 1.16 https://www.curseforge.com/minecraft/mc-mods/environmental

CletusMAngo:
They also added a mod that adds integration for Abnormals food and Farmers delight  
https://www.curseforge.com/minecraft/mc-mods/abnormals-delight

PR #1294, opened 2021-02-12
Abnormals Integration Pass

MuteTiefling:
Update wood, dye, and plant, tree, and soil registries.

Move Create cutting into unify_sawables with all the other sawing recipes...

update sand crushing recipes.

add integrated dynamics squeezers to unify_dyes. Cut base output, but introduced bonus materials similar to other dye recipes.

Note, ID removals of ID stuff are a bit tricky since JEI doesn't show the ID. Have to `/kubejs export` or infer from the folder structure of the mod.

~updated any `item.of` to `Item.of` where noticed.~

updated search tags for abnormals mods.

added a few abnormals stones to cobbleworks

blacklisted Occultism Summoned mobs from spawners


MuteTiefling:
https://github.com/NillerMedDild/Enigmatica6/issues/1277

PR #1298, opened 2021-02-12
added new biomes to conf and set weight to 10 for testing

theboo:


theboo:
hold up dont commit yet

theboo:
oke all good lol

Issue #1175, opened 2021-02-01
It is a downgrade to put vanilla bees in tier 3 apiary, compared to tier 2


HaxtonSale1:
**Modpack version**
0.3.8

**Describe the bug**
Vanilla honeycombs need 4 to make a block, Resourceful bees' need 9
T2 apiary makes 6 honeycombs, T3, 1 block

NillerMedDild:
Thank you for the report, I will investigate the issue.

PR #1303, opened 2021-02-13
Comb recipes

Merlo17:
Decided to keep in the same file at the end of the day since we already had the code up to replace ids and I didn't notice, I think a section at the end is good enough.

Also added resourceful bees thermal compat, we'll just need to rememeber to add new bees we add to the const array


NillerMedDild:
Fixes #1175

PR #1313, opened 2021-02-15
Better End : The Taggening - Add basic tags to many items/blocks from better end 

MuteTiefling:
Did a bit of cleanup in the tag scripts I touched too.

The ores are not yet handled by unification properly. Amber ore outputs nothing and end ore is giving pearls in some cases.

Both of these drop 'shards' that are used for different things. I've tagged them as such for now, but perhaps they should go into another tag?

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/crushing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/crushing.js
@@ -6,8 +6,19 @@ events.listen('recipes', (event) => {
                 input: 'minecraft:obsidian'
             },
             {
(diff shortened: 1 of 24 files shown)


Merlo17:
To add to this a bit, I'd rather hold off on the tags PR until I'm more familiar with forge and modding. I've looked into it this afternoon and though I could probably do it via trial and error I don't really want to PR to someone else's work with stuff I don't fully understand or I just copypasted from elsewhere.
I am however starting to look into modding so hopefully I'll be more useful \o/

NillerMedDild:
Nice work Motoko, I appreciate the cleanup! 😄 

Totally understandable Merlo, I thought adding tags to BetterEnd would simply be submitting jsons :D Also, you're very useful already, but more modding power is always nice 😄

NillerMedDild (approved):

PR #1319, opened 2021-02-15
Tag Maintenance Pass - Clean up Tag Search Scripts

MuteTiefling:
These have been bugging me for a while, so I'm condensing them a bit and using regex to hopefully reduce the amount I need to touch them.


NillerMedDild:
Nice, glad to have things more neat.

PR #1323, opened 2021-02-16
Add Sulfuric Rock to stone generators

MuteTiefling:
Missed in the first pass.


NillerMedDild:
Looking at that config makes me reconsider specifying them as loot tables through KJS :P

MuteTiefling:
Yeah. It's getting a little out of hand

PR #1337, opened 2021-02-18
Compact Machines recipes

Merlo17:
I'm still not convinced on these since they might feel a bit too much and I'm probably looking over them again, but I'll still open the PR to get feedback from you guys as well.

The main idea is give three different routes to make walls centered around enderium/enderperals, giving both magic and tech alternatives to make them.

The actual machines then start with a 5x5x5 and tier up to a 13x13x13 for the max tier (how it was in the original mod) and the catalysts gate them a bit more, with the different tiers of IF black hole units or Occultism stabilizers

The main cost here is enderium, so let me drop a quick table to show you the costs:

| Tier  | #Enderium Blocks|
| ------------- | ------------- |
| Tiny  | 2 |
| Small  | 2 |
| Normal | 4 |
| Large | 8 |
| Giant | 17 |
| Maximum  | 31 |

the rampup is quite steep there at the end since I'm using enderium in the structure too which might change. In the end I still think this is a bit too extreme for non expert, so if you have any other ideas let me know

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/normal/recipetypes/mekanism/metallurgic_infusing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/normal/recipetypes/mekanism/metallurgic_infusing.js
@@ -1,0 +1,20 @@
+events.listen('recipes', (event) => {
+    if (global.packmode != 'normal') {
+        return;
+    }
+
+    var data = {
+        recipes: [
+            {
+                output: Item.of('compactmachines:wall', 32),
+                inputItem: '#forge:storage_blocks/ender',
+                infusionInput: 'mekanism:refined_obsidian',
+                infusionAmount: 160
+            }
+        ]
+    };
+
+    data.recipes.forEach((recipe) => {
+        event.recipes.mekanism.metallurgic_infusing(recipe.output, recipe.inputItem, recipe.infusionInput, recipe.infusionAmount)
+    });
+});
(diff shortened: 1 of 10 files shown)


NillerMedDild:
I think those Enderium Block counts are a bit wild. I think I'd rather have it be Ingots tbh.
Do the Tunnel recipes work?

But! Very interesting recipes, I like what you've done here.

Merlo17:
Tunnel recipes work yeah, they're just simple crafting recipes really. I can make them ingots and maybe put the count back to 32, with the higher ones still requiring some blocks (since they're in the structures). Ingots might fit less thematically but it's a sacrifice that's needed I'd say

NillerMedDild:
Sounds good to me :)

Merlo17:
like that?

kubejs/server_scripts/enigmatica/kubejs/normal/recipetypes/mekanism/metallurgic_infusing.js
```diff
@@ -6,10 +6,10 @@ events.listen('recipes', (event) => {
     var data = {
         recipes: [
             {
-                output: Item.of('compactmachines:wall', 64),
+                output: Item.of('compactmachines:wall', 32),
                 inputItem: '#forge:storage_blocks/ender',
```
NillerMedDild, line 10: So close, storage_blocks -> ingots

Merlo17: look again, those are ender blocks (4 pearls) not enderium. Mek uses refined obsidian so I felt like enderium and refined obsidian was a bit much. I think of it as an alternative gate per say

NillerMedDild: Ah of course, my bad :D

PR #1341, opened 2021-02-19
Added to Occultism wild hunt ritual loot table

theboo:
added a bit to the loot tables for this ritual as it is fairly expensive(gold block, silver block and copper block + a diamond and cant be fully automated) and the only perk to it before was guaranteed wither skulls which are pretty easy to get in this pack. There is a high chance of nether fortress loot(nothing special at all), a decent chance at a random enchanted book enchanted at lvl 20-30, a pretty rare chance at an apo tome at lvl 50, and a suuuuuuuuper rare chance at a random artifact(right around 5%), along with the guaranteed skull.


kubejs/data/occultism/loot_tables/entities/wild_hunt_wither_skeleton.json
```diff
@@ -0,0 +1,263 @@
...
+                            "function": "enchant_with_levels",
+                            "levels": 50,
+                            "treasure": true
+                        }
+                    ]
+                },
+                {
+                    "type": "loot_table",
+                    "weight": 20,
+                    "name": "minecraft:chests/nether_bridge",
```
ochotonida, line 239: Something to keep in mind is that Artifacts actually adds some items to this loot table, so the chance of getting an artifact will be higher than 1/49, and it will be biased in favor of the items artifacts adds to this loot table (see [artifacts:loot_tables/inject/chest/nether_bridge.json](https://github.com/ochotonida/artifacts/blob/1.16/src/generated/resources/data/artifacts/loot_tables/inject/chests/nether_bridge.json))

kubejs/data/occultism/loot_tables/entities/wild_hunt_wither_skeleton.json
```diff
@@ -0,0 +1,263 @@
...
+                }
+            ],
+            "entries": [
+                {
+                    "type": "item",
+                    "weight": 20,
+                    "name": "minecraft:book",
+                    "conditions": [
+                        {
+                            "condition": "killed_by_player"
```
ochotonida, line 86: Since all of the entries in this loot pool have this condition, it's probably better to move this to the conditions of the pool

theboo: Ahh yes that would make sense to move it to the pool itself... Tbh it's my first time messing with loot pools.. and yes, I see the artifacts in the nether fortress pool, but I'm not worried about it... The Occultism ritual is very expensive and fairly slow to activate(must be done manually) and summons 3 of these buffed wither skeletons each time

theboo:
totally forgot this ritual requires a player/villager sacrifice as well

NillerMedDild:
Did you verify that this is working in game Sir Boo? :)

theboo:
and yes, this is what the thunderdome is for haha

theboo:
oke much testing in the thunderdome later and i am happy with this, should be gucci

PR #1355, opened 2021-02-21
Adds elytra and dragon head to endcity loot

theboo:
Always complaints of all the elytra/dragon heads taken on servers. This adds elytra and dragon heads to the end city lootR chests, and re-adds the artifacts back in as their injections get yeeted when we replace the datapack. The only other mod that is injecting in to end city loot right now is sophisticated backpacks, i decided to leave that out as who needs a backpack when they are at end city loot hunting stage haha.

Each chest has the same loot as now(minus backpacks), but with a 10% chance at either an elytra OR a dragon head.

### DIFF
--- a/kubejs/data/minecraft/loot_tables/chests/end_city_treasure.json
+++ b/kubejs/data/minecraft/loot_tables/chests/end_city_treasure.json
@@ -1,0 +1,344 @@
+{
+    "pools": [
+        {
+            "rolls": {
+                "min": 2,
+                "max": 6
+            },
+            "entries": [
+                {
+                    "type": "item",
+                    "weight": 5,
+                    "name": "minecraft:diamond",
+                    "functions": [
+                        {
+                            "function": "set_count",
+                            "count": {
+                                "min": 2,
+                                "max": 7
+                            }
+                        }
+                    ]
+                },
+                {
+                    "type": "item",
+                    "weight": 10,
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Hmm, can't we inject into the loot table instead of replacing it?

theboo:
if you know a way without a separate mod, lmk... we tried haha

NillerMedDild:
I thought it was as simple as a certain namespace in the datapack tbh :)

theboo:
yeh most mods(like botania and artifacts) put it in loot_tables/inject/ but they also have code to actually inject it

NillerMedDild:
Damn :P

ochotonida:
Are you sure replacing the loot table actually prevents mods from injecting loot though? I might be wrong, but as far as I understand it loot is injected after the loot table is loaded, regardless of whether it was overwritten by a data pack or not.

theboo:
yeah, when i just added elytra and dragon heads to the vanilla end city loot, there wasnt a single artifact to be found. it seems the datapack loaded after the injections

NillerMedDild:
Hmm, we should probably ask Lat if this is a use case for KubeJS.

theboo:
already brought it up to max yesterday... he said he put it on the "high priority" list, whatever that means haha

PR #1371, opened 2021-02-22
Disable Better End Anvils

MuteTiefling:
Disable Better End Anvils. Make anvil crafting recipes available to standard anvils instead.


MuteTiefling:
https://github.com/Beethoven92/BetterEndForge/issues/67

PR #1376, opened 2021-02-23
Adds more mountain oregen

theboo:
Moved copper, zinc, tin and iron up to higher base level, and increased spread. they all still generate under sea level but the mountains are much less bare.  I didnt touch lapis as i wasnt sure if we wanted to make this mountain only or not, and moving it up there while keeping it down at lower levels would make for a gigantic spread. Also added astral marble to worldstripper config.

also, putting these ores up higher is gonna make for some large y-gaps in the ee ore info tabs... idk what to do about that tbh but it will be hard for people to target ores using it... I'm also totally okay with scrapping this idea alltogether, the ores look good in game with this new config IMO, but the jei info tab is kinda useless for some of them lol.


Ridanisaurus:
What do you mean by large Y-gaps in the Info Tab?

theboo:
answered rid in discord, but I mean things like coal having a y=0-180 range... I wonder if it would be nice to put the base level in the info tab as well? So they can see that it CAN generate at y=0-180 but also see that y=80 is the best level for it?

Ridanisaurus:
I have to disagree with you on this, sadly. 90% of people (if not more) don't know know what is a Baseline and what is the Spread. This is why I had to do the math on that Tab to convert them into MinY and MaxY so they can be more readable for players.
Regardless, at some point in the future, I'm reworking that info tab to have more information, and presented in a nicer way. Who knows, might even look into implementing a Graph, and when you hover over the peak it will show you the baseline ¯\_(ツ)_/¯

Trostmm:
Distribution graphs are the best! Fun to see the kurtosis from the weighting.

NillerMedDild (approved):

Issue #1377, opened 2021-02-23
Add groups to mods with foods not natively covered by Diet


Merlo17:
- [x] PNC
- [x] Resourceful Bees
- [x] ID
- [x] Better End
- [x] Undergarden
- [x] BYG
- [x] Ars Nouveau
- [x] Meet your fight

https://github.com/NillerMedDild/Enigmatica6/issues/1333

PR #1378, opened 2021-02-23
Added Diet food groups to unsupported mod foods

Merlo17:
The five scripts could easily be merged into a single one but I liked having them separate, felt a bit more tidy to me. If you think it should be merged I can do that quickly.

Also diet is really neat, just have to add the tag for the food group and it figures out the values by itself. I really like it :P

https://github.com/NillerMedDild/Enigmatica6/issues/1377

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/diet/fruits.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/diet/fruits.js
@@ -1,0 +1,24 @@
+var fruitFoods = [
+    'integrateddynamics:menril_berries',
+    'betterendforge:shadow_berry_raw',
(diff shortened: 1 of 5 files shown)


NillerMedDild:
Nice work :D I'm gonna change how the tags are added a little bit though, because you can simply add arrays.

TheIllusiveC4:
I'm not sure where to properly put this comment, but I'm the developer of Diet and I just wanted to give a heads-up that the most recent version, 1.16.5-0.9, now categorizes these particular modded foods natively. If the modpack updates to that version or higher, most of these can be removed. The exceptions are any custom Resourceful Bees foods, as Diet only uses the defaults.

NillerMedDild:
Oh cool, thank you :)

PR #1381, opened 2021-02-23
Tweaked diet effects

Merlo17:
- yeeted negative effect
- split up the effects a bit more to fit them thematically: more HP is with fruits veggies and carbs, attack strength and knockback is with proteins and attack speed is with sugar. 

More ideas?


NillerMedDild:
Not at the moment, but I'm sure they'll come 😄

PR #1414, opened 2021-02-25
Fixed recipes for thatch materials

theboo:
Byg thatch building materials are made from their reeds which are only found in one biome, dont generate frequently, and have no way to be grown by the player(besides maybe ars sylphs). Added reeds to growables.

Upgrade aquatic thatch is made from beach grass which does not seem to be generating in world right now, I added a tag to UA beach grass and byg beach grass(which is not used for any recipes currently). This way people can make the UA thatch building mats with byg beach grass. I went with a tag input in case UA beach grass is able to start generating in the future. 

I was not sure what tag to put this under as forge:grass seems to be used by the grass/dirt blocks, if anyone has a better idea for the tag feel free to change it lol. I went with an enigmatica:crafting_materials tag.


SilverRuin:
Out of curiosity, would it be possible to add Beach Grass from Project: Vibrant Journeys to the same tag group as the BYG and UA Beach Grass? Currently neither the BYG or UA Beach Grass spawns in world, and there are no crafting recipes that use P:VJ Beach Grass, meaning the Beachgrass Thatch from UA is uncraftable.

I didn't want to drop a new issue since this was apparently already a topic, but since this is from February I hope this doesn't qualify as a necro post (Still getting that hang of GitHub and submitting issues to help out).

NillerMedDild:
I'll look into it :) 

A new issue would be perfectly acceptable for this, but commenting here is just fine as well - I'll just create a new issue 😄

SilverRuin:
Awesome, thanks. I'll keep that in mind!

PR #1428, opened 2021-02-26
Lowered artifact rates in end city loot (and accidentally fixed #1391)

### LINKED ISSUE #1391: Missing recipe for Pumpkin Slice
**Modpack version**
0.4.0
**Describe the bug**
Using the knife (in my case, iron) with a large pumpkin slice from Autumnity does not allow you to make a Pumpkin slice.
If this is intended behavior, I apologize.

**Screenshots**

image

theboo:
I think the artifact's weights for these chests is based on the 1-3 loot chests you would normally find in an end city... These new end cities can have 10-20+ chests in them, and with lootr basically a whole server will have every single artifact from 1-2 end cities found. I moved the artifacts in to their own pool and gave them a 30% chance of generating putting it much closer to what a vanilla end city might have.

Edit: accidentally put the fix for #1391 here too, whoops

### DIFF
--- a/kubejs/data/minecraft/loot_tables/chests/end_city_treasure.json
+++ b/kubejs/data/minecraft/loot_tables/chests/end_city_treasure.json
@@ -335,13 +335,13 @@
             "conditions": [
                 {
                     "condition": "random_chance",
-                    "chance": 0.3
+                    "chance": 0.2
                 }
             ],
             "entries": [
                 {
                     "type": "loot_table",
-                    "weight": 3,
+                    "weight": 8,
                     "name": "artifacts:artifact"
                 },
                 {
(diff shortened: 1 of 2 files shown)


theboo:
whoops, didnt mean to put both of these in one... oh well

ochotonida:
I just noticed you originally added the artifacts to the main loot pool, while artifacts injects it's own loot pool. This only exacerbated the problem, because the main loot pool has multiple rolls (4 on average), increasing the average amount of artifacts per chest from 0.45 to almost 1.4.

These new numbers look mostly fine to me, although lowering the chance for an artifact even more (to around 0.15 perhaps) wouldn't hurt. Assuming there's around 4-10 times as much chests in these new end cities, you would still get ~130% to ~330% the amount of artifacts compared to vanilla end cities

theboo:
and now i understand why you had the weight at 40, makes much more sense

Rodg88:
Just anecdotal, but so far the artifact rates seem kinda low now. From 16 chests in a city (12 in main tower and 4 in ship), i got a single umbrella, and then just vanilla armor and ingots/gems.

PR #1447, opened 2021-02-27
Modded mob icons for FTB Chunks

Lykrast:
This adds icons on the FTB Chunks minimap for several modded mobs (but not all in the packs, notably no Alex Mobs). Should probably also PR that to the mod itself but it's 3 am right now.

Includes:

* Aquaculture (hides all the fishes and turtles like vanilla fishes are)
* Artifacts (hides Mimics so they're a lil bit more sneaky)
* Eidolon
* Meet Your Fight
* Quark (the Frog has exactly the right proportions to not look good on it)
* Thermal
* The Undergarden (something seems to block mobs showing in the dimension itself dunno why, but the icons work if like you put the min the Overworld)


NillerMedDild:
Thanks Lykrast :D

Lykrast:
Made a PR to FTB Chunks. https://github.com/FTBTeam/FTB-Chunks/pull/118

Lykrast:
The PR to FTB Chunks was merged so the resources files might not need to be here if it gets updated before the next pack update.

NillerMedDild:
Ah nice, I'll remove them :)

PR #1466, opened 2021-02-28
Add Autumnity/Atmospheric saplings to market

theboo:


NillerMedDild:
Nice, thanks!

Quick note, if you're using Visual Studio Code I recommend that you install the prettier plugin and configure it to format on save, that way our files always follow the same code style, more or less.

PR #1473, opened 2021-02-28
Added QoL log to chest recipes for all modded logs

Merlo17:
This should also add them for future logs if any mods add them and should give the correct chest variant if there is one. My only doubt was if we should move all of this to normal, since these probably don't exist in expert.


NillerMedDild:
Looks good! I don't plan to change the chest recipe this time around, too much can go wrong with it tbh.

PR #1474, opened 2021-02-28
Blacklist Artifacts campsites from Undergarden biomes

ochotonida:
Pretty self-explanatory 🙂
I updated the default config value for Artifacts not too long ago, but it seems like it got overlooked


NillerMedDild:
Thank you kindly!

PR #1489, opened 2021-03-01
Improved Traveling distance and angle for travel anchors

Buuz135:


NillerMedDild:
I'm not sure how making the angle less forgiving improves it?

Buuz135:
It makes things like this better when both are in range

javaw_quRo06RuWe

NillerMedDild:
Thanks! I changed the distance to 100 blocks, I think that's plenty.

Issue #1519, opened 2021-03-03
Blast Furnace Smelt Iron Dust at Half Speed/Half Efficiency


telemenar:
**Modpack version**
0.4.2

**Describe the bug**
Blast Furnace Smelt Iron Dust at Half Speed/Half Efficiency

The first recipe in priority for the blast furnace is a 10s recipe which results in the blast furnace smelting at normal speed, and also using double the normal fuel.

**Logs**
Input pastebin link to crash-report or `latest.log`, if applicable.

**Screenshots**
![2021-03-02_17 08 27](https://user-images.githubusercontent.com/1326160/109737025-31671d80-7b7a-11eb-9224-3b5869ed9abc.png)

NillerMedDild:
Thanks for the report, should be fixed in-dev :)

PR #1521, opened 2021-03-03
remove duplicate ore and dust smelting recipes

MuteTiefling:
Should fix https://github.com/NillerMedDild/Enigmatica6/issues/1519

hopefully I haven't overlooked anything, but so far all checks seem good.


NillerMedDild (approved):
Looks good to me :D

PR #1554, opened 2021-03-05
Adds Liquid Starlight to Powah coolants

OneLemonyBoi:


MuteTiefling:
On the topic, I think we need to buff the production rate of ic honey. Right now even 16 bees weren't enough to keep up with a single thermo generator.

MuteTiefling:
-20 also seems a bit much considering how easy starlight is to get.

OneLemonyBoi:
Starlight is historically one of the coldest substances in Modded MC. But it'd make sense to change that due to the fact that it takes so long to consume. Powah should have a config section to choose how fast the fluid is consumed.

theboo:
is fluid coolant even consumed at all?

MuteTiefling:
> Starlight is historically one of the coldest substances in Modded MC. But it'd make sense to change that due to the fact that it takes so long to consume. Powah should have a config section to choose how fast the fluid is consumed.

I've submitted a request to change the way coolant works. Right now reactors don't consume it and thermo gens use it very slowly. 

As for historical use... yeah, I know it's been that way since like.. Sevtech at least. But that doesn't necessarily mean it _has_ to be. I think I'd personally make it about even with i-c honey. Though that will mean that i-c honey will probably never be used... considering the extra work that goes into it vs making starlight.

NillerMedDild:
I'd say ramp up i-c honey to -12/-15 and set liquid starlight to -9 instead. That or have Liquid Starlight at ~-5

NillerMedDild:
I changed the Liquid Starlight value to -5, I hope that's alright with you @OneLemonyBoi?

OneLemonyBoi:
Works for me!

MuteTiefling:
Alright, the honey situation seems resolved. 2 bees in a basic tier 1 beehive are juuuust barely able to supply a single thermo gen. Would be better with a higher tier hive and, of course, the player can always use a couple more bees.

MuteTiefling:
Yep, left running over night and it works perfectly. I've accumulated a small surplus if i-c honey in the system while voiding all the excess energy and self powering the machines (rftools crafter, magma crucible, and basic crucible)

PR #1556, opened 2021-03-05
repurposed structures preset, with some structure changes

theboo:
made villages slightly more common, dungeon crawl dungeons a bit more common, astral temples slightly less common. Tried adjusting towers of the waild but thats gonna require a config change, these dont seem to affect it... this wont have to be changed if we decide to not use repurposed structures as they will be removed from config on first load without the mod installed

### DIFF
--- a/config/terraforged/presets/Enigmatica.json
+++ b/config/terraforged/presets/Enigmatica.json
@@ -31,29 +31,32 @@
     "stronghold": {
       "distance": 32,
       "spread": 3,
-      "count": 128
+      "count": 128,
+      "salt": 0,
+      "constrainToBiomes": false,
+      "disabled": false
     },
     "structures": {
       "astralsorcery:ancient_shrine": {
         "spacing": 32,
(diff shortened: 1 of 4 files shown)


Merlo17:
We should probably axe mineshafts since we have YUNGs no?

theboo:
yeah, i dont think they do anything anyway with yungs installed, but we can look at it

theboo:
this might need to wait until next breaking update yes? if so we can close it as i wouldnt set these until closer to update time

NillerMedDild:
Only the biome weights need to wait, I'm fairly certain we can alter structure generation in a patch version with no issues :)

theboo:
@NillerMedDild idk if it is possible to merge just part of a PR, but the towers of the wild config and new TF preset for repurposed structures can be added to dev with no issues

NillerMedDild:
It's not :/ I'll save a temp file and overwrite

Issue #1534, opened 2021-03-04
Compact Machine Walls are unbreakable in survival


Merlo17:
**Modpack version**: 0.4.2

reported here https://github.com/CompactMods/CompactMachines/issues/455

Merlo17:
v4.0.0-beta.2 should fix this

Jafner:
You can work around this in the meantime by using Ars Nouveau's spellbook with the break glyph.

PR #1557, opened 2021-03-05
Add Building Gadgets Patterns for Compact Crafting

Mendanbar:
- Adds Patterns for the 6 different machine sizes, as well as the max field projection setup (needed to craft all of them)


Mendanbar:
Hoping this eases some pain points while we wait for https://github.com/NillerMedDild/Enigmatica6/issues/1534 to be fixed.

NillerMedDild:
Did you verify these in-game? :)

Mendanbar:
Just checked in 0.4.3 and I'm seeing some interesting behavior.  Field projector paste works fine, but the only way I can get the machines to craft consistently is by leaving one machine wall out of the copy-paste, and then placing it by hand before throwing in the catalyst.  Something about dropping in the whole structure at once causes it to not be recognized. 🤷‍♂️ 

I can update the recipes to have one missing machine wall block, and include a readme to explain some of the quirks.  Would that be ok?

NillerMedDild:
Hmm, yeah that's fine for now, thank you :)

Mendanbar:
Updated.  It's working very consistently now in all of my tests using the steps in the readme.

TemperedFool:
I will snag these and mess with them with the Schematicannon. Would be interesting to see if it behaves a bit better, since it places the structure block by block. I'll submit a PR with them after.

NillerMedDild:
Nice work, thank you for your contribution :D

Issue #1573, opened 2021-03-07
Drawer quest improvement


2xMJet:
**Description**
Add Functional Framed Drawers in a row with common ones in drawer opening quest. 
And add some info about their crafts in quest description.

**Why would you like this added/changed?**
Not all the time there is vanilla wood around with BYG+Terraforged generation.

**Screenshots if applicable.**
I think that info can help workaround, there is no NBT-info stored in fresh-crafted Framed drawer
![image](https://user-images.githubusercontent.com/26283528/110231864-ef105880-7f4c-11eb-8d77-72933b1951eb.png)

And there is 3 NBT-tags after applying materials to it, i suppose one of them can mean functioning of drawer.
![image](https://user-images.githubusercontent.com/26283528/110231822-bec8ba00-7f4c-11eb-8be6-d5b0b800c1e5.png)

NillerMedDild:
I'm not sure what you're asking for to be honest. I understand that you would like a quest to describe Framed Drawers and how to craft them, but that's about it ^^'

2xMJet:
> I'm not sure what you're asking for to be honest. I understand that you would like a quest to describe Framed Drawers and how to craft them, but that's about it ^^'
Sorry for my poor english, its not my native.
I'm talking about this quest.
![image](https://user-images.githubusercontent.com/26283528/110232709-f5ed9a00-7f51-11eb-8bab-547e4a3db373.png)

First of all, i think it needs to accept not only Storage drawers mod drawers but framed drawers too.
 BUT Fresh crafted Framed drawer isnt functional.
From that i tought about functionality-detecting problem, since there is much and much combinations of materials for Framed drawers, you cant just put them all in detect-list.
But Fresh Crafted Framed drawers isnt functional before you apply materials to it, so i think for detecting functionality of drawer NBT-data can be used.

Merlo17:
What they are asking is to allow framed drawers for that task, since BYG woods can't craft any drawers and you have to make framed ones with sticks instead. I think the solution here is to make a recipe in our unification script that gives out an oak drawer with every non vanilla wood honestly

2xMJet:
Ye, thats what i meant, but walked in the wrong direction in my toughts

PR #1575, opened 2021-03-07
Added convenient recipes for drawers using every wood type

Merlo17:
Should fix https://github.com/NillerMedDild/Enigmatica6/issues/1573, although I feel like an info tab to explain how framed drawers work could still be useful


NillerMedDild:
Have you tested whether this overrides recipes for other vanilla wood type drawers? :)

NillerMedDild:
And I agree JEI info would be very useful here.

Merlo17:
They showed up in JEI, it shouldn't override because of that return at the beginning no?

NillerMedDild:
Ah right my bad, I didn't see that in the PR overview.

PR #1583, opened 2021-03-08
Aquaculture integration

ochotonida:
In this PR I've made the various changes mostly related to Aquaculture's fish, details below. (This is the first I've used KubeJS, or javascript for that matter, so if there's anything I can do better please let me know)
- I've added fish fillet cutting recipes to the Farmer's Delight cutting board for all relevant aquaculture fish. I considered removing the default shapeless recipe entirely, but it uses a custom serializer and manually adds the recipes to JEI. This means that even if the recipe is removed it would still be visible in JEI.  The recipes on the cutting board do provide some bone meal similar to the Farmer's Delight fish cutting recipes, to give some kind of bonus over the crafting recipe.
- The atlantic cod and pink salmon cutting board recipes give raw cod/salmon slices from Farmer's Delight instead of raw fish fillet
- The frog, shrooma and goldfish recipes have been moved to the cutting board and removed from the crafting table
- Frogs now craft into Quark's frog leg instead of Aquaculture's as they have more uses (Aquaculture's frog legs are now unobtainable afaik, ~~but I'm not sure how to remove them~~ and I've hidden them from JEI)
- Goldfish craft into a gold chunk instead of a gold nugget
- Brown shrooma can be used as an alternative for the brown mushroom colony in the stuffed pumpkin recipe from Farmer's Delight
- Added some tags to the raw fish fillet, cooked fish, atlantic cod and pink salmon to allow them to be used in more recipes
- Aquaculture's driftwood now acts as a log for recipes and now crafts into Upgrade Aquatic's driftwood planks instead of oak planks
- I've added additional knife tags to give more knives the same functionality as Farmer's Delight's knives, and I've removed the same tags from the unobtainable knives to stop them from showing up in recipes.


Merlo17:
Looking great :D to hide stuff from JEI go in client_scripts/item_modifiers/jei_hide_items.js btw ^^

ochotonida:
Good to know!

ochotonida:
Something that might need investigating is that the disabled knives from Aquaculture/Abnormal's Delight still show up in JEI.

Merlo17:
Are you working after a /reload? That borks JEI hidden stuff for some reason, I always try a fresh launch for those just to be sure. F3+T *might* also work

ochotonida:
It seems this isn't actually caused by this PR in the first place, they also show up in JEI in 0.4.3

Merlo17:
just noticed that as well, they should be hidden according to the script. I even tested spamming some console lines a bit and it runs those lines fine, so I'm not sure what the issue is there. I'll leave it to Niller :P

NillerMedDild:
Excellent work, thank you for your contribution!

I have 2 minor notes:
- The file `kubejs/base/tags/items/aquaculture/driftwood_logs.js ` should be placed in the folder `kubejs/base/tags/items/upgrade_aquatic` according to our conventions. I've gone ahead and moved it.
- When removing a file, it must either be:
    - Removed explicitly through the `serverconfigupdater-common.toml` config (I've done this for the removed file from this PR)
    -  Not removed and tagged for removal, by replacing it's contents with something like `// TODO: Remove in 0.5.0`. The reason for this is that CurseForge doesn't remove deleted files between updates, so removing them from dev can have unintended side effects.

NillerMedDild (approved):

ochotonida:
Thanks for letting me know! I'm not quite sure how I didn't notice I placed the tag in the wrong folder, I guess the mod names are a bit too similar 😄

PR #1609, opened 2021-03-09
Re-add missing Chunk to Ore Shapeless Recipes

MuteTiefling:
Not sure what happened to these, but they're gone on 0.4.3 on the server and also in dev. Re-adding.


NillerMedDild:
Hmm I removed them because our output unification was messing them up, I thought I re-added them though :s

Issue #1576, opened 2021-03-07
Unify Wheat Flour


alephic:
**Description**
Both Pneumaticcraft and Create add wheat flour, but only Pneumaticcraft flour is accepted in the sourdough recipe. It would be nice if Create flour would work in this recipe as well.

**Why would you like this added/changed?**
It seems unintuitive to have two incompatible types of flour in the game.

**Screenshots if applicable.**
N/A

NillerMedDild:
We'll get on that :)

Merlo17:
I would go as far as axing PNC flour all together tbh

ochotonida:
I just noticed pedestals also adds flour. While it doesn't show up in JEI it still has a recipe. 
This would also be a good opportunity to unify Create's and Farmer's Delight's doughs. Farmer's Delight's dough is only used for bread, while Create's also has the slime ball recipe. I'd prefer to keep the dough from Farmer's Delight instead of the dough from Create though, since it's edible and has (in my opinion) the better texture.

PR #1618, opened 2021-03-09
Flour and dough unification

Merlo17:
Unified flour and dough as per https://github.com/NillerMedDild/Enigmatica6/issues/1576.
Chose create flour as default because nice texture and farmer's delight dough because of additional functionality (edible).
Also added some new flour recipes for fun


NillerMedDild:
Nice work, thanks! Love the PNC recipes :P

Merlo17:
believe it or not they're actually vanilla pnc recipes for the flour I removed :P

NillerMedDild:
That's possibly even better :P I tip my hat to Desht then!

PR #1629, opened 2021-03-10
BYG nether enhancements v2

Merlo17:
Added tags to nether bricks and replaced their input to make them interchangeable (couldn't do it with red nether bricks and quark nether bricks - they have recipes that can only be done by them and manually overriding them would've been a bit of a headache, plus I'd be fine with considering them different variants).
Also added crushing recipes for raw quartz -> quartzite sand and create washing for sand -> quartz

we don't talk about the first PR, it never happened


NillerMedDild (commented):
Looks good!

PR #1630, opened 2021-03-10
Remove missing items from create:upright_on_belt

ochotonida:
I'm not entirely sure what happened here, but it seems like I pulled right at the wrong moment when I made my previous PR


MuteTiefling:
ah, good was just noticing this in dev. <3

PR #1659, opened 2021-03-12
Update squeezer.js

DarkAngel22579:
Cleaned up squeezer recipes to match the updated fermenter recipes.


NillerMedDild:
Looks good, thanks :D

If you're up for it, perhaps change it to a single json object, in the form:
```json
const recipes = [
{ inputs: ['...'], value: 20 }, { inputs: ['...'], value: 40 }
];
```
Best of both worlds then :D these values vary a bit more than the Fermenter ones, so it makes more sense to use a json structure over simple arrays.

NillerMedDild (approved):
Great work, thank you :)

DarkAngel22579:
Had to take a detour and learn how objects and arrays work in javascript, but I got there in the end :D

NillerMedDild:
Important detour :D

Issue #1662, opened 2021-03-12
Add low passive generation to Mekanism fusion


NillerMedDild:
Depending on how the config works, perhaps 100-150krf/t

MuteTiefling:
Aww, went a bit too far? 

Should be able to dial in just about any value you want with a bit of testing.

NillerMedDild:
Yeah it seems like it :( People are asking what's wrong on the Mekanism discord very frequently, so I think that means we need a slight adjustment. I still think the passive generation should be a fraction of the default though.

NillerMedDild:
Would be good with a warning-style tooltip, but I also think we need to bring back passive generation.

MuteTiefling:
PR ready for approval. I'll get the quest text re-re-modified in a bit too.

PR #1665, opened 2021-03-12
Mekanism Power Pass

MuteTiefling:
Re-introduces Passive energy from the Fusion reactor, though at a still greatly reduced rate over default. An Injection Rate of 99 settles in at about 116k FE/t without water cooling. With water cooling, it drops to around 35k FE/t.

Also backed out the steam changes a tiny bit as it has no impact on total energy production, just on the number of turbines needed to get that energy and your ability to cool fission reactors. Should be slightly easier now. 

#1662

### DIFF
--- a/config/Mekanism/generators.toml
+++ b/config/Mekanism/generators.toml
@@ -82,7 +82,7 @@
 	[generators.fusion_reactor]
 		#The fraction of the heat dissipated from the case that is converted to Joules.
 		#Range: 0.0 ~ 1.0
-		thermocoupleEfficiency = 0.007
+		thermocoupleEfficiency = 0.014
 		#The fraction fraction of heat from the casing that can be transfered to all sources that are not water. Will impact max heat, heat transfer to thermodynamic conductors, and power generation.
 		#Range: 0.001 ~ 1.0
 		casingThermalConductivity = 0.1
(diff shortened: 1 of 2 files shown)


config/Mekanism/generators.toml
```diff
@@ -82,7 +82,7 @@
 	[generators.fusion_reactor]
 		#The fraction of the heat dissipated from the case that is converted to Joules.
 		#Range: 0.0 ~ 1.0
-		thermocoupleEfficiency = 0
+		thermocoupleEfficiency = 0.007
```
NillerMedDild, line 85: Would a value of 0.014 translate to roughly double the values in the PR description?

MuteTiefling: ... maybe?

MuteTiefling: I'll test it here and let you know

MuteTiefling: Short answer, no.

Long answer.. it's going to take a few minutes for the heat to stabilize to get a good read.

MuteTiefling: Well, ok. I'm a dirty liar. It looked like it was going to grow faster than linearly after the initial adjustment, but it has stabilized off to pretty much exactly double the air cooled value with an efficiency of 0.014.

NillerMedDild: 😄 I would like the value to be `0.014` then, seems fair considering the injection rate I think. If I'm understanding correctly :P

MuteTiefling: And up she goes

NillerMedDild (commented):
Thanks!

PR #1679, opened 2021-03-13
Tag updates

ochotonida:
I've updated Create's `upright_on_belt`, `fan_transparent` and `fan_heaters` tags with additional blocks/items, and I've added the heat sources to Farmer's Delight's cooking pot as well


MuteTiefling:
Totally going to steal some of this to update PNC and Powah heat sources too. Looks good from what I've seen

TemperedFool:
Oooooh, Create's Fan quest needs a note that these tags exist if it doesn't already have it...

ochotonida:
JEI doesn't allow you to search for block tags I'm afraid, but maybe we could add them as item tags as well

TemperedFool:
Hmm, the redundancy is bothersome, but it would be really cool. With striking ignorance he asks: "how tricky is that actually?"

MuteTiefling:
Create also has auto compat for certain types of blocks. Like open doors, trap doors, and glass panes that aren't connected on the sides. Basically à lot of partial blocks.

NillerMedDild:
> Hmm, the redundancy is bothersome, but it would be really cool. With striking ignorance he asks: "how tricky is that actually?"

Not tricky, just annoying to maintain :P

NillerMedDild (approved):
Looks good to me!

PR #1686, opened 2021-03-13
Remove duplicate aquaculture recipes

ochotonida:
This PR removes the duplicate cutting board recipes added by Farmer's Delight Compats. 
I decided to keep the cutting recipes we already had, because they have bone meal as a secondary output. (Otherwise there wouldn't really be an incentive to cut fish on the cutting board instead of using the regular crafting recipe)

To be honest, I'm not entirely sure whether we really need Farmer's Delight Compats in the first place. From looking at the curseforge page, it seems this mod exclusively adds data files, most of which are already covered by the kubeJS scripts in the pack

(Also: no it's not a typo, for some reason FD Compats adds new recipes under the aquaculture namespace)


NillerMedDild:
I've gone ahead and removed Farmer's Delight Compats, your last PR covered the vast majority of its content as you also point out, thanks :D

Issue #764, opened 2021-01-05
EnderStorage & Refined Storage [External Storage Issue]


Octr:
**Modpack version**
0.3.2

**Issue**
When attaching external storage directly to enderstorage ender chests the Refined Storage system will lock up and you can no longer take or put in items trough the crafting grid / portable crafting grid etc. Seems like they do not like eachother, not sure if there is a fix or if they should block each other from interacting instead.

Darkere:
Known issue. I've made a PR that fixes it and that will probably end up in the next RS version.

PR #1688, opened 2021-03-14
Add extra hot and cold blocks to PNC and Powah

MuteTiefling:
Had to move the PNC block heat data packs. The old ones were in the wrong spot and doing nothing. Since they're doing nothing and won't cause any conflict, I don't think we need to put a placeholder file there. Up to you, however.

PNC appears to have some trouble with the blaze burner currently. Investigating with Desht here: #764. Leaving them in for now as they simply don't work. If this requires a change in PNC, I'd expect them to start working without modification. Otherwise, I'll alter them later.

Added a note to various Emerald Ices on how to acquire them in a renewable manner.


MuteTiefling:
This is going to need a rework. Most of these are now included in the mod and it's changed over to a recipe system.

PR #1705, opened 2021-03-15
Apiary Beelancing

TemperedFool:
Per discussion #1695 item #1.

Remove Nether Star requirement for T1 Apiary to encourage moving to it and reducing number of entities required for volume production.
Move from ~4:1 scaling to 2:1 scaling for T2-T4 Apiaries.

T1 Apiaries costs no longer encourage the player to go wide rather than upgrading. T4 cost is still expensive, but requires about 10% as much grinding, with nether star progression at 0:2:6:16 vs 1:5:25:100.

Old T1 Cost:
T1_old
New T1 Cost:
T1_new
Old T4 Cost:
T4_old
New T1 Cost:
T4_new


NillerMedDild:
Excellent change, thank you :)

Issue #1342, opened 2021-02-19
Recutting with stonecutter


Rodg88:
**Description**
Recipes for the stonecutter to enable alreay "cut" stones to be recut. Could be an oredict list on the input to allow any of the 1:1 stonecutter crafting results for the particular stone. Eg, allowing polished variants to be put into the stonecutter to re-craft them into bricks.

**Why would you like this added/changed?**
Would allow the stonecutter to act like chisel used to, being able to change between block patterns freely.

NillerMedDild:
Sure that seems reasonable, we'll have a look at setting that up :)

NillerMedDild:
After having looked into this a bit, I realize that this would be extremely time consuming, the is no fast or easy way to find all these recipes, and they would all have to be manually added. I'll gladly apply the change if you find the stones that should apply to though.

Rodg88:
Ok, I've looked through byg as well now, and if you were going to refactor the script anyway, here's the relevant blocks:

soapstone_cutables:
byg:soapstone
byg:polished_soapstone
byg:soapstone_bricks
byg:soapstone_pillar
byg:soapstone_tile

dacite_cutables:
byg:dacite
byg:dacite_bricks
byg:dacite_pillar
byg:dacite_tile

+ one way byg:dacite -> byg:dacite_cobblestone

+ one way byg:rockey_stone -> minecraft:cobblestone

red_rock_cutables:
byg:red_rock
byg:red_rock_bricks
byg:mossy_red_rock_bricks
byg:chiseled_red_rock_bricks
byg:cracked_red_rock_bricks

remove recipe ids: byg:(null/mossy_/cracked_)red_rock_(slab/wall/stairs)_from_stonecutting (they're all duplicates)

byg_scoria_cutables:
byg:scoria_stone
byg:scoria_pillar
byg:scoria_stonebricks

+ one way byg:scoria_stone -> byg:scoria_cobblestone

+ uncomment the recipes to swap between create and byg plain scoria I put in the original script if wanted.

byg_(black/white/blue/purple/pink)_sandstone_cutables:
byg:(black/white/blue/purple/pink)_(null/chiseled_/cut_/smooth_)sandstone

NillerMedDild:
Hmm it seems I won't have time to refactor the script in the near future, could you perhaps add the above blocks to the existing script @Rodg88? :)

Rodg88:
no problem!

PR #1727, opened 2021-03-17
Reverse-stonecutting

Rodg88:
Added ability to use the stonecutter to switch between different variants of stones, eg from bricks to pillars. Mod compatability with Quark, Create, (some)Embellishcraft, PNC, Botania, Ars Nouveau, and Blood Magic.


Rodg88:
Not auto-generated as only some of the block variants in a stone type I wanted to be used as a source, so it's hand-curated. Also I'm not a coder lol

Merlo17:
It looks pretty nice but I would probably use a foreach loop to clean it up a bit :P

Rodg88:
Yeah well that requires some knowledge of how to do them :P lol If you have examples I can try tidy it up a bit, at least for the bigger lists. 

I also noticed astral marble to quark marble conversion wasn’t working, so I added that in as simple recipes that just swap between the base blocks. I quite liked how it worked not having astral marble variants visible if I’m currently working on quark marble, but still being able to swap between them by going via the plain stone variant.

Merlo17:
I'd look at all of our other recipe examples :P the structure is pretty clear there I'd say. Tell me if you need anything ^^

Rodg88:
Looking at it, it seems a bit complicated and all new to me so would take a while, and not sure it would even save any lines of code. Due to the requirement of only specific stone variants being desired inputs, it'd still need a list of each item type I've picked for a loop to go through? I think it's partly why Niller said here it'd be a time consuming manual task to do this lol: https://github.com/NillerMedDild/Enigmatica6/issues/1342

I could separate out the recipe deletions, it just made it simpler to keep the deletions next to the additions for each stone type to keep track of the tweaks

NillerMedDild:
I don't mind going over it and converting it to loops and such :)

Issue #1722, opened 2021-03-17
[0.4.3] Multiservo Press, Gearworking Die and Gear Mold


HaxtonSale1:
**Modpack version**
0.4.3

**Describe the bug**
Gear Mold (IE) can't go into Multiservo press, and Gearworking Die (TE) doesn't appear on JEI, though it has a recipe
Given how Gearworking die is not on JEI and how TE press molds can be inserted into Multiservo press, I expected the Gear Mold would be able to be used in the Multiservo Press also

**Screenshots**
![2021-03-17_11 41 50](https://user-images.githubusercontent.com/13237198/111406523-cf7beb80-8715-11eb-9bb2-52f75e899dcf.png)

telemenar:
Oh, I bet this is hidden by the same script that hides the native to thermal gears to try to expose only the gears from EE

PR #1729, opened 2021-03-17
Unified dies & molds

ochotonida:
This PR adds tags to thermal's dies and immersive engineering's molds to allow them to be used interchangeably, fixing #1722.
*However*, there's currently an issue with immersive engineering's press recipes, preventing the usage of tags as the mold. Because of this I just went ahead and unified the dies/molds, removing the packing, unpacking and gear dies from thermal.

### DIFF
--- a/kubejs/client_scripts/constants.js
+++ b/kubejs/client_scripts/constants.js
@@ -177,7 +177,7 @@ var regexHide = [
     /emendatusenigmatica:\w+_basalt_ore/,
     /titanium:\w+_gear/,
     /thermal:\w+_dust/,
-    /thermal:\w+_gear/,
+    /thermal:\w+_gear$/,
     /thermal:\w+_ingot/,
     /thermal:\w+_nugget/,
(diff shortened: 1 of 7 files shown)


NillerMedDild:
Looks good! :D I'm going to un-disable the thermal ones and just hide them though, as the disabled tooltip they'd get might confuse people who already have the dies.

NillerMedDild:
The Metal Press seems to be missing compacting recipes, and Packing 2x2, 3x3 and unpacking dies are only craftable through the Engineer's Wokrbench. Is this intended? :)

ochotonida:
The compacting recipes from immersive engineering don't show in JEI, but they still work (I'm pretty sure it's supposed to be like that).
As for the missing recipes, I initially only had the thermal dies in the client side disabled items list. It seems that after adding them to the server side disabled items, their recipes get deleted before they get changed by the output replacement script.

NillerMedDild:
Ah of course, you need the recipes for that, my bad!

Issue #1694, opened 2021-03-14
[Suggestion] recipes for legendary varients of infinity tools


KrAzYGEEK32:
**Description**
Crafting recipe for legendary varients of infinity tools in a nucleosynthesizer with any tool plus 2 (arbitrary amount) buckets of antimatter and power.

**Why would you like this added/changed?**
Would be a nice endgame challenge and a worthy reward for producing anti matter.
**Screenshots if applicable.**

MuteTiefling:
Really like this idea. Only trouble is that I really doubt we'd be able to keep NBT, so enchants and things would be lost if they weren't removed first. Plenty of ways to do that, however.

TemperedFool:
Is it possible to require that an unenchanted Infinity tool is used? That way there's no way they can lose a bunch of enchants in the process.

NillerMedDild:
This will need a bit more work - The recipe isn't visible in JEI, and I wasn't able to make it show. We may need to add some JEI info about the recipe.

MuteTiefling:
Odd, it was showing for me as a separate item in jei.

PR #1739, opened 2021-03-19
Infinity tools update

MuteTiefling:
#1694

Removes the special tag, which enables patreon only options.

Adds a JEI description to help guide people.


MuteTiefling:
I think I see what you mean with JEI. It doesn't show up if you look at Recipes for the tools, likely because NBT data is different from the ones in JEI. But the recipes do show if you look at Uses on the tools.

PR #1749, opened 2021-03-20
Add Alex's Mobs to Torchmaster blacklists for Mega Torch and Dread Lamp

MuteTiefling:
Noticed on a stream last night that Centipedes keep spawning in areas protected by mega torches. Considering the problems we already had with mosquitos, I figured we should just add everything to the appropriate list.

This should do that.

### DIFF
--- a/config/torchmaster.toml
+++ b/config/torchmaster.toml
@@ -8,7 +8,7 @@
 	#Same as the mega torch block list override, just for the dread lamp
 	#Block squid: +minecraft:squid
 	#Allow pigs: -minecraft:pig
-	dreadLampEntityBlockListOverrides = ["alexsmobs:crimson_mosquito"]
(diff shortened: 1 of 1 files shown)


Mendanbar:
I haven't fully tested, but I suspect this might also be happening despite an AS Lucerna ritual on my server.  I've been running `forge entity list` in the console to keep an eye on active mobs in the world, and it seems like Alex's mobs have increased substantially in spawn count after I set up the ritual.  It's hard to say for sure, though, because I have a few loaded chunks, and the Lucerna ritual is only active in one of the regions being actively loaded.

NillerMedDild:
Lucerna only covers hostile mobs though, right?

TemperedFool:
I've noticed the same in my 0.3.13 world (haven't done too much in 0.4.x for realz) - those stupid centipedes are EVERYWHERE underground since Lucerna isn't blocking them.

Mendanbar:
Yeah I thought the cave centipede was hostile.  I also have a ton of monkeys around, but figured the Lucerna ritual didn't cover those.

Mendanbar:
I'm on the latest release (0.4.3) btw

NillerMedDild:
Seems like an Alex's Mobs issue, based on this: https://github.com/HellFirePvP/AstralSorcery/issues/1647#issuecomment-759951409

PR #1768, opened 2021-03-22
Add byg stones to the stonecutter tweaks

Rodg88:
Added byg stone to the stonecutter tweaks, including tidying up duplicate recipes. Also added some vanilla stone recipes


NillerMedDild:
Thanks!

PR #1776, opened 2021-03-22
Minor mold/die unification fixes

ochotonida:
This PR re-enables thermal's die recipes which were erroneously removed in 382a88d (see #1729), and changes the reward of the multiservo press quest to Immersive Engineering's gear mold


NillerMedDild:
Ah that's where I left those values ^^' thank you!

PR #1777, opened 2021-03-22
Add PNC essence compat to Pedestals and increase conversion rates

MuteTiefling:
The Aerial Interface converts 1 bucket of Essence to 50 xp points, so I've brought this up to parity with that.


NillerMedDild:
> Does this also adjust IF essence to 50, or just PNC?

Both :)

PR #1802, opened 2021-03-24
Fix Apiary Recipies

### LINKED ISSUE #1800: [0.4.5] Apiary honeycombs
Apiary doesn't acccept any comb except vaniila one.

TemperedFool:
😰 Fix my boo-boo and set the apiary recipes to accept any honeycomb block, not just vanilla. Fixes #1800.


NillerMedDild:
Thanks!

PR #1811, opened 2021-03-25
Pedestals Quests

TemperedFool:
This should be final updates to push Pedestals quests out. Mowmaster let me know that significant changes are coming, so I'm going to take a peek at the dev builds over there and start working on things. Meantime I think this is ready to go.

If/once you commit this, can you push it to the Dev server? Would be good to let everyone take a peek at their leisure.


NillerMedDild:
Will put it on the dev server now, thanks 👍

PR #1816, opened 2021-03-26
Batch spellchecking fixes to quest descriptions

ylou:
Handy script for reference (requires installing hunspell + the python nbtlib library)

```bash
#!/bin/sh
function main() {
  for f in $(find $1 -name "*snbt"); do
    echo "[== Checking SBNT file $f ==]"
    nbt -s $f --path "quests[].description[]" --pretty | hunspell -l -p ~/.hunspell_default | sort -u
  done
}
main "$@"
```


NillerMedDild (approved):
Nice, thank you :D

PR #1835, opened 2021-03-27
Restore Pedestal's cobble and stoneworks

MuteTiefling:


NillerMedDild:
I don't understand what's going on here but I'm guessing you fixed something I messed up, so thank you ^^'

MuteTiefling:
The cobblegen stuff for Thermal was using the same function name as the Pedestals one, effectively over-writing it. The actual pedestals function, therefore, never got called : D

PR #1836, opened 2021-03-27
Fix beam stonecutting recipes

ochotonida:
c82a575 broke the recipes for Decorative Blocks' beams, this should fix that


NillerMedDild:
Thank you :)

Issue #1820, opened 2021-03-26
[0.4.6] Stonecutter tweaks issues


Rodg88:
**Modpack version**
0.4.6
**Describe the bug**
Some minor issues with the newer stonecutter tweaks:
Stone pillars and panels from Builders Crafts & Additions, mackaws roofs, and stone brick stairs/slabs walls appear in cut lists for plain stone variants, eg plain granite or plain red rock will have stone versions of these blocks.

![2021-03-27_07 02 56](https://user-images.githubusercontent.com/2611674/112686692-7dcc2180-8eca-11eb-971f-76518e29c879.png)
![2021-03-27_07 02 43](https://user-images.githubusercontent.com/2611674/112686698-7efd4e80-8eca-11eb-8504-027d1ae0bb15.png)

You can swap between the different colours of botania quartz blocks. These require different resources to make, so this bypasses that.

![2021-03-27_07 03 55](https://user-images.githubusercontent.com/2611674/112686773-9b998680-8eca-11eb-946a-eda4a6b7bed5.png)

You can get a stone variant's slab/stairs/wall directly from the base stone, eg being able to get mossy granite stairs from plain granite. This may not be an issue, but I'd argue requiring the variant as the source to get the slabs/stairs/walls would be a nice way to do things to cut down on the large amount of clutter in some of the stonecutter lists. Since stonecutting is free and no durability, It would make sense to me at least that you first turn your granite to mossy granite, then you cut it into stairs. This wold make the main list for stonecutting the base blocks much smaller and easier to find what you're after. If it's a lot of work to split up these recipes though I wouldn't worry about it.
![2021-03-27_07 04 18](https://user-images.githubusercontent.com/2611674/112686853-b8ce5500-8eca-11eb-90a0-3e9ab7269c49.png)

ochotonida:
Turns out the issue with stone isn't actually caused by the stonecutter scripts, it seems this commit (https://github.com/NillerMedDild/Enigmatica6/commit/2d1f2485bf1aa1be5fe9e2ec9322cf1837fb1d75) got a little out of hand...
![image](https://user-images.githubusercontent.com/37985539/112691624-df19d380-8e7d-11eb-85d5-351228b8d393.png)

As for the other issues, I'll look into it (not entirely sure what's going on with the quartz tbh)

NillerMedDild:
[...]

PR #1841, opened 2021-03-28
Split up Masonry wall/slab stonecutting recipes

ochotonida:
Masonry adds all their wall/slab variants to all variants of a stone in the stonecutter (for example,  when you put granite in a stonecutter you can craft Granite Brick Walls, Granite Column Walls etc). This PR changes that such that you need a specific stone variant to craft the corresponding wall/slab, to reduce the amount of clutter in some of the stonecutting lists. (see #1820)


NillerMedDild:
Nice, thanks!

PR #1849, opened 2021-03-28
Move material substitution tags into it's own script and refactor

MuteTiefling:
Refactor to ease expansion into plates and rods, as well as making new combinations where needed.


NillerMedDild (approved):
Perfect, thank you :D

PR #1868, opened 2021-03-30
Buff IF Fluid Laser BM Life Essence Recipe

TemperedFool:
This seems like it should be a fun alternate way to generate Life Essence, and Buuz has given us the tools to make it so! This now generates 500mb per operation (vs 5mb before), but requires a Villager under it much like the Ether Gas recipe requires a Wither. In my testing this is on par with a modest multi-purpose T4 Blood Altar with a Well of Suffering and ~8 Runes of Sacrifice. It's not going to compete with a dedicated life-essence setup, but I don't really intend it to. Should be a fun way for tech-y players to get into BM.

Requires a decent setup for regenerating/healing the villager, as the drill will not operate if their health is under 5. This isn't hard - astral, BM, Ars, and other methods can do it - but it does mean it makes for a fun setup that isn't free. 

Would like to get it into BM players hands and find out if this is a little to stingy!


NillerMedDild:
Very cool!

PR #1881, opened 2021-03-31
Integrated Dynamics/Tunnels Nerf

TemperedFool:
This PR stems from a discussion with @MuteTiefling regarding Integrated Dynamics/Tunnels. These mods are super useful for their logistical capabilities, but they make it super easy to set up a transfer system for fluids and energy that blows away most other options - and can be interdimensionally wireless! - for the cost of trees. This aims to bring the max transfer rates down significantly, while still making Tunnels useful for their logistical integration with Dynamics. These changes are as follows:

1. Default new part update frequency of 10 ticks (up from 1)
2. Lower max network FE transfer rate to 16,384 rf/tick
3. Lower max network fluid transfer rate to 16,384 mb/tick
4. Add JEI info tabs and tool tips indicating new limits.

This change may/will be controversial, so no worries if you either want to reject it or table it for now. It's also worth noting that we will want to give users fair warning before this goes into effect - if they are using ID/IT with sinks for their Fission setups, this will at least take their system down, and at worst blow it up if they don't have an automatic safety shutdown.

### DIFF
--- a/config/integrateddynamics-common.toml
+++ b/config/integrateddynamics-common.toml
@@ -21,7 +21,7 @@
 		#When enabled, networks will stop ticking and values will not be shown and evaluated again. This can be used to fix crashing networks by temporarily enabling this option.
 		safeMode = false
 		#The maximum network energy transfer rate.
-		energyRateLimit = 2147483647
+		energyRateLimit = 16384
 		#How deep the recursion stack on an operator can become. This is to avoid game crashes when building things like the omega operator.
 		operatorRecursionLimit = 256
 		#The number of threads that the ingredient network observer can use.
@@ -97,7 +97,7 @@
 		#Priority list of mod id's when determining tag-based recipe outputs.
 		recipeTagOutputModPriorities = ["emendatusenigmatica", "minecraft", "immersiveengineering", "thermal", "create", "mekanism", "jaopca", "kubejs", "pneumaticcraft", "occultism", "industrialforegoing", "botania", "quark", "pedestals"]
 		#The default update frequency in ticks to use for new parts.
-		defaultPartUpdateFreq = 1
+		defaultPartUpdateFreq = 10
 		#The NBT tags that are not allowed to be read by displaying NBT tags or performing operations on them.
 		nbtTagBlacklist = []
 
--- a/config/integrateddynamics-common.toml
+++ b/config/integrateddynamics-common.toml
@@ -21,7 +21,7 @@
 		#When enabled, networks will stop ticking and values will not be shown and evaluated again. This can be used to fix crashing networks by temporarily enabling this option.
 		safeMode = false
 		#The maximum network energy transfer rate.
-		energyRateLimit = 16384
+		energyRateLimit = 65536
 		#How deep the recursion stack on an operator can become. This is to avoid game crashes when building things like the omega operator.
 		operatorRecursionLimit = 256
 		#The number of threads that the ingredient network observer can use.
@@ -162,4 +162,3 @@
 	[biome.meneglin]
 		#The weight of spawning in the overworld, 0 disables spawning.
 		spawnWeight = 5
-
--- a/config/integratedtunnels-common.toml
+++ b/config/integratedtunnels-common.toml
@@ -61,7 +61,7 @@
 
 	[core.general]
 		#The maximum network fluid transfer rate.
-		fluidRateLimit = 16384
+		fluidRateLimit = 65536
 		#If particles should be shown and sounds should be played when tunnels are interacting with the world.
 		worldInteractionEvents = true
 		#If an anonymous mod startup analytics request may be sent to our analytics service.
@@ -74,4 +74,3 @@
 		ejectItemsOnInconsistentSimulationMovement = true
 		#If the version checker should be enabled.
 		versionChecker = false
-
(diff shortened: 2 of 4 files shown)


Rodg88:
Is the 16k transfer per tick, or per update ie 10 ticks? I’m also not sure forcing it to update only every ten ticks is a great idea since even that can be too slow for some things, especially things that you’ve used ID specifically for its speed. Even just the fluid limit by itself might be enough of a nerf to start with. Or I’d maybe look into just making the parts cost more instead?

MuteTiefling:
Putting limits in place to bring them in line with other pipes is ultimately far easier than recipe changes to make them more expensive to craft. And limits would still be needed in any case. As they are, there's no reason to ever use any other item/fluid pipes in the back as these are both the cheapest and fastest.

It's Botany Pots all over again ^_^

NillerMedDild:
> Is the 16k transfer per tick, or per update ie 10 ticks? I’m also not sure forcing it to update only every ten ticks is a great idea since even that can be too slow for some things, especially things that you’ve used ID specifically for its speed.

Per tick, and nothing is forced to be every 10 ticks, it would just be the default going forward.

MuteTiefling:
Slower default is more of a sanity check too. For most things, you don't need it checking every tick as that's just far faster than any other machines even process. The extra checks are wasted and just contributing to lag.

Rodg88:
Ah so every 10 ticks would just be the new default, that makes sense

Rodg88:
And what’s the extract rate of mek ultimate fluid pipes? Should ID be able to keep up with them? I’d almost say it’d be worth them being on-par with them just so there’s no particular need to use mek pipes instead.

telemenar:
In general I like this and would also really love to see the parts need energy to work as another part of the balancing act. 

However I think that 16k limits are a bit harsh, even 64k is still pretty low at least for energy.
Powah Transmit speeds: Blazing 18k Niotic 40k Spirited 90k
Mekanism Power Speeds per connected cable: Basic 8k Advanced 128k Elite 1M
Mekanism Fluid Speeds Pump: Basic 0.5B, Advanced 1B, Elite 16B, Ultimate 64B
Mekanism Fluid Speeds Throughput per connected pipe: Basic 4B Advanced, 8B, Elite 64B, Ultimate 512B

The wireless transmission does require going to the End for Chorus so it isn’t “Just” Trees and I could easily see balancing that with adding a heavy RF cost for the connector parts. (Which I think is available in the config.)

TemperedFool:
I'll address a few things here:

The 10 tick default is just for new parts, and can still be adjusted down to run every tick. The thing with ID/IT is it's easy to destroy a server just because you don't think to adjust them up. If someone needs the speed, no reason not to crank it to operate every tick.

The limits are *per network*. So, exporters overall can only export 16k/t total, importers can only import 16k/t total. I would argue that since the pipes & tunnels just cost trees*, they should not be competitive with pipes and wires from Powah/Mekanism. Powah cables are complicated to make, and I think Mekanism requires steel?

Wireless transmission doesn't actually require going to the end - there is a recipe for liquid chorus that just requires Merril berries and ender pearls...

MuteTiefling:
> In general I like this and would also really love to see the parts need energy to work as another part of the balancing act.
> 
> However I think that 16k limits are a bit harsh, even 64k is still pretty low at least for energy.
> Powah Transmit speeds: Blazing 18k Niotic 40k Spirited 90k
> Mekanism Power Speeds per connected cable: Basic 8k Advanced 128k Elite 1M
> Mekanism Fluid Speeds Pump: Basic 0.5B, Advanced 1B, Elite 16B, Ultimate 64B
> Mekanism Fluid Speeds Throughput per connected pipe: Basic 4B Advanced, 8B, Elite 64B, Ultimate 512B
> 
> The wireless transmission does require going to the End for Chorus so it isn’t “Just” Trees and I could easily see balancing that with adding a heavy RF cost for the connector parts. (Which I think is available in the config.)

As a counter to this, Mek and Powah are both considerably more expensive to craft too. And Chorus is no longer required for dimensional stuff. There's proto chorus now.

TemperedFool:
I will also say what I would LIKE to be able to do is have a power cost to run the importers/exporters that scales with throughput, but this isn't currently possible. I've made a suggestion on the Integrated Tunnels issues page, but I don't imagine that will get implemented tomorrow.

telemenar:
Hmm, if this is total transfer per network rather than connector limits, you are basically saying please don’t ever use this for fluid or energy. Which is ok, there are a lot of other options in the pack, but that is basically what this feels like to me.

Rodg88:
> Hmm, if this is total transfer per network rather than connector limits, you are basically saying please don’t ever use this for fluid or energy. Which is ok, there are a lot of other options in the pack, but that is basically what this feels like to me.

I wouldn’t even say a _lot_ of other options, in 1.16 currently, there’s some options. Particularly for fluids. Maybe an xnet buff to go with it, since that takes “real” materials and power to run?

NillerMedDild:
> I would be for a boost to XNet transfer maximums for power/fluids if we have control over that.

We do have control over that, I'll have a go at boosting them and link it here.

TemperedFool:
If that commit did what it's supposed to do, new limits are 65,536. I'll be curious if there are non-fission-reactor setups that don't work with these limits; especially with XNet buffed.

NillerMedDild (approved):

boneskull:
It seems odd to hardcode a config value in the JEI descriptions, when someone could easily just change the value in the `.toml` file.

NillerMedDild:
It's the best we can do, it's not like we can check the config value :P If people change configs that's on them imo.

PR #1889, opened 2021-04-02
Refactoring emendatusenigmatica-common.toml for better navigation

w0rldx:
I take this config from newer version of mod and change all entries manual from pack version (develop branch). 
Only for better navigation in file and default layout for all entries.


NillerMedDild:
Thanks for doing this, but to be honest I'd rather not merge it. There are a lot of changes here and I don't want to validate them all.

Issue #1892, opened 2021-04-02
[0.4.8] Cannot obtain IE Wires


Master2Mind88:
**Modpack version**
Current Dev Version as on 2:30
**Describe the bug**
I found a bug with Immersive Engineering Quest Line and wondering if this should be this way for all the the wires and also one of the tools.

**Screenshots**

![](https://cdn.discordapp.com/attachments/524992559341371433/827649531280621568/unknown.png)

![](https://cdn.discordapp.com/attachments/524992559341371433/827649583400222800/unknown.png)

NillerMedDild:
Thanks for the report. I'm not sure what the issue is here, could you perhaps rephrase?

Master2Mind88:
So the only way to get the wire cutters and the wire bundles is to trade with the the villager that is in the picture. You can not make them at all.
There is also no way of getting insulated lv wire for the quest.

NillerMedDild:
Oh! I'll investigate asap.

PR #1905, opened 2021-04-04
Fix missing Wire Coil Recipes

MuteTiefling:
Removes superfluous recipe removal responsible for missing wire coil and wire cutter recipes.

#1892


NillerMedDild:
Thank you :)

Issue #1875, opened 2021-03-31
Recipe to re-dye a botany pot


Rodg88:
Botany pot + dye to change the colour of the pot would be a nice recipe, as you can't change them currently, but you can re-dye terracotta.

PR #1907, opened 2021-04-04
Make Botany Pots and Terracotta washable and Botany Pots re-dyeable and other improvements

MuteTiefling:
…nd other improvements

Implements recipes to dye any botany pot to any other color, as well as washing recipes to revert to the plain terracotta look.

#1875

Also expands existing Mek Injecting recipe to work with any color terracotta (converts to clay) and adds a TPP recipe to do the same.

Reduces processing time of Essence > Memory Essence in TPP. Adds reverse recipe. Same for Dissolution chamber.

As a note, the washables tag was implemented in order to prevent recipes such as "Plain Terracotta washing to Plain Terracotta". In particular, for Create's washing this seems like something that would be bad for TPS since it has to keep checking the Plain Terracotta and attempting to re-wash it so long as it remains in the wash stream.


MuteTiefling:
Requesting review due to odd tag structure. Lemme know if it needs changing.

kubejs/server_scripts/enigmatica/kubejs/base/recipes/shapeless.js
```diff
@@ -310,6 +322,30 @@ events.listen('recipes', (event) => {
...
+        otherHopperPots.push('botanypots:hopper_botany_pot');
+
+        event.shapeless(`botanypots:${color}_botany_pot`, [Ingredient.of(otherSimplePots), `#forge:dyes/${color}`]);
+
+        event.shapeless(`botanypots:hopper_${color}_botany_pot`, [
+            Ingredient.of(otherHopperPots),
+            `#forge:dyes/${color}`
+        ]);
+    });
+
```
NillerMedDild, line 348: I don't really understand what's going on here - Is the extra looping and logic to prevent crafting a pot into the same pot?

kubejs/server_scripts/enigmatica/kubejs/base/tags/items/enigmatica/washables.js
```diff
@@ -0,0 +1,15 @@
...
+    colors.forEach((color) => {
+        washableBotanyPots.push(`botanypots:${color}_botany_pot`);
+        washableHopperBotanyPots.push(`botanypots:hopper_${color}_botany_pot`);
+        washableTerracotta.push(`minecraft:${color}_terracotta`);
+    });
+
+    event.get('enigmatica:washables/simplebotanypots').add(washableBotanyPots);
+    event.get('enigmatica:washables/hopperbotanypots').add(washableHopperBotanyPots);
+    event.get('enigmatica:washables/terracotta').add(washableTerracotta);
+});
```
NillerMedDild, line 15: Using regex here would save us a good 3 loops.

kubejs/server_scripts/enigmatica/kubejs/base/tags/items/botanypots/botany_pots.js
```diff
@@ -0,0 +1,8 @@
+events.listen('item.tags', (event) => {
+    var simple_botany_pots = ['botanypots:botany_pot'];
+    colors.forEach((color) => {
+        simple_botany_pots.push(`botanypots:${color}_botany_pot`);
+    });
+
+    event.get('botanypots:simple_botany_pots').add(simple_botany_pots);
+});
```
NillerMedDild, line 8: More regex pls :D

NillerMedDild (changes requested):

kubejs/server_scripts/enigmatica/kubejs/base/recipes/shapeless.js
MuteTiefling: Yep. Though now that you mention it I think I can do this with an array filter instead of the loop.

NillerMedDild: I don't see the point to be honest :P

MuteTiefling: Mostly just to keep things tidy in game

NillerMedDild (approved):

NillerMedDild:
Good stuff, thank you :D

Issue #1906, opened 2021-04-04
No Recipe For Sap or Syrup Bottles


Tsilkani:
Version 0.4.8

The only recipes in JEI for Syrup Bottles are various ways of cooking Sap Bottles, and there is nothing in JEI for making Sap Bottles. Trying to use a Glass Bottle on an Arboreal Accumulator or tank of Sap doesn't work, and neither does Create's Spout or Thermal's Fluid Encapsulator.

NillerMedDild:
Thank you for the report, I will investigate the issue.

MuteTiefling:
Sap is obtained from the Arboreal Extractor on Maple and Dark Oak trees. The fluid can be sent to a Fluid Encapsulator to bottle it. 

Don't know if this is showing in 0.4.8 but it's effectively fixed next version :D

HaxtonSale1:
Team Abnormals' way of obtaining Sap is using a glass bottle on a stripped maple log, if it has different texture then you can get Sap. 
But yeah, Thermal way removes a significant tedium

MuteTiefling:
Yep. I've actually just added a JEI description to describe that process and am in the process of double checking some recipes surrounding sap, syrup, and tree extraction in general.

PR #1910, opened 2021-04-05
Tree Enhancements

MuteTiefling:
Add note to Sap and Autumnity Sappy logs on how to obtain sap bottles. #1906 This sap was already mostly unified with Thermal Sap.
Remove Autumnity Sap to Syrup processing in favor of Thermal's.
Add Thermal and PNC Recipes to complete the chain from Sap > Syrup > Sugar. Every bucket of Sap yields 4 sugar once fully processed.
Add Arboreal Extractor support for more trees. Thermal Latex may now be obtained from eucalyptus, mahogany, and baobab. Sap may be obtained from any Maple.
More fertilizers tagged for use by Industrial Foregoing machines.
Industrial Foregoing fertilizer now works in the Arboreal Extractor, slightly better than bonemeal, but worse than PhytoGro.
Tree Oil processing added to PneumaticCraft TPP.


Rodg88:
Minor comment, but eucalyptus probably shouldn’t make latex, it should refine to tree oil. You can make turpentine from it irl lol

MuteTiefling:
Hrm fair. None of them should, really. Latex only comes from rubber trees.

Was mostly interested in adding more routes to it.

Issue #1891, opened 2021-04-02
[0.4.8] Pedestals quests


Rodg88:
**Modpack version**
0.4.8
**Describe the bug**
Some small things I've noticed so far in pedestals quests:
In the current version of the pack, you can't freely swap between the linking tool and the reverse linking tool, the reverse tool has a separate recipe. The linking tool quest says you can swap between them. I'd almost suggest adding a swapping recipe instead of adjusting the quest text, as people may want to try the feel of one tool or the other, and that way you can do so freely.

And the cobblegen quest seems to imply it only generates a stack of cobble, doesn't mention that it keeps generating in an internal buffer. It also doesn't mention that to enchant the upgrade with silk touch, you need to use the XP anvil.

TemperedFool:
For some reason I thought I had enchanted the Cobble Gen with silk touch in a vanilla anvil - you've tested and it doesn't work?

Either way, good catches! @NillerMedDild, you can assign this to me. I assume we'll update to Pedestals 1.16.5 0.8s (or later) for the next pack update? If so I have some reworking to do anyway ^_^

Rodg88:
Yeah it definitely needs the XP anvil pedestal for silky gen

NillerMedDild:
> Either way, good catches! @NillerMedDild, you can assign this to me. I assume we'll update to Pedestals 1.16.5 0.8s (or later) for the next pack update? If so I have some reworking to do anyway ^_^

Roger! We will indeed update :)

TemperedFool:
#1937 should close this out once it's merged.

PR #1919, opened 2021-04-05
Pedestals Quests Updates

TemperedFool:
Addresses #1891, although I think I want to make a few more changes before we close the issue out. This should at least be enough to make the text factually correct.


TemperedFool:
My dev instance is currently broken so I haven't tested this in world, FYI - it literally just modifies 3 lines of description text, so I can't imagine how it could cause any issues ^_^

NillerMedDild (approved):

NillerMedDild:
Looks good to me :D

Issue #1916, opened 2021-04-05
Iron Jetpacks


NillerMedDild:
**CurseForge Link**
https://www.curseforge.com/minecraft/mc-mods/iron-jetpacks

**Mod description**
Extremely configurable Jetpack mod, investigating in stead of #1488 

**Why would you like the mod added?**
More flight options.

**Screenshots if applicable.**

MuteTiefling:
wow these are still so fast as to be nearly unusable for building. But at least that can be modified.

MuteTiefling:
Alright, been playing with them for a bit here and I think I've got some nice numbers to share. 

I've created our own progression. No side tiers, just straight upgrades.  Tier order and numbers are all easily changed, as I've made myself a spreadsheet to generate the JSON files. 

With this initial setup, I had a couple of goals: 
1. Make them costly to run and with a target flight time. This is, of course just on internal power. It's kind of expected that a player will either make a player transmitter from Powah or carry some form of portable storage. So the base tier would get about 2.5 minutes of actual flight time with no external source.
2. Make them feel... jumpy. That is to say, they are jet packs. they should move vertically pretty well. so vertical speed has been boosted a fair bit from default ranges. Even the lowest tier will help you quickly climb out of a hole or ravine.
3. Bring horizontal speeds down. By default they're silly fast, going 100+ meters per second at the top tier. I've instead aimed for roughly 32 m/s at the top end with an additional 30% sprint bonus at that tier. Lower tiers have considerably lower 'sprint bonuses' and are very inefficient while sprinting. Even at the top tier, sprinting will cost you. 

Having played around with them a bit in a test world, they feel pretty good. The early tiers would be pretty accessible to build early on, but difficult to power. They'd be for occasional use, effectively. Would be good incentive to upgrade power generation, I think. 

Numbers below:

| name      | disable | tier | color  | armorPoints | enchantability | craftingMaterial           | rarity | flight time (minutes) | capacity      | usage | speedVertical | accelVertical | target speed (m/s) | speedSideways | speedHoverDescend | speedHover | sprintSpeedMulti | sprintSpeedMultiVertical | sprintFuelMulti |
|-----------|---------|------|--------|-------------|----------------|----------------------------|--------|-----------------------|---------------|-------|---------------|---------------|--------------------|---------------|-------------------|------------|------------------|--------------------------|-----------------|
[...]

PR #1930, opened 2021-04-07
Iron Jetpacks Update

MuteTiefling:
Default recipes were super boring. Made my own. Powah and Thermal components now used heavily throughout recipes. Pattern follows a sort of refresh/upgrade cycle as it goes from Powah to Thermal Tier.

Most IronJetpacks components left unused, so disabled them. Only the jetpacks and thrusters remain in use.

Reworked tiers to follow as discussed in #1916, however started with Hardened tier of Powah rather than Basic.

Did my best with the colors... Lumium and Spirited came out super weird in particular.


NillerMedDild (approved):
Looks good to me, good work! Will have a look in-game at the recipes to get an overview, hard to visualize recipes in code :D

PR #1937, opened 2021-04-07
Update Pedestal Quests

### LINKED ISSUE #1891: [0.4.8] Pedestals quests

TemperedFool:
I think this should bring the quests up to date for 0.8s_hotfix_1 - this introduced a bunch of changes. Should close #1891.

### DIFF
--- a/config/ftbquests/quests/chapters/pedestals.snbt
+++ b/config/ftbquests/quests/chapters/pedestals.snbt
@@ -139,98 +139,28 @@
 			x: -2.5d,
 			y: -2.5d,
 			subtitle: "For Discriminating Consumers",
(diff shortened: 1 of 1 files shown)


NillerMedDild (approved):
Woot! \o/

Issue #1941, opened 2021-04-08
[Dev] Metal Press Gear from Immersive Engineering


Master2Mind88:
**Modpack version**
Dev as of 10:30 MST
**Describe the bug**

The metal gear from Immersive Engineering does not work with Thermal Expansion Multi Servo Press.
You can not create gears since the dye does not work with it and also can't create the gear work from Thermal Expansion.

**Logs**
Input pastebin link to crash-report or `latest.log`, if applicable.

**Screenshots**

https://clips.twitch.tv/DoubtfulBraveHerringWutFace-1Y4z5-5eicz1eqDQ

PR #1944, opened 2021-04-08
Gear unification fix

ochotonida:
A recent commit seems to have broken gear recipes in thermal's press (#1941), fixed that and a typo or two


NillerMedDild:
Thanks :P

PR #1945, opened 2021-04-08
RFTools Control Spawner Expansion

MuteTiefling:
This re-implements all of the RFTools Control Spawner 'recipes'.

The primary aim was to add compat for other mods such as Alex's mobs, Thermal, Autumnity, Undergarden, and Upgrade Aquatic. In doing so, I found it simpler to 'normalize' recipes a bit.

There was a certain logic to the original recipes where cost was roughly tied to the size of the creature and 'usefulness'. Villagers were far more expensive to spawn than Zombies, for example.

Since none of our other spawning methods makes this distinction, I normalized it for the sake of simplicity in the script.

This new system ties the cost simply to the size of the mob. In addition, new substrate categories are defined allowing for a little more diversity in the blocks that are required for spawning. Polar Bears now use snow or ice, for instance, instead of the old dirt/gravel/sand.

Notable Changes:
Wither and Ender Dragon spawning was made a bit more expensive in terms of FE, and the Ender Dragon's Catalyst was changed from an XP bottle to End Crystals.


NillerMedDild (approved):
Damn, this is extensive! Great work.

MuteTiefling:
are those toml.bak files from me? >.< sorry :(

MuteTiefling:
Not sure what's causing them. they keep sneaking in even after I clear them out of the commit

NillerMedDild:
Forge is reluctant to release configs while the game is running I think :(

Issue #1797, opened 2021-03-23
Zoology Challenge


CletusMAngo:
A challenge for the player to tame every tamable mob in the game such as Alexs mobs,Vanilla minecraft etc etc 
(doesnt effect gameplay and should be optional for the player)

PR #1954, opened 2021-04-09
Zoology challenges

ochotonida:
This PR adds two challenge quests, one for breeding and one for taming (#1797). Both are implemented via an advancement generated with kubejs, since that seemed like the simplest way to do it. Vanilla minecraft already has a breed-all-mobs-advancement so I overrode that one, I'm fairly certain progress already made towards vanilla mobs won't be affected.
I'm *think* I found all the relevant mobs, but there's always a chance I missed a few...

I wasn't entirely sure what to do with the folder structure for the advancements, feel free to move things around.

### DIFF
--- a/config/ftbquests/quests/chapters/challenges.snbt
+++ b/config/ftbquests/quests/chapters/challenges.snbt
@@ -559,6 +559,48 @@
 					}
 				}
 			}]
+		},
+		{
+			title: "Zoology",
+			x: 3.0d,
+			y: 0.5d,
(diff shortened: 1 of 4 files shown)


NillerMedDild:
Very very cool, nice work! Did you verify this to be working in-game, do the advancements work as expected?

Folder structure looks good, I would've done the same I think.

ochotonida:
Yes, I did quite a bit of testing for this one, I'm pretty certain everything should work XD
image

PR #1966, opened 2021-04-11
Fix Interactio script: move tag recipe to correct recipe type array.

sciguyryan:
This will fix an error I spotted in the KubeJS logs.

This relates to a recipe that allows for any item with the tag '#minecraft:logs' to be converted into the item 'upgrade_aquatic:driftwood_log' when the item interacts with water. It was incorrectly added to the `simpleItemRecipes` list (which species that the input is an item) rather than `simpleTagRecipes` list.

This patch merely moves the recipe to the correct type.


NillerMedDild (approved):
Awesome, thanks!

PR #1985, opened 2021-04-14
Minor refactoring of create mechanical crafting recipes.

sciguyryan:
A minor refactoring of Create's mechanical crafting recipes. As the create.mechanical_crafting method is being used here the entire method call structure can be simplified.

This new format yields the same results, as per the documentation here: https://www.curseforge.com/minecraft/mc-mods/kubejs-create

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/mechanical_crafting.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/mechanical_crafting.js
@@ -1,21 +1,13 @@
 events.listen('recipes', (event) => {
-    data = {
+    const data = {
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Ah nice, thanks! Did you verify that this works in-game?

NillerMedDild (approved):

sciguyryan:
> 
> 
> Ah nice, thanks! Did you verify that this works in-game?

Indeed. Here is a screenshot from the modified instance that I grabbed earlier:

Screenshot 2021-04-14 163002

Everything is working just fine. This change also brings the style into line with the other files that use the KubeJS Create methods, which I forgot to mention in my PR.

Issue #1692, opened 2021-03-14
Make Tetra more usefull.


2xMJet:
**Before you start reading this, i want to say:**

* This is general concepts. Ideas to think of. Not a "please, add it now!!" and not essential for most of modpack players things.
* This enhancements can be done for expert mode, as it seems too much for common E6.
* This changes can completely revamp difficulty curve of the modpack.
* This will be hard to implement. I'm sure of it. But it can lead to something good imo.

**Description**
* **TOOLS**

I think modpack needs complete tool rebalancing around Tetra tools to make this mod more preferable in most cases of tool chosing.

For example in some other modpacks, other mods-own tools and vanila ones is 1 durability and only for crafting.
I've seen that thing realised with Tinkers construct, but i think it can work with Tetra too (with some balance changes to materials and their characteristics ofc, and with adding harder recipes for other modpacks OP tools)

Or maybe just rebalance Tetra materials at least.

* **(Expert)Other mods crafts**

If its even possible.
Maybe add custom recipes of some things on Tetra benches and forges to make them wrapped around Tetra, To make progress in modpack dependable on Tetra progression.

* **Use Tetra in Further Updates**
 
Using Tetra Hammers to obtain Dense ores from veins, for example. Or for prospecting. (not sure if its possible)

**Why would you like this added/changed?**

* Tetra requires much effort to do not so good tools.

* Tetra have almost no use in comparison to other modpack's mods tools, even vanilla ones is not more worse than their Tetra analogues. Not even mentioning other imbalance tools such as bores, hammers, excavators, disassemblers, PAXELs, AIOTS etc. 

* With Apotheosis, Tetra tools is more useless, since they cant be enchanted as good as any common tools.

* I cant see why i can use Tetra, except doing quests, after getting to first good tool. (in my case its Ref. Obsidian PAXEL from quest reward)

P.S Sorry for bad english, if i'm speaking not clear enough in some cases. Working on improving it.

Merlo17:
There are a few things that concern be about focusing everything around Tetra

- for starters, players just don't seem to enjoy it that much. now clearly I can't prove this universally and it's still a matter of taste, but from interactions on discord it seems the majority of people (myself included) finds it clunky to use and overall just not really enjoyable. you could point to a number of different things that make the mod feel worse, and maybe one of them is tools not having enough functionality to make them worth it, but I don't think that's the only factor.
[...]

PR #1996, opened 2021-04-16
Tetra Scriptification

MuteTiefling:
Moves Tetra materials and replacements into scripts where they can be more easily manipulated as a first step towards #1692

This also brings in all base Tetra materials, allowing them to be modified more simply as well. Left some notes for future material improvements in comments.

As proof of concept, all materials have been granted a 20% bonus on durability and magic capacity. All seems to be working well.

For simpler management in the future, we may want to consider setting up set 'tiers' for each stat. This way we could say, for instance, that Iron is Tier 3 durability and Tier 2 Speed, while Gold is Tier 1 durability and Tier 3 speed, as a rough example. This way we can have a central place to re-balance all tools against each other by simply tweaking the values for the tiers themselves.

I've opted to delete the JSONs in this PR even though they'll likely remain on a player's instance. The JSON injection is over-writing those same files, however, so it  makes no difference. I was seeing the changes from the scripts even with them in place. So this is just a bit of cleanup.


NillerMedDild:
Awesome work! 👍

PR #2000, opened 2021-04-16
Bees!

TemperedFool:
This is a fairly big PR, making progress on #1695. This is focused on bee breeding/spawning and moving comb outputs as much as possible to things that need processing like EE chunks and dusts. This was all working prior to pulling the most recent dev changes this afternoon; currently trying to sort out my dev instance after updating as it CTD's. I would recommend updating to ResourcefulBees-1.16.5-0.6.7b if possible, since it should resolve the "Bees clump around 1 flower" behavior.

1. Rearranged how the bees are stored - they're now stored in folders by bee type - Alloy, Gem, Material, Metal, Natural, Nether. We'll get an Ender folder if we end up with nether-related bred bees. I found I was trying to modify a group together - say, to adjust spawn weights - and this arrangement is making those kinds of changes easier.
2. Moved all bee-specific biome entries to their own folder to reduce clutter a bit. I think these are likely redundant and can be removed later, but they can live here for now.
3. I boosted production for some things, but mostly converted over to EE chunks or dust for everything possible. Production balancing can come after the breeding tree is finished.
4. I removed the Apiary Output specs from all the bees - this is an override for the common.toml configs, so without it here it will be a little easier to raise or lower bee output in general (while still allowing us to boost/nerf anything specifically problematic).
5. Significantly lowered breeding costs across the board.
6. I implemented the beginning of a breeding tree, removing a lot of bees from world-gen. I'm currently using block -> item to generate spawn eggs for alloy bees (will move this to block -> entity when they implement it), and entity -> entity for Nether bees T3-5.
7. I've upped the spawn weights to 12, made them spawn at any light level (in forests is tough with a "daytime" light level requirement), and added more biome tags to the natural bees. I removed the use of bee-specific dictionaries - things like the BooBee call out biomes specifically in their configs
8. Added Atmospheric, Autumnity, BYG, and Terraforged biomes to the biome tag dictionary.
9. Make BooBee much rarer - spawn weight 1 - and specific to the same biomes as a Ghast.
10. Bee Faces for the  FTB Chunks Map! My terrible attempt at changing the color of the vanilla Bee face. Chose blue coloring, which makes them much easier to spot on the map. @Ridanisaurus is more than welcome to make them look better.


NillerMedDild:
Good work, it looks nice!

I'd like to request a few naming scheme changes:

- `/bees/gem bees/Diamond.json` -> `/bees/gem/Diamond.json`
    - The same applies for all the variant folders, please remove ` bees`
- All `-` in file paths should be replaced with `_`
    - Example: `biome_dictionary/bee-specific/pigman.json` -> `biome_dictionary/bee_specific/pigman.json`

NillerMedDild (changes requested):
Good work, it looks nice!

I'd like to request a few naming scheme changes:

- `/bees/gem bees/Diamond.json` -> `/bees/gem/Diamond.json`
    - The same applies for all the variant folders, please remove ` bees`
- All `-` in file paths should be replaced with `_`
    - Example: `biome_dictionary/bee-specific/pigman.json` -> `biome_dictionary/bee_specific/pigman.json`

TemperedFool:
Done! I think that should cover it?

NillerMedDild:
Very close, `bees_specific` -> `bee_specific` please 😄

NillerMedDild (approved):

TemperedFool:
@NillerMedDild - it's also worth noting in the change logs that more changes to exact parentage are coming - this was a first pass, and given Rid's encouragement (it's as if he wants more texture work), I will likely make a few more bees and shift exactly who is breeding who around as I make them. My plan is to have a flow chart of some sort once things settle down, time allowing.

PR #2010, opened 2021-04-19
Botania Orechid Rebalance Attempt

MuteTiefling:
Having a bit more trouble with this than expected. It's getting late so I think I'm maybe just looking at it cross-eyed.

The scripts do not appear to overwrite the default values, however. I've tried purely with the script as written (with the key component uncommented, of course) and I've tried a straight datapack injection in the data folder. Neither worked :(

In both cases, I'm still getting ruby/sapphire and no fluorite is being generated.

Care taking a look to see if I've missed something?


NillerMedDild (commented):
It does work :D
I removed the datapack recipe override, the script is doing the job.

MuteTiefling:
... weird! I restarted the instance sooo many times last night and it refused to give any fluorite and kept dumping me with silly thermal gems >.<

Well, great then!

NillerMedDild:
Hmm, there are still Thermal Gems in there :/ but also Fluorite! Seems like the file isn't being overriden, just added to.

MuteTiefling:
Are you only seeing Stone based ores? Or still a mix. That was part of what I was hoping to fix with this too. Since it only converts stone, I wanted it to only output the basic ore, not all the pretty variants.

NillerMedDild:
All pretty variants too, I'mma make a bug report.

PR #2022, opened 2021-04-20
Sync generatableCobblestone and generatableStone lists between client…

sciguyryan:
… and server scripts.

As it stands there are items in the server list that are not present in the client scripts, meaning that the JEI descriptions are not added for them and so on.

It is unfortunate that there is currently no way to create a central list and be able to refer to that from both the client and server scripts. :(


NillerMedDild:
Thank you!

> It is unfortunate that there is currently no way to create a central list and be able to refer to that from both the client and server scripts. :(

Yes, very much :(

PR #2042, opened 2021-04-23
Remove the need for the Ars Nouveau glyph recipe, use a not filter in…

sciguyryan:
…stead.

The Ars Nouveau glyph recipe was added because the nether brick consolidation replaced nether bricks with a tag, which the glyph recipes do not support. This is now superfluous as KubeJS now supports a "not" filter on recipe types, meaning that the recipe type can simply be excluded from consolidation instead.

I also slightly reworked the glyph recipes script, just in case it is ever used in the future.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/ars_nouveau/glyph_recipe.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/ars_nouveau/glyph_recipe.js
@@ -1,28 +1,17 @@
 events.listen('recipes', (event) => {
     const recipes = [
-        {
-            tier: 'TWO',
-            input: 'minecraft:nether_brick',
-            output: 'ars_nouveau:glyph_dampen',
-            id: 'ars_nouveau:glyph_dampen'
-        }
     ];
 
     recipes.forEach((recipe) => {
-        recipe.id
-            ? event
-                  .custom({
-                      type: 'ars_nouveau:glyph_recipe',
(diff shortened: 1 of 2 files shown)


NillerMedDild:
that not filter is glorious, thank you ❤️

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/ars_nouveau/glyph_recipe.js
```diff
@@ -1,28 +1,17 @@
...
-                      output: recipe.output
-                  })
-                  .id(recipe.id)
-            : event.custom({
-                  type: 'ars_nouveau:glyph_recipe',
-                  tier: recipe.tier,
-                  input: recipe.input,
-                  output: recipe.output
-              });
+        const re =
```
NillerMedDild, line 6: I can't accept this part :P It's shorter I know but `const re` pains me to read.

NillerMedDild: I'll just change it myself actually.

sciguyryan: Really? That syntax is already used in quite a few places (e.g. in crushing.js) and it is considerably less clunky and easier to read than the original.

Well, your code base, your rules! 😅

PR #2095, opened 2021-05-04
Preliminary Blood Magic and Occultism Progression

MuteTiefling:
Reworked ID removals to use ID replacements where valid. This requires an update to KubeJS.

Eidolon will open up the Blood Altar, Alchemy Table, and Low Grade rituals in Occultism, opening up some extra crafting mechanics. Blood Altar Tier 2 and up will require parallel progress in Occultism.

Petal Apothecary can make Sprouted Fungus, making a simpler way to create potions. I'm thinking enchanting should be pretty late game, considering the power level. We have a lot of really fancy potions and a cool new system with Ars Nouveau for making some very powerful brews, but enchants almost entirely invalidate potions. By making them easier and enchanting later, I'd hope to get people to use them a bit more.

Alchemy Table recipe for Nocturnal Powder and Illumination powder added. Nice reagents for other crafting, and very handy when searching for mobs early on.

Cutting Fluid made a bit easier, but more varied. Now requires snake rattles (farmable), apatite dust (goes a long way), charcoal dust, and plant oil. This should hopefully give magic players a decently automatable ore doubling system.

Considering boosting output for ores, considering Create's Milling is also fairly early game and considerably less effort.

Otherstone now crafted from Arcane Stone

Leather Stripe now a cutting board recipe

Spirit Attuned gem now made from Shadow Gem

New Recipe for Divination Rod

Impure White chalk moved to Alchemy Table

White Chalk moved to Blood Altar

Occultism Bowl recipes moved to Enchanting Apparatus

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
@@ -1,4 +1,7 @@
 events.listen('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
     data = {
         recipes: [
             {
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
@@ -1,0 +1,38 @@
+events.listen('recipes', (event) => {
+    data = {
+        recipes: [
+            {
+                input: 'eidolon:unholy_symbol',
+                output: 'bloodmagic:weakbloodorb',
+                syphon: 7000,
+                ticks: 200,
+                altarLevel: 0,
+                consumptionRate: 5,
+                drainRate: 1,
+                id: 'bloodmagic:altar/weakbloodorb'
+            },
+            {
+                input: 'occultism:chalk_white_impure',
+                output: 'occultism:chalk_white',
+                syphon: 1000,
+                ticks: 200,
+                altarLevel: 0,
+                consumptionRate: 5,
+                drainRate: 1,
+                id: 'occultism:spirit_fire/chalk_white'
+            }
+        ]
+    };
+
+    data.recipes.forEach((recipe) => {
+        const re = event.recipes.bloodmagic
+            .altar(recipe.output, recipe.input)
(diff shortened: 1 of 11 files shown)


sciguyryan:
Just a heads up, you missed the ```if (global.isExpertMode == false) {``` check on the altar.js and cutting.js files.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
```diff
@@ -0,0 +1,38 @@
+events.listen('recipes', (event) => {
+    data = {
```
sciguyryan, line 2: You probably want to add the ```if (global.isExpertMode == false) {``` escape condition here.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/farmersdelight/cutting.js
```diff
@@ -0,0 +1,29 @@
...
+
+function filletRecipe(fish, filletCount) {
+    return cuttingRecipe(Ingredient.of(fish), Ingredient.of('#forge:tools/knives'), [
+        Item.of('aquaculture:fish_fillet_raw', filletCount),
+        Item.of('minecraft:bone_meal', Math.ceil(filletCount / 3))
+    ]);
+}
+
+events.listen('recipes', (event) => {
+    var data = {
```
sciguyryan, line 18: Ditto here.

MuteTiefling: Good catch, thanks!

NillerMedDild:
> I'm thinking enchanting should be pretty late game, considering the power level. We have a lot of really fancy potions and a cool new system with Ars Nouveau for making some very powerful brews, but enchants almost entirely invalidate potions. By making them easier and enchanting later, I'd hope to get people to use them a bit more.

I'll keep that in mind 👍 

> Considering boosting output for ores, considering Create's Milling is also fairly early game and considerably less effort.

I think that would be a good choice, I always felt like it wasn't worth it for 2x

NillerMedDild (approved):

MuteTiefling:
So vanilla BM currently does 2x and 3x. Shall we bump it to 3x and 4x?

PR #2144, opened 2021-05-12
Add missing break statement for 'avocado' wood variant case.

sciguyryan:
Add missing break statement for 'avocado' wood variant case to prevent case fall-through. While the default case is not currently used, this will prevent it being triggered for 'avocado' should that ever change.


NillerMedDild:
Thank you kindly!

PR #2145, opened 2021-05-12
Minor alphabetization and sorting for mod list

GregariousJB:
A minor tweak of just sorting this list alphabetically and adding in a few missing names. Used the TextFX plugin for Notepad++ to sort. A few entries had version numbers before the name, so I moved them to the end after the name to allow for easier sorting. A few other entries had no name at all, just a version number, so I visited the link to see what they were called and manually filled in the names. 

It's meant mostly as an introduction, although if this list is autogenerated somehow I may have already overstayed my welcome :)

Unrelated to this PR, I noticed a minor thing about quite a few quest rewards in-game - lots of them don't have titles, which means the user can't see what they are when mousing over them. Unless this is already being worked on, I thought I might take a stab at just adding the titles only without touching anything else about the quests. Is this something that's wanted?


NillerMedDild:
I'm afraid the modlist is autogenerated, the displayName is taken from the filename, which is why the name is off from time to time. I do appreciate you fixing it, would this will break again next update.

> Unrelated to this PR, I noticed a minor thing about quite a few quest rewards in-game - lots of them don't have titles, which means the user can't see what they are when mousing over them. Unless this is already being worked on, I thought I might take a stab at just adding the titles only without touching anything else about the quests. Is this something that's wanted?

Yes that would be very welcome :D

PR #2147, opened 2021-05-12
Bee Comb Recipes

TemperedFool:
Added recipes to allow conversion of Rocky, Forest, and RGBee combs into relevant products - stone, wood, and dye respectively. This is an initial stab at numbers balance-wise - we can clearly turn things down if they feel too generous, and I don't think they'll be too stingy.

I also actually committed the bee faces for Forest and Rocky for FTBChunks mapping this time...

Out of curiosity, would you generally prefer slightly more frequent but smaller PR's like this, or should I try to roll more together at once?


NillerMedDild (approved):

NillerMedDild:
> Out of curiosity, would you generally prefer slightly more frequent but smaller PR's like this, or should I try to roll more together at once?

I prefer small bite sized PRs 😄

TemperedFool:
Roger that, I'll try to keep them modest ^_^

PR #2148, opened 2021-05-12
Add Apotheosis Fletching recipes for Archer's Paradox and Alex's Mobs arrows

pm065:
I don't think this is already a thing? I didn't see anything about it in the files or issues/PRs. They might not be entirely balanced, but I tried my best. Honestly, I'm half inclined to just disable the crafting table recipes for these in their entirety, bringing them in line with the Apotheosis ones, but that feels a _little_ experty.


NillerMedDild (approved):
Good stuff, thank you!

I hope the extra arrows will be enough to convince players to make the Fletching Table instead of going with regularly crafting 😄

pm065:
These aren't all the arrows, though, since I couldn't figure out how to fit the recipes with more than three ingredients into three slots without a complete loss of balance.

pm065:
Oh, oops! I made a mistake with the shark tooth arrow; the fletching recipe gives 5, while the crafting one gives 6. One moment...

NillerMedDild:
I don't mind correcting that, fast fix 😄

NillerMedDild:
Oh please submit PRs to the Develop branch in the future btw :)

PR #2153, opened 2021-05-13
Added missing titles to quest rewards

GregariousJB:
Only added _reward titles_, nothing else. This allows mouseover indication of item name:

image

I also discovered a bug with FTB Quests - _single item_ rewards will not show mouseover names. Only item rewards with a quantity greater than 1, Enchanted Books, and those loot boxes will show up when moused over. I've added a bug report to the FTB Quests repo about this here, including a short video showing the issue: https://github.com/FTBTeam/FTB-Quests/issues/582


GregariousJB:
Someone snuck in an update before me. Fixing this and re-pulling

PR #2158, opened 2021-05-13
Added missing titles to quest rewards

GregariousJB:
Only added _reward titles_, nothing else. This allows mouseover indication of item name:

image

I also discovered a bug with FTB Quests - _single item_ rewards will not show mouseover names. Only item rewards with a quantity greater than 1, Enchanted Books, and those loot boxes will show up when moused over. I've added a bug report to the FTB Quests repo about this here, including a short video showing the issue: https://github.com/FTBTeam/FTB-Quests/issues/582

3rd time's the charm? :)

### DIFF
--- a/config/ftbquests/quests/chapters/botania.snbt
+++ b/config/ftbquests/quests/chapters/botania.snbt
@@ -28,6 +28,7 @@
 			rewards: [{
 				id: "0000000000000504",
 				type: "item",
+				title: "Pestle and Mortar",
 				item: "botania:pestle_and_mortar"
 			}]
 		},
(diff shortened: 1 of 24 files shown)


config/ftbquests/quests/chapters/botania.snbt
```diff
@@ -410,12 +414,18 @@
 				{
 					id: "0000000000000523",
 					type: "item",
+					title: "Open Crate",
 					item: "botania:open_crate"
 				},
 				{
 					id: "00000000000005E8",
 					type: "item",
-					item: "botania:monocle"
```
GregariousJB, line 418: Only section that looks odd to me is this one. I don't recall adding these other lines, just the title. Perhaps this is OK and FTB Quests simply added this data in automagically?

GregariousJB: Github isn't showing the lines I wanted it to, but if you View Changes and search for "botania:monocle", you'll see the section I'm talking about.

NillerMedDild: I saw this change on the last PR, I think you're right that FTBQ adds it automagically :) all good.

NillerMedDild (approved):

NillerMedDild:
Thank you very much :D

PR #2167, opened 2021-05-14
Miecraft?

### LINKED ISSUE #2166: [0.4.11] Diamond Honeycomb has no Centrifuge recipe
**Modpack version**
4.11

**Describe the bug**
You can not centrifuge diamond honeycomb to get anything from it.

**Logs**
Input pastebin link to crash-report or `latest.log`, if applicable.

**Screenshots**

image
What should be there

image
What is there

TemperedFool:
Typing is hard - correcting a few spelling bumbles in Forest/Rocky/Diamond/Soupy bees. Fixes #2166.

### DIFF
--- a/config/resourcefulbees/bees/special/Soup.json
+++ b/config/resourcefulbees/bees/special/Soup.json
@@ -21,7 +21,7 @@
         "mutations": [
             {
                 "type": "ITEM",
-                "inputID": "minecraft:bowl",
+                "inputID": "minecraft:oak_log",
                 "outputs": [{ "outputID": "minecraft:mushroom_stew", "weight": 10 }]
             },
             {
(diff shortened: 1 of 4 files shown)


config/resourcefulbees/bees/special/Soup.json
```diff
@@ -21,7 +21,7 @@
         "mutations": [
             {
                 "type": "ITEM",
-                "inputID": "minecraft:bowl",
+                "inputID": "minecraft:oak_log",
```
NillerMedDild, line 24: Is this change intended? The bowl made a bit more sense imo.

TemperedFool: It totally does, and currently isn't supported unfortunately. That's why the recipe was failing to load - you can mutate TO an item, but not from one. Closest thing to a bowl I could come up with was a block of wood... I would change this if/when Epic implements item mutations.

NillerMedDild:
Thanks!

TemperedFool:
Least I can do is clean up my own mistakes when I'm capable of it...

PR #2172, opened 2021-05-14
Chunk expansion part 1

pm065:
Added Emendatus chunks for:
Cloggrum
Froststeel
Regalium
Utherium
Iesnium
Starmetal
Thallasium
Elementium

Also changes Undergarden, Alfheim and Atum (in preparation for 5.0) ore variants (e.g. Undergarden Gold Ore) to drop chunks instead of ore.
The textures are currently very basic recolours. I couldn't make decent textures for Clusters even by recolouring, so I left those out for now.
They _should_ be completely fine apart from that.
See https://gyazo.com/56ffa89ef82f4892db3aa4c2a19f08b8 for a clip showing the chunks dropping and https://gyazo.com/da9bdeb7d03ea7100f5cfb69af833496 showing that the chunks can be smelted and stonecut (stonecutted?).
By the way, the lang keys are implemented, it's just that I was too lazy to refresh the client (:


pm065:
finished probably
Someone might want to add elementium to jaopca though

pm065:
in future I should probably not use pm065-master because there's some other stuff I want to do

NillerMedDild:
Very good work, but frankly, I don't think this is needed, with the exception of dropping chunks from Atum's / Undergarden's vanilla ores. Chunks were added to alleviate the issue of having multiple types of the same ore in your inventory when mining, you'll never come across that issue in Atum, Undergarden og with Starmetal :/

pm065:
🤔 Honestly, I just did this for consistency with EE, since it was bothering me how some ores dropped chunks and others didn't. I'm pedantic like that, to the point where I almost didn't include BYG in my own modpack _because it didn't have chests for its wood and that felt weird to me_. (I did later remove it anyway because of worldgen lag but that's not the point.)

NillerMedDild:
That's understandable, but making EE items through KubeJS does not bring consistency imo. If anything we would need to ask Rid to add these chunks 😄

pm065:
True (: I'll change it to just be the loot tables. I'll do a separate PR.

NillerMedDild:
Thanks, and sorry for wasting your time 😕

pm065:
No, no, it's fine! It's not a loss in any way whatsoever since I was intending to use this in my own modpack (:

PR #2240, opened 2021-05-21
Update end city loot table

ochotonida:
The latest Artifacts update changed a few loot tables including the end city one, I've updated the custom end city loot table to reflect those changes.


NillerMedDild:
Thanks!

PR #2243, opened 2021-05-22
Astral Sorcery and Blood Magic Progression

MuteTiefling:
Aquamarine criminally underused, changed heart of the sea recipe to use that instead of lapis dust
image

Shifted some recipes using Heart of the Sea to use Conduit instead.

Blood Shards, still used for tier 4 blood orb and tier 4 altar
image

Blood Stone Brick used to expand the blood altar:
image

Final Blood Orb:
image

Arc Reverter recipes changed to reflect the current blood orb crafts
image

Update to Endoflame recipe
image

Update to Hydroangea recipe
image

Both of these contain non-farmable items. Crimson Aura Fungus are found in the nether and Neptunium ingots from fishing. The intent is to make them accessible but not something that can auto farmed and made en masse.

Astral Progression
Linking Tool moved to standard crafting
image

Eidolon Workbench
image

Luminous Crafting Table
image

Sooty Marble now made with a Create Mixing recipe
image

Mechanical Crafting for Looking Glass
image

Resonating Gem moved to Mana Infusion. Costly. Moonstone catalyst required
image

Lightwell. Hello Create:
image

Tier 2 Altar:
image

Glass Lens:
Hard to tell, but it uses mana glass panes
image

Spectral Relay:
image

Attunement Altar
image

Observatory:
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/astralsorcery/block_transmutation.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/astralsorcery/block_transmutation.js
@@ -2,8 +2,7 @@ events.listen('recipes', (event) => {
     var data = {
         recipes: [
             { inputTag: '#forge:ores/iron', output: 'astralsorcery:starmetal_ore', starlight: 100 },
-            { inputTag: '#forge:ores/diamond', output: 'emendatusenigmatica:emerald_ore', starlight: 1000 },
-            { inputTag: '#forge:workbench', output: 'astralsorcery:altar_discovery', starlight: 60 }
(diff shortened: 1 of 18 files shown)


MuteTiefling:
Astral Wands:
While these are somewhat limited without access to attunement perks, they make a nice magical alternative to building gadgets and such. They feel like they should be available earlier than the current AS progression would normally allow.

Resonating Wand
image

Traversal Wand
image

Conversion Wand
image

Impulsion Wand
image

Illumination Wand
image

Formation Wand
image

MuteTiefling:
Observatory has a small problem in that it is actually available before it should be. Tried slipping a Lucerna Mantle in there, but due to the random NBT Astral adds, that didn't work out. 

Will likely make a special lens item to go into the recipe instead. The lens would be made on the appropriate altar and then used in this.

Issue #2249, opened 2021-05-23
valid apiary block suggestions


yuri0r:
**Description**
right now the list of valid apiary blocks is rather short so id like to propose the following additions:

**vanilla**
- blackstone bricks variants (polished blackstone brick stairs are already valid its weird that the bricks themselves are not)
- basalt (has woodlog-esk texture and makes for a nice addition

**blood magic has plenty of decorative blocks**
- all "path" 
- demonic bricks (and variants)
- runes (basic and all upgrades makes for very evil looking apiaries)
- ritual blocks (and variants)

jei search term to find the blocks more easily
`@bloodmagic safe | @bloodmagic path | @bloodmagic Demon bricks| @minecraft polished blackstone bricks | @minecraft basalt`

NillerMedDild:
Thank you for the suggestion, I'll have a look :)

yuri0r:
may i add to my suggestions? IE treated wood planks and the botania grass blocks (jei search `@botania grass -aiot`) those grasblocks also disable passive mob spawns wich is usefull, gets anoying to remove sheeps from apiary. :)

<details>
  <summary>a few screenshots using some of the suggested blocks so you dont have to play around with all that many of them</summary>

![2021-05-23_23 56 01](https://user-images.githubusercontent.com/9966894/119277918-96cf2580-bc22-11eb-9b64-ade18ec457be.png)
![2021-05-23_23 56 07](https://user-images.githubusercontent.com/9966894/119277920-99317f80-bc22-11eb-9c55-b16034e31869.png)
![2021-05-24_00 03 08](https://user-images.githubusercontent.com/9966894/119278105-7c497c00-bc23-11eb-95ba-e545b6e9e20b.png)

</details>

TemperedFool:
Those should be covered once #2257 is closed. Any others I can add before that (and this) are closed @yuri0r?

yuri0r:
thats all i am aware of right now. i guess if engeniers decor is in the pack than just about anything that one adds is beatifull

PR #2257, opened 2021-05-24
More Valid Apiary Blocks

TemperedFool:
Making Apiaries pretty one block at a time. Addresses #2249.

Adds all stonecuttables (covers Basalt and Blackstone variants).
Add various Blood Magic blocks (Runes, Ritual Stones, Paths, Bricks).
Add Botania Grasses.
Add Treated Wood.


TemperedFool:
Also blocks from Engineer's Decor.

natebc:
@TemperedFool might want to add botania planks, currently `$valid_apiary @botania livingwood` results in.
image
hard to see but there's no planks, only some bits.

PR #2259, opened 2021-05-24
Proof of Concept Magical 5x Ore Processing

MuteTiefling:
Proof of Concept Magical 5x Ore Processing.
All art assets are temporary. To be replaced later.

Drop the ore into a Mana Pool (requires Creational Catalyst under it)
image

Zap result with lightning and a Rune of Asgard catalyst
image
Return chance doesn't appear to work with lightning. On the other hand, it does let you craft many sets with only 1 run as input. So it will be up to the player to decide how efficient they want to be. Drop 1 rune and 64 inputs? more? Considering the recipe for the Ingot of the sky used in this, it'll be unwise to do small batches.

This step produce roughly 3x output with a mix of the input item and a secondary such as Nickel from Iron.

Crumble in Natural Infusion Altar
image

Wash in Starlight:
image
Chance to use up both starlight and the winter rune.

This is the final multiplication step, producing roughly 15x of the input material. 9 of the output goes into making an ingot.

Finally, combine in a Runic Altar. This rune, as with all runes used in Altar crafting, is always returned. Calx of the End is something that will likely be on auto craft already given it's many uses.
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/unification/unify_materials.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/unification/unify_materials.js
@@ -12,10 +12,27 @@ events.listen('recipes', (event) => {
         var rod = getPreferredItemInTag(Ingredient.of('#forge:rods/' + material)).id;
         var wire = getPreferredItemInTag(Ingredient.of('#forge:wires/' + material)).id;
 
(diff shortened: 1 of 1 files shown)


NillerMedDild (approved):
Thanks! Will have a look :D

PR #2267, opened 2021-05-25
Remove redundant Valhelsia Structures datapack recipes

ochotonida:
This PR removes a few empty recipe jsons, which I'm assuming were added to remove the recipes with the same id. Since these recipes are already in the recipe removal script, removing these files will make the log a couple errors lighter as they're not valid recipe jsons anyways.


Darkere:
The problem is that people that update to a newer version of the pack do not get their instance data wiped. 
The new version just gets copied on top of the old one.
So if we don't ship a file there, the old version will persist. 

That said not sure how old those are...

ochotonida:
Right, forgot that that applies to data packs too. 
In this case it doesn't really matter though, they've been empty from the beginning (except the acacia one for some reason). Even if someone upgrades from an old version where the acacia recipe isn't empty, I'm pretty sure the recipe removal script would just remove it.

MuteTiefling:
The removal scripts run after these anyway, so removing them should be perfectly safe. Worst case, an old world still has old clutter data packs doing nothing.

NillerMedDild:
Oh those were still there, woops 😄

PR #2269, opened 2021-05-25
More datapack cleanup

MuteTiefling:
Move aquaculture datapack recipes into scripts.
Remove botany pots datapacks. everything has been in scripts for a while now
Move Create datapacks to scripts
Move IE's datapack recipes to scripts
IP's reservoir changes moved to scripts
Industrial Foregoing recipe replacements moved into scripts


NillerMedDild:
And these all work? O_o
IE and IP were not migrated as they didn't work in KJS at the time.

MuteTiefling:
yep, all working :D If I recall correctly the issue before was ID replacements not being a thing and we just never cleaned stuff up after it was

NillerMedDild:
Ahh I see! Awesome work :D

MuteTiefling:
trying to get through all of these old datapacks that are lying around. was a good day for it

PR #2278, opened 2021-05-26
TerraForged changes attempt #1

theboo:
Removed some horizontal spread and added a bit of verticality to some of the mountainous terrain features, also lowered sea level a bit which seems to allow for surface caves and ravines to appear!!!

if anyone has time to test this that would be great! lmk if u have any feedback

### DIFF
--- a/config/terraforged/presets/Enigmatica.json
+++ b/config/terraforged/presets/Enigmatica.json
@@ -55,10 +55,58 @@
         "salt": 1549795968,
         "disabled": false
       },
+      "atmospheric:arid_shrine": {
+        "spacing": 32,
+        "separation": 8,
+        "salt": 304972544,
+        "disabled": false
+      },
+      "bettermineshafts:mineshaft": {
+        "spacing": 1,
+        "separation": 0,
+        "salt": 593751808,
+        "disabled": false
+      },
+      "betterstrongholds:stronghold": {
+        "spacing": 85,
+        "separation": 50,
+        "salt": 596441280,
+        "disabled": false
+      },
       "dungeoncrawl:dungeon": {
         "spacing": 40,
         "separation": 10,
         "salt": 10387313,
+        "disabled": false
+      },
+      "dungeons_plus:bigger_dungeon": {
+        "spacing": 12,
+        "separation": 5,
+        "salt": 973181,
+        "disabled": false
+      },
+      "dungeons_plus:leviathan": {
+        "spacing": 36,
+        "separation": 8,
+        "salt": 719643,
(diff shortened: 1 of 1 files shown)


theboo:
looks like we have control over the slime islands through TF, def will need to lower their frequency a bit xD

NillerMedDild:
I spawned next to a cave opening, gg :P

NillerMedDild:
Several even! 

Could we remove BYG's Autumnal Valley? I think the "pumpkin trees" are an abomination :P

I found ~15 slime islands in 2 minutes flying in creative, you're right, slightly too many xD

theboo:
Woohoo! Yeah, I never touched sea level before but when I was going through settings I saw it's defaulted to 63 which is a block higher than vanilla, I lowered it to 56 since many flat areas are still well above sea level seems to have allowed the caves to pop through

theboo:
Yeh easy mode, imma go through the biome weights again later this week... There's a pvj biome that looks horrendous (overgrown spires) in TF that I think needs yeeted as well..

NillerMedDild:
PVJ has no biomes though :P nor do we have a biome called overgrown spires

pm065:
> PVJ has no biomes though :P nor do we have a biome called overgrown spires

It does now (it's in the process of being returned to the state it was in in 1.15)

NillerMedDild:
No? Unless they were added without a trace in commits and changelogs.

theboo:
There's 2 pvj biomes in the pool, idr the other ones name tho but do /terra locate biome project and they will show up

theboo:
One of em looks nice the other is just surface ore everywhere lol TF might flatten the spires depending on how they added them

pm065:
Verdant Sands is the other one I think

theboo:
Ahh yeah it's lush style desert... We might have that covered with atmospheric tho so it might need yeeted too

pm065:
See changelog here: https://www.curseforge.com/minecraft/mc-mods/project-vibrant-journeys/files/3275473
Isn't Atmospheric gonna get oofed in 5.0?

theboo:
Neither are in the biome weights conf yet, it may be a WIP thing similar to byg canyons which are disabled by default in byg but TF picks them up anyway so have to manually change weight to 0

NillerMedDild:
Ahh I see, my bad - It is a bit meh in TF, agreed.
billede

> See changelog here: https://www.curseforge.com/minecraft/mc-mods/project-vibrant-journeys/files/3275473

Oh derp, I thought they were odd names for groundcover :P

PR #2342, opened 2021-06-02
Add Hive UpgradeRecipes

### LINKED ISSUE #2303: Apiary Recipe Deletes Bees
**Modpack version**
0.4.10 possibly newer version have it too as the bug is still in the GitHub files.

**Describe the bug**
Resourceful Bee Apiary deletes bees when making it via the recipe while the base mods hive recipes don't this is because the apiary recipes were edited and don't use the Resourceful Bees hive upgrade recipe type. 
As seen here you just do a normal shaped recipe https://github.com/NillerMedDild/Enigmatica6/blob/master/kubejs/server_scripts/enigmatica/kubejs/base/recipes/shaped.js#L886
While Resourceful Bees does https://github.com/Resourceful-Bees/ResourcefulBees/blob/1.16.5/src/main/resources/data/resourcefulbees/recipes/t1_apiary.json which makes it so that it will return a bee box with all the bees in the hives.

**Logs**
N/A

**Screenshots**
N/A

TemperedFool:
Closes #2303. Added Hive_Upgrade_Recipes for all of the RB hive/apiary recipes we're changing. I also took a moment to redistribute things between base/normal/expert - this will leave room to adjust the T2-4 apiary recipes for Expert.

There are some caveats regarding upgrades and returning bees (detailed in commit b211a11), so I've also included JEI Descriptions/tooltips as a warning.


NillerMedDild (approved):

NillerMedDild:
Looks good! :D

PR #2370, opened 2021-06-07
Botania utility[Expert] Redstring and Corporea rework

MuteTiefling:
New Alternate Red String recipe. This makes redstring contraptions available a bit earlier than normal.

image

Livingrock still required for the blocks, so getting up to Nature's Aura to make will be the limiting factor.

Corporea is a very useful storage system early on (before things like RS beat it out) so I feel it's better to bring it up a bit considering how deep most Botania is. The following recipes bring it up to Nature's Aura level of magic as well as requiring some exploration of Atum

Revamp to Corporea recipes:
Base spark:
image

Corporea Spark:
image

Master Corporea Spark requires a jaunt to Atum to get a god shard:
image

Corporea Block makes use of Infused Rock, made through Nature's Aura.
image

Corporea Index requires a bit of ectoplasm from Atum Wraiths
image

Corporea Crystal Cube:
image


NillerMedDild:
Great idea to make Corporea networks available early, and the recipes look good!

Issue #2380, opened 2021-06-08
Allow colored honey blocks in tier 4 beehive recipe


ItsAMemeMario:
Tier 4 beehive's recipe calls for only vanilla honey blocks which is a bit of an oversight by the mod author in my opinion. Allowing RB honey block variants to be used in this recipe will remove the need to farm vanilla bees and give us more reason to craft honey into blocks.

TemperedFool:
Oh don't blame Epic and the RB team, this is an oversight on my part. Easy fix, will PR it this morning.

PR #2381, opened 2021-06-08
T4 Hive/T1 Apiary use all Honey Blocks

TemperedFool:
What it says on the tin. Turns out I can take 50% responsibility here; the default T4 Hive recipe is coded to Minecraft Honey Blocks, and I did the same with the T1 Apiary recipe. Fixed both to use all honey blocks.

I can't for the life of me figure out why this won't auto-merge; I've just modified a line and added a few. Thanks GitHub...


TemperedFool:
Thanks to @MuteTiefling for pointing out I can just use the web editor and resolve it myself. I'm so glad GH suggested that rather than checking things out using the command line...

TemperedFool:
Also, fixes #2380.

PR #2433, opened 2021-06-17
Restore #forge:netherbricks tag to nether bricks

### LINKED ISSUE #2432: [0.5.0] minecraft:nether_bricks is missing #forge:netherbricks tag
**Modpack version**
0.5.0 take 2

**Describe the bug**
Missing forge:netherbrick tag breaks cobblegen and recipies using nether bricks

**Logs**
N/A

**Screenshots**

MrBazlow:
This may not be the right way to fix this as somewhere else it has lost its native tag but this looks good, is easy to find again and it works too.
Fixes #2432


NillerMedDild (approved):

NillerMedDild:
Fine solution, thank you :)

Issue #2441, opened 2021-06-17
No recipe in JEI for volcanic accumulator from ARS


billyty08:
**Modpack version**0.5 alpha

**Describe the bug**
A clear and concise description of what the bug is. I am not sure if it is a big issue but there is no recipe in Jei for the volcanic accumulator. When you use control to look at it in the book the book has the recipe and it actually makes it.  It is just not very intuitive when you are so used to seeing it in Jei and all others recipes I checked from Ars was in there. Clicking on it in the questline didn't bring up recipe is how I noticed it
**Logs**
Input pastebin link to crash-report or `latest.log`, if applicable.

**Screenshots**

MuteTiefling:
Can you do a search for `$gold_brass` in JEI and let me know what you see?

MuteTiefling:
nevermind, figured out the cause.

anivlys:
Is this the same issue that also effects Botania's Spark? No recipe in JEI, but has the working recipe.

MuteTiefling:
Yes, it was the same issue.

PR #2444, opened 2021-06-17
Unyeet yeeted expert recipes in normal

MuteTiefling:
Oops #2441


NillerMedDild:
We may need to keep an eye out for issues here, hiding things based on mode is flaky unfortunately.

MuteTiefling:
Oh.. Hrm

PR #2447, opened 2021-06-18
adds generic bee icon to in-world spawned bees in journeymap

theboo:
copied most of these over from the ftbchunks resource pack we have, and added the rest of the in-world spawned bees. Tested in-world


NillerMedDild:
Nice work!

PR #2466, opened 2021-06-19
[Expert] Rebalance Cutting Fluid crafting

MuteTiefling:
Add a step to the creation to increase the yield the mob drops used in the recipe. Now uses rattlesnake rattles and komodo dragon spit.

Cutting Essence:
image

Cutting Fluid:
image


NillerMedDild:
> Now uses rattlesnake rattles and komodo dragon spit.

How hard it that to come by? I'm thinking the yield could be more than 8, if the entities aren't very common.

MuteTiefling:
They can both be tamed and bred.

Tame komodo dragons drop about 1 spit per minecraft day.

Snakes you can just breed endlessly with chicken and kill them for rattles. So no different than farming cows or whatever. They even work with the feeding trough from quark for a simple farm.

PR #2469, opened 2021-06-20
[Performance] Set alwaysSetupTerrainOffThread to true

yagoplx:
Back in 1.12 foamfix would set this to true for you, but the pack doesn't have this on.

This actually helps my (low end) computer to cope with terrain loading, and doesn't seem to make anything behave oddly; just like the old times.


yagoplx:
I screwed up. Let me redo this...

Issue #1233, opened 2021-02-06
Make Multithreaded Terrain Rednering Default?


TemperedFool:
**Description**
There are frequent questions on #performance about poor FPS in E6 with otherwise decent systems; these are frequently addressed by changing the following in forge-client.toml:

alwaysSetupTerrainOffThread=true

I'm not clear that there is a downside to making this the default; if I understand correctly the potential downside is with low core count systems, which for lots of reasons probably won't run E6 particularly well anyway.

Darkere:
There are issues with chunks not rendering or even only partially rendering with this option. Which is why it is not default.

TemperedFool:
Interesting. If it's unstable I suppose making it the default doesn't make sense. I suppose we can just keep pointing those with FPS issues at it when they ask ^_^

PR #2470, opened 2021-06-20
[Performance] Set alwaysSetupTerrainOffThread to true

yagoplx:
Repeat PR because I messed up the first one.

Back in 1.12 foamfix would set this to true for you, but the pack doesn't have this on.

This actually helps my (low end, dual core) computer to cope with terrain loading, and doesn't seem to make anything behave oddly; just like the old times. It's always worked.

### DIFF
--- a/config/forge-client.toml
+++ b/config/forge-client.toml
@@ -8,7 +8,7 @@
 	#Enable Forge to queue all chunk updates to the Chunk Update thread.
 	#May increase FPS significantly, but may also cause weird rendering lag.
 	#Not recommended for computers without a significant number of cores available.
-	alwaysSetupTerrainOffThread = false
+	alwaysSetupTerrainOffThread = true
 	#EXPERIMENTAL: Enable the Forge block rendering pipeline - fixes the lighting of custom models.
 	experimentalForgeLightPipelineEnabled = true
 	#Enable the Forge block rendering pipeline - fixes the lighting of custom models.


DarkAngel22579:
https://github.com/NillerMedDild/Enigmatica6/issues/1233
>There are issues with chunks not rendering or even only partially rendering with this option. Also some general rendering funkyness that can occur. Which is why it is not default.

yagoplx:
https://github.com/NillerMedDild/Enigmatica2Expert/blob/master/config/forge.cfg

Then why is it default on E2 but not here?

I have literally never seen anyone complain about this option, ever, since it seems to work on most systems. The benefits outweigh the risks, even for me who doesn't have a "significant amount of cores available".

TemperedFool:
> Then why is it default on E2 but not here?

Because E2 is not E6. I'm not s java programmer, so just speculating, but Forge got a major rewrite for 1.13, and the "performance tweak mod" landscape is different. I've always seen this turned on in 1.12 in context of a performance mod that was already doing some yucky under the hood stuff to try to fix issues/performance, and I remember a number of people needing to turn it off.

> I have literally never seen anyone complain about this option, ever, since it seems to work on most systems. The benefits outweigh the risks, even for me who doesn't have a "significant amount of cores available".

The general consensus from people smarter than I (such as Darkere in the above referenced issue) is that there is non-trivial risk of this option causing issues (breaking rendering if I understand correctly).

If it isn't a default, we can suggest it as a performance remedy while warning it may cause rendering glitches. If it's the default, we'll just get a bunch of "E6 rendering is broken" reports, which seems worse from a support POV.

billyty08:
well now you 

> https://github.com/NillerMedDild/Enigmatica2Expert/blob/master/config/forge.cfg
> 
> Then why is it default on E2 but not here?
> 
> I have literally never seen anyone complain about this option, ever, since it seems to work on most systems. The benefits outweigh the risks, even for me who doesn't have a "significant amount of cores available".
> 
> >  GitHub**[NillerMedDild/Enigmatica2Expert](https://github.com/NillerMedDild/Enigmatica2Expert)**The official repository and issue-tracker for the modpack Enigmatica 2: Expert for Minecraft 1.12.2 - NillerMedDild/Enigmatica2Expert

Well now you have heard a complaint. It works good when I am in my base but the second it tries to render new chunks it completely lags out to a crawl where I was before changing the config getting close to normal rendering

PR #2511, opened 2021-06-23
Remove Enriched Tin to fix resource loop

### LINKED ISSUE #2510: Mekanism enriched tin results in resource loop
**Modpack version**
0.5.0 Dev

**Describe the bug**
There is currently a tin resource loop that can generate infinite tin with FE

1 Tin Ingot -> 1 Enriched Tin (Worth 8 Tin in Metallurgic Infuser) -> Combine with copper in Metallurgic Infuser to create 32 Bronze -> Crush/Pulverize to Bronze Dust -> Centrifugal Separator = 8 Tin and All copper back

Since Enriched Tin isn't used for anything else but making bronze in the infuser and the fact that it results in 8x the value of tin when alloying it should just be removed and force users to use tin dust in the metallurgic infuser instead.

MVethana:
Fixes #2510

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
@@ -320,5 +320,5 @@ onEvent('recipes', (event) => {
     event.remove({ type: 'mekanism:enriching', output: '#forge:dyes' });
     event.remove({ type: 'thermal:centrifuge', output: '#forge:dyes' });
     event.remove({ type: 'immersiveengineering:crusher', output: '#forge:dyes' });
-    event.remove({ type: 'mekanism:enriching', input: '#forge:dusts/tin'});
+    event.remove({ id: 'mekanism:enriching/enriched/tin'});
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Thank you for your contribution! Could you perhaps remove the recipe through it's id instead? I know it's unlikely that a different recipe will be affected by your removal, but I like future-proof where possible :D

MVethana:
Removed with recipe id instead, lmk if anything else needs to be changed for it

NillerMedDild (approved):

NillerMedDild:
Looks good, ty - Would you mind also removing the 2 bronze Metallurgic Infuser recipes? :)

MVethana:
Just pushed it. I'm assuming making bronze via the infuser was never intentional?

NillerMedDild (approved):

NillerMedDild:
Thanks! I knew it was possible to make Bronze with the Infuser, I just hadn't made the connection between Centrifugal Separator + cheap Bronze creating a loop though 😄

Issue #2507, opened 2021-06-23
Merge Wither Skeleton bones


g0urra:
**Description**
Replace source and functionality of Necrotic Bone (Tinker's Construct) with Withered Bone (Architect's Palette), or vice versa.

**Why would you like this added/changed?**
Both items drop from the same mob (Wither Skeleton), however the first one currently only drops when using the Severing modifier (basic modifier on the Cleaver, and can be applied in Tinker's Anvil and Tinker's table). Necrotic Bone is also only used to convert into regular Bone, while Withered Bone has many more functionalities.

The name of the item may have to stay the same, due to the Necrotic Bone parts that currently exists from Tinker's Construct. Withered Bone Block may have to be renamed.

MuteTiefling:
In the interest of not changing Tinker's progression, a tag should probably just be made to allow both to be used for architect's palette without changing how they're used for Tinkers. Can take a look later to see how that'll work out.

PR #2514, opened 2021-06-24
Wither Bone Unification

MuteTiefling:
Completes #2507

Tinker's wither bones are now useful for making Architect's Palette blocks and can be converted to bonemeal.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
@@ -146,6 +146,8 @@ onEvent('recipes', (event) => {
         'simplefarming:apple_pie',
         'simplefarming:blueberry_pie',
 
+        'tconstruct:tools/modifiers/wither_bone_conversion',
+
         'thermal:compat/tconstruct/chiller_tconstruct_molten_debris_ingot',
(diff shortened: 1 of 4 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/recipes/remove.js
```diff
@@ -146,6 +146,8 @@ onEvent('recipes', (event) => {
         'simplefarming:apple_pie',
         'simplefarming:blueberry_pie',
 
+        'tconstruct:tools/modifiers/wither_bone_conversion',
```
NillerMedDild, line 149: What does removing this effectively do? 😄

MuteTiefling: It's a conflict. Shapeless one wither bone to one normal bone. But with the tags it can be converted to bone meal already with that recipe.

If we really want a conversion to normal bone, we can add a shaped craft, I guess. Just didn't seem that important.

NillerMedDild: Oh alright, because of the name I thought it was the "recipe" for the Necrotic Bone modifier.

Issue #2516, opened 2021-06-24
Mekanism Crusher Recipes Missing


MVethana:
**Modpack version**
0.5.0 Dev

**Describe the bug**
Mekanism Crusher/Factory is missing ingot-> dust recipes. Ex. silver dust, brass dust, aluminum dust, etc.

Also existing ingot/gem -> dust recipes use a mix match of naming conventions for the recipe id Ex.
mekanism:processing/iron/dust/from_ingot
mekanism:processing/steel/ingot_to_dust
mekanism:processing/diamond/to_dust

PR #2517, opened 2021-06-24
Implement mekanism_ingot_gem_crushing(),

MVethana:
- Implement mekanism_ingot_gem_crushing() to add missing Crusher recipes
- Standardizes mekanism crusher ingot/gem -> dust recipe id's as /to_dust
- Removes redundant ender dust recipe in [crushing.js](https://github.com/NillerMedDild/Enigmatica6/blob/develop/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/mekanism/crushing.js) as a result of these changes

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
@@ -48,6 +48,7 @@ onEvent('recipes', (event) => {
         immersiveengineering_hammer_crushing(event, material, ore, dust);
         immersiveengineering_gem_crushing(event, material, dust, gem);
 
+        mekanism_ingot_gem_crushing(event, material, ingot, dust, gem);
         mekanism_metal_ore_processing(
             event,
             material,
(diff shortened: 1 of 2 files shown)


MVethana:
Fixes #2516

kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
```diff
@@ -627,6 +628,34 @@ onEvent('recipes', (event) => {
             .id(`immersiveengineering:crusher/ore_${material}`);
     }
 
+    function mekanism_ingot_gem_crushing(event, material, ingot, dust, gem) {
+        if (dust == air) {
```
NillerMedDild, line 632: Please check ingot too :)

kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
```diff
@@ -627,6 +628,34 @@ onEvent('recipes', (event) => {
...
+            return;
+        }
+      
+        var input,
+            output = Item.of(dust, 1);
```
NillerMedDild, line 637: `Item.of` is redundant here since there's only 1 dust.

NillerMedDild (changes requested):
Looks good for the most part, thanks for your contribution!
I'd like my comments addressed before I merge though :)

kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
MVethana: It seems redundant to check for ingot and I would assume gem here aswell? The if & elif check below checks both of them to ensure they are not air, to set the proper input type. If it is air, it falls to the else statement anyways and returns nothing

NillerMedDild (approved):

kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_materials.js
NillerMedDild: Oh, right ^^' you're completely right about that.

PR #2518, opened 2021-06-24
[Expert] Outlining progression via temp(ish) quests and more changes to progression

MuteTiefling:
Change flower progression for Botania:

First flower likely to be gourmaryllis, requiring materials from Undergarden. Could technically make a hydroangea if the player is truly hard pressed to find masticator scales. Starting off with a good generator is going to be relevant considering the increased cost of runes, however. This cost trend may apply to other mana crafts too...
image

Endoflames now require runes to make:
image

Small change to pure daisy, now uses conduit as most starting magic crafting methods do.
image

Same for the Crucible:
image

Divination Sigil:
image

Iesnium Pickaxe changed again, bearing in mind that the sentient pick also mines iesnium and is available earlier.
image

Rough outline of progression created in hidden quest chapter.  Does not cover the entire progression yet, but it gives a good basis to start from. Will work on expanding this as a reference for further progression work and testing. Ultimately it will be pruned down to just a high level 'guide' for the player.

### DIFF
--- a/config/ftbquests/quests/chapters/expert__tier_2_wip.snbt
+++ b/config/ftbquests/quests/chapters/expert__tier_2_wip.snbt
@@ -13,5 +13,548 @@
 	}
 	always_invisible: true
 	default_quest_shape: ""
-	quests: [ ]
+	quests: [
+		{
+			x: -0.5d
+			y: -5.0d
+			id: "72D589CE18202510"
+			tasks: [{
+				id: "46003D18E8B6E35F"
+				type: "item"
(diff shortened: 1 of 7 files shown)


MuteTiefling:
Upon further inspection, the Masticator doesn't spawn naturally right now, so I'm thinking a Nature's Aura Altar of Birthing summon will be in order.

MuteTiefling:
Yeah, Masticator spawner:
image

NillerMedDild (approved):

Issue #2530, opened 2021-06-25
[Worldgen] Investigate BYG + terraforged weird interaction


yagoplx:
**Modpack version**
0.5.0

https://github.com/TerraForged/tracker/issues/103

Seen this one in person while playing with the worldgen. Especially the red rock biomes, they all look quite jarring. You might want to look into the issue (eg. reporting to byg, removing the biomes from the weights file or do nothing) whatever you think is best.

NillerMedDild:
Thanks for the report - We're aware, and okay with it :)

Issue #2522, opened 2021-06-24
[Mod Suggestion] Better Biome Blend


yagoplx:
Can't find this on the issues or modlist so I assume it's never been brought up.

**CurseForge Link**
https://www.curseforge.com/minecraft/mc-mods/better-biome-blend
https://github.com/FionaTheMortal/Better-Biome-Blend

**Mod description**
Performance mod. Implements faster biome color blending, increasing performance on the client only, especially while moving around and loading chunks. Sodium has a similar optimization.

**Why would you like the mod added?**
speed

**Screenshots if applicable.**
One from curseforge, max blend setting is increased:
![better-biome-blend-maximum](https://user-images.githubusercontent.com/57966308/123315422-1726c600-d502-11eb-99ad-2301cd1dfca0.jpg)

yagoplx:
Mod is missing from 0.5.1 ;/

NillerMedDild:
It's not though :P

yagoplx:
wot

PR #2532, opened 2021-06-25
[Worldgen] Tweak biome warp for biome transition effect

yagoplx:
**Part of the worldgen changes patch series. Individual features are separate PRs to allow each one to be discussed separately.**

**What this do?**
Changes biome warp so that biomes bleed into each other's borders. This makes biome transitions a steep gradient instead of a harder edge. The visual effect is synergistic with https://github.com/NillerMedDild/Enigmatica6/issues/2522. These transition zones are not big.

Transition zone with this setting. Notice the minimap. The bleed-in effect can be increased by increasing the strength of the warp but keeping the scale low.

**Questions you may ask**
"Wouldn't this cause ugly chunk borders?"
No, changing the preset file does not seem to affect existing worlds at all. Old worlds will not get these changes, but old worlds will still suffer biome border weirdness due to biome mod changes in 0.5.0.

"Wouldn't this cause terrain weirdness?"
Biomes don't affect the heightmap in Terraforged so no. Biomes that misbehave like https://github.com/NillerMedDild/Enigmatica6/issues/2530 may not benefit from this setting, but you probably won't be living near them anyway.

### DIFF
--- a/config/terraforged/presets/Enigmatica.json
+++ b/config/terraforged/presets/Enigmatica.json
@@ -577,8 +577,8 @@
     "biomeShape": {
       "biomeSize": 326,
       "macroNoiseSize": 4,
-      "biomeWarpScale": 150,
-      "biomeWarpStrength": 80
+      "biomeWarpScale": 27,
+      "biomeWarpStrength": 180
     },
     "biomeEdgeShape": {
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Also a nice change, gj 👍

PR #2558, opened 2021-06-28
Removed Quark buttons from crafting station's GUI

DarkAngel22579:
Prevents this from happening when you place a chest next to the crafting station:


NillerMedDild:
Thank you kindly :D

PR #2572, opened 2021-06-29
Fix broken Kenaf to String recipe

MuteTiefling:
Fix broken Kenaf to String recipe


NillerMedDild:
How was it broken? :o

MuteTiefling:
It was gone :-D

The second recipe there was another recipe adding the same ID

NillerMedDild:
Oh, I see ^^'

PR #2583, opened 2021-06-30
Tooltip explaining how to obtain Ham from Farmer's Delight

Thiana:
Add a tooltip explaining how to obtain Ham from Farmer's Delight.


NillerMedDild:
Thanks!

PR #2588, opened 2021-06-30
Fix Curing Rubber take sulfur gems when most recipes produce sulfur dust

Larandar:
Most recipes for pulverizers create sulfur dust as a secondary output (except blaze rods) but curing requires sulfur gem.

Therefore this PR change curing to use sulfur dust (both in induction smelter and blood altar).

This PR also makes it so the only recipes that create sulfur gems are the ore processings or Sulfur Gems (secondary outputs being dust) (and 2 recipes from the induction smelter using slag, but induction smelter creating dust doesn't fit for me?)

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipes/replace_input.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipes/replace_input.js
@@ -14,7 +14,7 @@ onEvent('recipes', (event) => {
     event.replaceInput({}, 'mythicbotany:elementium_ore', '#forge:ores/elementium');
     event.replaceInput({}, 'thermal:rubber', 'industrialforegoing:dryrubber');
     event.replaceInput({}, 'thermal:cinnabar', '#forge:gems/cinnabar');
-    event.replaceInput({}, 'thermal:sulfur', '#forge:gems/sulfur');
+    event.replaceInput({}, 'thermal:sulfur', '#forge:dust/sulfur');
     event.replaceInput({}, 'thermal:apatite', '#forge:gems/apatite');
     event.replaceInput({}, 'thermal:niter', '#forge:gems/niter');
     event.replaceInput({}, 'thermal:bitumen', '#forge:gems/bitumen', true);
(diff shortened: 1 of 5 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/recipes/replace_input.js
```diff
@@ -14,7 +14,7 @@ onEvent('recipes', (event) => {
     event.replaceInput({}, 'mythicbotany:elementium_ore', '#forge:ores/elementium');
     event.replaceInput({}, 'thermal:rubber', 'industrialforegoing:dryrubber');
     event.replaceInput({}, 'thermal:cinnabar', '#forge:gems/cinnabar');
-    event.replaceInput({}, 'thermal:sulfur', '#forge:gems/sulfur');
+    event.replaceInput({}, 'thermal:sulfur', '#forge:dust/sulfur');
```
MuteTiefling, line 17: This might have some unintended consequences, like making sulfur blocks from dust instead of gems. And the sulfur + earth charge recipe becoming dust to dust.

MuteTiefling:
While we're on the topic, it may be a good idea to adjust this recipe too: 

image

Make it all dusts, perhaps.

kubejs/server_scripts/enigmatica/kubejs/base/recipes/replace_input.js
Larandar: Mmmh, will check then

Larandar:
What about removing this one then? There is one from nitrate dust + sulfur dust already?

MuteTiefling:
Yeah, actually, I think both of these can go and just keep the first one that yields 6

image

Larandar:
It's done:

alchemy_table
curing_induction
from_storage
shapeless
to_storage

I tried to also unify eidolon sulfur dust but couldn't find the way... I removed the corresponding commit
eidolon

MuteTiefling:
Yeah don't worry about eidolon. It doesn't have datapack support yet.

kubejs/server_scripts/enigmatica/kubejs/base/recipes/replace_input.js
Larandar: Resolved in 41beae2

NillerMedDild:
Very nice work, thanks for your contribution :)

Larandar:
Thank you, with the amount of entertainment contributing small config changes like that is the minimum I can do

PR #2598, opened 2021-07-01
Add Wiki button to Pack Menu

TemperedFool:
Shuffles the Quit and bottom row of buttons down to make room for a Wiki button. I've just copied the Github button png for it, so someone with artistic skills (which always turns out to be @Ridanisaurus it seems...) might should probably replace packmenu/resources/assets/enigmatica/textures/wiki.png.


TemperedFool:
It ends up looking like this:

image

TemperedFool:
Improved png from Rid incoming! I'll merge it after.

TemperedFool:
Final results - @Ridanisaurus is on point with the art as usual!

image

NillerMedDild:
Good work :D

PR #2610, opened 2021-07-03
Starry Bee color changes

g0urra:
Gives updated color and textures to Starry Bee.
image
image


NillerMedDild:
Nice change, ty :)

PR #2619, opened 2021-07-04
Refactor shaped.js - Expert

TemperedFool:
Move all shaped.js recipes to mod-specific files in their expert/recipetypes folders. The only recipe that seems to have an issue is Industrial Foregoing's Pity Machine Frame; I'm not clear what's going on with it, whether or not it worked previously, but I will investigate tomorrow.

Leaving a commented-out shaped.js in expert/recipes both to prevent pack upgrade issues & in case I somehow missed a recipe.


NillerMedDild (approved):

NillerMedDild:
The Pity Machine Frame did not work before it seems :P gj!

TemperedFool:
> The Pity Machine Frame did not work before it seems :P gj!

In that case I'll take a swing at fixing it this evening. I'll also work on Normal shaped.js - the base one looks like it's going to take a fair amount of cleanup, so that can go last...

NillerMedDild:
Yeah the normal one is.. a monster :P 

Already fixed the pity frame thing: https://github.com/NillerMedDild/Enigmatica6/commit/9b1ee757e838f242f84b06f0ac1fc636fe8e0a31

PR #2640, opened 2021-07-06
Changed Dissolution chamber power/t to default

Buuz135:
400FE/t was the default by mistake a long time ago


Darkere:
@Buuz135 Possibly related? The power values in JEI made 0 sense. For simple machine frame it said it would take 18k FE but took like 120k FE.

NillerMedDild (approved):

PR #2646, opened 2021-07-06
Cleanup curios inventory

ochotonida:
This PR removes the glasses & bracelet slots from the curios menu:
* The AR glasses from Advanced Peripherals now use the goggles slot instead
* The bracelets from Atum now use the hands slot (they're tagged as vanity but they don't even show up on the player model, so these definitely don't need their own slot)

### DIFF
--- a/defaultconfigs/curios-server.toml
+++ b/defaultconfigs/curios-server.toml
@@ -54,3 +54,13 @@
   identifier = "feet"
   override = false
   hasCosmetic = true
+
+[[curiosSettings]]
+  identifier = "bracelet"
+  override = true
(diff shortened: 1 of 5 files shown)


NillerMedDild:
Have you verified that the AR glasses and bracelets still function in their new slots? :)

ochotonida:
The bracelets don't do anything so it shouldn't be a problem for those. I'm not familiar enough with Advanced Peripherals to test the goggles (just equipping them doesn't do anything), but I dug through their code a bit and didn't find anything that should prevent it from working in other curio slots.

PR #2663, opened 2021-07-09
Eidolon: clarify the need to find a witch and not a swamp in the first quest

Larandar:
I spent a few hours looking for a swamp and throwing a book in water (granted I was tired) before reading the book and facepalmed so hard I finally got some sleep 🤦 

Please people like me need to be helped 👍

### DIFF
--- a/config/ftbquests/quests/chapters/eidolon_wip.snbt
+++ b/config/ftbquests/quests/chapters/eidolon_wip.snbt
@@ -96,7 +96,7 @@
 				""
 				"While most signs are learned from the gods themselves while progressing in theurgy, a few basic ones might be known to magical users in the normal world."
 				""
-				"Try paying a visit to your local swamp and throwing your Ars Ecclesia at one of them, you might learn something new..."
+				"Try paying a visit to your local witch and toss her your Ars Ecclesia, you might learn something new..."
 			]
 			dependencies: ["0000000000000E80"]
 			id: "0000000000000E85"


NillerMedDild:
To be honest I think it's clear enough - Find a magic user in a swamp, sounds like a witch to me :P

Larandar:
With terraforged swamps are not that common (the nearest for me was 13k block aways from spawn and was a few chunks only), the quest is IMO misleading since you don't need a swamp but only a witch.

NillerMedDild:
Hmm :/ I know it says visit a swamp, but it's just to solidy which kind of magic user you need to find. I guess we can change it.

NillerMedDild:
I'd like 

> "Try paying a visit to your local witch, giving her your exemplar of Ars Ecclesia, and if you ask nicely, you might learn something new..."

to be changed to 

> "Try paying a visit to your local witch and toss her your Ars Ecclesia, you might learn something new..."

Larandar:
Done 👍

NillerMedDild (approved):

NillerMedDild:
Thanks!

Merlo17:
FYI I initially left it vague on purpose, since the mod itself doesn't want to tell you straight up where the book should be thrown, and I feel like giving away the answer like that would spoil the fun for players who are legitimately interested in solving the little puzzle themselves (the ones who aren't will probably google the answer just as quickly)

I can see the swamp line being misleading and it should probably be changed, but maybe we can come up with something a bit more creative that doesn't ruin the process straight up?

NillerMedDild:
Yeah that'd be nice, I couldn't think of anything though :/

TemperedFool:
Something about "Double, double toil and trouble"?

Merlo17:
Maybe "find Vanilla's most magical mob, brewing spells (deep in the swamp if we feel like that's clear enough, or something else to end the line)

TemperedFool:
What, is my MacBeth reference too obscure? 😂

NillerMedDild:
... Yes 😄

TemperedFool:
For those who are curious: https://en.m.wikipedia.org/wiki/Three_Witches

PR #2731, opened 2021-07-14
[Expert] Initial RFTools Progression

TemperedFool:
Gating for RFTools. Leaving the machine base as it's default recipe (which is just stone and gold/copper nuggets), which I'm using for things I want to pull earlier than RFTools.

Machine Base - uses PNC Assembly Line, converts 2 Thermal frames into an intermediary, which is converted into an RFTools frame. Intermediate item needs art.

image

image

Screen Controller, Matter Transmitter/Receiver, Machine Infuser, Builder - converted to Machine Base (from Frame) to pull out of RFTools progression.

image

image

image

image

image

Module Template - requires a PNC PCB Board, gating it behind PNC; they will also need a Machine Frame for the environmental controller.

image

Charged Porter - also requires a PNC PCB, gating easy teleport from anywhere back to a matter receiver.

image

Builder's Pump & Quarry cards - Gated with PNC PCB, may need to be pulled further back. This leaves the builder available earlier for actual building and voiding purposes.

image

image

Shield Projector - Uses IF Dissolution Chamber so will require significant progression. Took the time to build out a nicely templated dissolution chamber script that can be used to refactor the base script.

image

Dimension Builder - Create Mechanical Crafting, using Mek Ultimate Circuits/Atomic Alloys. Should gate it pretty late.

image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/dissolution_chamber.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/dissolution_chamber.js
@@ -1,0 +1,71 @@
+onEvent('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
+
+    const recipes = [
+/*
+        {
+            inputs: [
+                '', // top left
+                '', // top 
+                '', // top right
+                '', // left
+                '', // right
+                '', // bottom left
+                '', // bottom
+                ''  // bottom right
+            ],
+            inputFluid: '', // optional
+            inputFluidAmount: 0, // leave at 0 for no fluid
+            processingTime: 50,
+            outputItemName: '',
+            outputItemCount: 1,
+            outputFluid: '', // optional
+            outputFluidAmount: 0, // leave at 0 for no fluid
+            id: ''
+        }
+*/
+        {
+            inputs: [
+                'minecraft:redstone',
+                'rftoolsbase:machine_frame',
+                'minecraft:redstone',
+                'minecraft:redstone_torch',
+                'minecraft:redstone_torch',
+                '#forge:ingots/gold_copper',
+                '#forge:ingots/gold_copper',
+                '#forge:ingots/gold_copper'
+            ],
+            inputFluid: 'tconstruct:molten_obsidian',
+            inputFluidAmount: 1000,
+            processingTime: 50,
+            outputItem: 'rftoolsbuilder:shield_block1',
+            outputItemCount: 1,
+            outputFluid: '',
+            outputFluidAmount: 0,
+            id: 'rftoolsbuilder:shield_block1'
+        }
+    ];
+
+    recipes.forEach((recipe) => {
+        let ingredients = [];
+
+        recipe.inputs.forEach((input) => {
+            ingredients.push(Ingredient.of(input));
+        });
+
+        const re = event.custom({
+            type: 'industrialforegoing:dissolution_chamber',
+            input: ingredients,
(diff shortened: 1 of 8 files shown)


TemperedFool:
After discussing with Mute I may update toy with some of these to add some magic requirements later, but for now I'd like to leave it as is and get the new recipetype files in.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/dissolution_chamber.js
```diff
@@ -0,0 +1,71 @@
...
+                'minecraft:redstone_torch',
+                '#forge:ingots/gold_copper',
+                '#forge:ingots/gold_copper',
+                '#forge:ingots/gold_copper'
+            ],
+            inputFluid: 'tconstruct:molten_obsidian',
+            inputFluidAmount: 1000,
+            processingTime: 600,
+            outputItem: 'rftoolsbuilder:shield_block1',
+            outputItemCount: 1,
```
NillerMedDild, line 44: I recommend merging outputItem & outputItemAmount into outputItem. It would like like this:

```js
output: { item: 'rftoolsbuilder:shield_block1', count: 1 }
```

and usage
```js
output: recipe.output,
```

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/assembly.js
```diff
@@ -0,0 +1,44 @@
+onEvent('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
+
+    recipes = [
+        {
+            input: 'thermal:machine_frame',
+            input_count: 2,
```
NillerMedDild, line 9: see previous recommendation.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/assembly.js
```diff
@@ -0,0 +1,44 @@
...
+            output: 'kubejs:rftools_frame_parts',
+            output_count: 1,
```
NillerMedDild, line 11: see previous recommendation.

NillerMedDild (changes requested):

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/dissolution_chamber.js
TemperedFool: Can do; was trying to make it "simpler" but possibly went too far ^_^

TemperedFool: Can also do the same for input items where appropriate.

NillerMedDild: It is a bit simpler, but it doesn't support tags which is unfortunate.

I think it works fine for input fluids as you made it though :)

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/assembly.js
TemperedFool: This (and below) are actually actually pulled from @MuteTiefling's base script; I'll also update that while I'm at it.

TemperedFool:
That should do I think!

NillerMedDild (approved):

NillerMedDild:
That's better, thanks!

Also, very good work on the recipes/progression :D

PR #2803, opened 2021-07-23
Industrious & Thermal Bees (and other updates)

TemperedFool:
Adding the Industrious Bee - crafted in the IF dissolution chamber (to a bee in a jar). Combs produce liquid latex, pink slime balls, and Meaty Honey. Meaty honey adds lots of saturation with a chance of a few other effects. The bee has a number of mutations that feel on-theme for IF, including mutating passives in a loop (Sheep -> Rabbit -> Pig -> Cow -> Chicken -> Sheep). Uses an acacia log as a flower.

image

image

image

image

image

Add Blizz/Blitz/Basalz bees. They still need textures, but they have the appropriate particle effects. They are all mutated from the proper mob by the Blaze Bee, and their combs put out rods and appropriate material chunks or snowballs. They're using Tinkers cakes as flowers, and moved Blaze bee to the Magma Cake. The only one that seems a little off brand is the Blizz Bee, using the Enderslime Cake; happy to take other suggestions.

image

Updated the Wasabee - it can now be created by mutating a Cat with a Water Bee - finding a Kitten Bee to breed one is a pretty hard gate. This does mean it is possible a mutation will occur in-world, if a water bee happens to fly over a village that has a cat.

image

Updated Resourceful Bees constants to add a bunch of the things I've added to RB but not added to their related constants.


TemperedFool:
@NillerMedDild Ok, this should be good to go with temp Thermal Bee coloring.

NillerMedDild (approved):
Very cool, gj :D

PR #2816, opened 2021-07-26
Nature's Aura Recipe Overhaul

MuteTiefling:
Oak trees consistently grow in ways that NA cannot detect, leading to frustrating failures about a quarter of the time...

Quark saplings do not have this issue. So all recipes have been re-written to use them instead as the growth catalyst. Figure they're abundantly available, so shouldn't cause a big issue for people getting into the mod. Lemme know if you'd prefer different trees though.

Also expands a few recipes either allowing tags where not previously supported, or allowing extra item types (expanded meats for Token of Sorrow).

Also reworks all Expert recipes so they have differing completion times.


NillerMedDild:
Great bug fix solution, good work :D

PR #2827, opened 2021-07-28
Rework Neptune's Bounty drops

MuteTiefling:
Normally with aquaculture, any excess armor/tools you fish up but don't want can be smelted down to recover a bit of neptunium. We've disabled that, however, so many times a neptune's bounty chest is just useless duplicate gear.

This would change that, removing all direct armor drops and instead giving nuggets directly. This sets the value pretty low by default, averaging out to a little less than we get currently, but evening out the distribution a bit. It also lets Luck apply to the rolls, potentially boosting yields greatly.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/loot_tables/aquaculture/neptunes_bounty.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/loot_tables/aquaculture/neptunes_bounty.js
@@ -1,0 +1,45 @@
+onEvent('server.datapack.high_priority', (event) => {
+    let loot_table = {
+        pools: [
+            {
+                rolls: 1,
+                entries: [
+                    {
+                        type: 'loot_table',
+                        weight: 1,
+                        name: 'aquaculture:box/neptunes_bounty_junk'
+                    }
+                ]
+            },
+            {
+                rolls: { min: 1, max: 5 },
+                bonus_rolls: { min: 0, max: 3 },
+                entries: [
+                    {
+                        type: 'item',
+                        weight: 1,
+                        name: 'aquaculture:neptunium_nugget',
+                        functions: [
+                            {
+                                function: 'set_count',
+                                count: { min: 1, max: 4 }
+                            }
+                        ]
+                    }
+                ]
+            },
+            {
+                rolls: 1,
+                entries: [
+                    {
+                        type: 'loot_table',
+                        weight: 1,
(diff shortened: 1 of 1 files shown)


NillerMedDild:
I feel like this is a pretty hard nerf, is that really needed?

MuteTiefling:
I had the opposite intent, really. I shoulda put numbers :D 

I was trying to get the full set of armor the other day and kept getting fishing rods instead. It was a pretty unfun experience, constantly getting the same item, and I felt it could have been nicer to just get a little something to use towards the gear, rather than nothing at all. 

Out of 128 rolls, in a test world just now, I got a loot spread looking like this: 

image

Averaging about 0.4 ingots per box and quite possibly an unwanted tool in each box that didn't result in an ingot.

Here's 128 rolls with the proposed numbers: 

image

Which actually comes out to 0.8 ingots per box. Not sure why my late night brain was coming up with this as less than the current output. Luck buffs would only bring this up, as well.

MuteTiefling:
Oh, right. I remember what I was doing for the calculation last night. I was counting each piece of gear as the equivalent number of ingots (8 for the chest, etc) and using that in the 'average ingot per chest' which, is fair up until you'd have all the gear. after that, of course, their worth drops to nothing.

Anyway, I'm just trying to come up with a non-recycling method of improving the neptunium gathering experience. I understand your reasons for removing recycling recipes in general, but the loss of those recipes for neptunium gear makes an already rare material feel harder to obtain.

TemperedFool:
Just to offer another opinion, this does seem like the perfect place to utilize recycling, if anywhere. Maybe one or two nuggets per ingot in the neptunium item?

MuteTiefling:
Yeah, the mod defaults to, I think 1 nugget per item in the blast furnace. Nothing crazy, but it gives something back for those missed rolls

NillerMedDild:
Ah okay, thanks for the numbers and images 😄 Would you personally prefer this new loot table, or that the gear was recyclable?

MuteTiefling:
I think I'd rather have recycling, to be honest. I can put that together instead if you're ok with it.

NillerMedDild:
Yes please :D but I'd like to get a release out within the hour, do you have time now? If not I don't mind doing it :)

MuteTiefling:
ah, no, I don't have time right now

NillerMedDild:
Done :)

PR #2839, opened 2021-07-29
Create Sequenced Assembly Script

### LINKED ISSUE #2835: You are unable to craft Percision Mechanisms in 0.5.7
The precision Mechanism is one of the new Sequential craftings added in the new version of Create, it is needed to make mechanical arms. By default there is a recipe for it but it appears to be missing. I checked against a pack containing just create(and flywheel) and the recipe is there and works. The other two sequential crafting recipes work for making cogs, but the recipe for precision mechanisms doesn't work,

TemperedFool:
Fixes #2835. I hate to ping you @NillerMedDild, but you may want to review this (and push an update) quickly, as this issue makes the Mechanical Arm and Rotational Speed Controller impossible to craft.

The default Create recipe for the Precision Mechanism has a chance output for "create:golden_sheet", which seems to be invalidating the recipe since we've removed it.

`{
      "item": "create:golden_sheet",
      "chance": 8.0
    }`

Implemented a clean-ish script for doing Sequenced Assembly, and am overriding all 3 default recipes to fix the above issue and remove the chance of "salvage" - for these at least the idea of chance failure seems like a super frustrating mechanic for a factory.

image

image

image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/sequenced_assembly.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/sequenced_assembly.js
@@ -1,0 +1,147 @@
+onEvent('recipes', (event) => {
+    const recipes = [
+        /*{
+            input: 'input_item_here',
+            output: [
+                Item.of('6x create:large_cogwheel').withChance(32.0), //withChance sets a weight for the output, default is 1 without it
+                Item.of('secondary_outputs').withChance(2.0),
+,               'more_secondaries_with_weight_1'
+            ],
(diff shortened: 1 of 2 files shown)


TemperedFool:
Also realized the Dissolution Chamber base script has an expert check I left in; that last commit fixes that. Also pretty important...

Quezler:
> super frustrating mechanic for a factory

especially since refined storage doesn't seem to have an equivalent of the ae2 level emitter with a crafting card

NillerMedDild (approved):

NillerMedDild:
Looks good, good work!

PR #2841, opened 2021-07-29
JEI description contradition for the `bottle of time`

Quezler:
First i tried it with a splash potion we (playing on a server) obtained somehow, when that didn't work i bartered with some piglins for a bottle of fire resistance, which also didn't work.

after digging through the code it appears that only the potion of swiftness 8:00 can be used.
(as shown correctly in the recipe for the bottle of time, but not in the tab suggesting otherwise)

https://github.com/NillerMedDild/Enigmatica6/blob/2e7ea246a8a88448b0889e91ff7a14cdd8dff5ed/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/interactio/item_anvil_smashing.js#L14

(so either the tip is wrong/outdated, or the recipe changed (or forgot to) sometime after, can't really tell from the git blame)

Screen Shot 2021-07-29 at 19 40 45
Screen Shot 2021-07-29 at 19 40 40

### DIFF
--- a/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
@@ -4,10 +4,6 @@ onEvent('jei.information', (event) => {
     }
     const recipes = [
         {
-            items: ['tiab:timeinabottle'],
-            description: ['Any Potion can be used for crafting this.']
-        },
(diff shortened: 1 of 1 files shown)


NillerMedDild:
I guess the recipe actually works as intended now, it didn't use to :P thanks!

PR #2897, opened 2021-08-06
mekanism quest improvements

Catchears:
reasoning for my changes:

 - **add osmium to ore multiplication**: why should osmium not work?
 
  - **change infuse types to item form**: no one wants to get a pressurized tube and pump the infuse into the infuser, using enriched items is far easier and does the same thing
  
 - **increase amount of tubes / pipes / cables to 32 from 16**: 16 never feels like enough to be a suitable reward while 32 might leave you with some left over for next time
 
 - **downgrade the induction cell to elite from ultimate**: with more than one player you end up with more of these than you'll ever need, so it seems fair to nerf that just a little
 
 - **add the advanced induction provider to quest rewards**: more transfer speed is always good; if induction cells can be given then why not providers?
 
 - **move the elite tier installer to the epic loot box (from rare)**: you get the rare loot boxes so early on that you can't even use the elite installer for quite a while, so moving it to the epic loot box (which also can drop an ultimate tier installer) makes sense
 
 - **add basic and advanced tier installers to the rare loot box rewards**: these will be far more useful than the elite installers that were here previously and also serve as a replacement
 
 - **remove biomass and fungi from quest rewards**: be honest, have you ever wished you had 128 buckets of fungi in your world? it's just disappointing to get this as a reward

all changes were tested in game.

### DIFF
--- a/config/ftbquests/quests/chapters/mekanism.snbt
+++ b/config/ftbquests/quests/chapters/mekanism.snbt
@@ -533,7 +533,7 @@
 								Count: 1b
 							}
 							{
-								id: "mekanism:osmium_compressor"
+								id: "mekanism:`osmium`_compressor"
 								Count: 1b
 							}
 						]
@@ -837,6 +837,10 @@
 								}
 								{
 									id: "mekanism:clump_copper"
+									Count: 1b
+								}
+								{
+									id: "mekanism:clump_osmium"
 									Count: 1b
 								}
 								{
@@ -2423,6 +2427,10 @@
 									Count: 1b
 								}
 								{
+									id: "mekanism:shard_osmium"
+									Count: 1b
+								}
+								{
 									id: "mekanism:shard_tin"
 									Count: 1b
 								}
@@ -2532,6 +2540,10 @@
 							}
 							{
 								id: "mekanism:crystal_copper"
+								Count: 1b
+							}
+							{
+								id: "mekanism:crystal_osmium"
 								Count: 1b
 							}
 							{
--- a/config/ftbquests/quests/chapters/mekanism.snbt
+++ b/config/ftbquests/quests/chapters/mekanism.snbt
@@ -533,7 +533,7 @@
 								Count: 1b
 							}
 							{
-								id: "mekanism:`osmium`_compressor"
+								id: "mekanism:osmium_compressor"
 								Count: 1b
 							}
 						]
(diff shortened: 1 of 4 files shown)


MuteTiefling:
> * **add osmium to ore multiplication**: why should osmium not work?

Simple oversight. Will be pushing a fix soon.

> * **change infuse types to item form**: no one wants to get a pressurized tube and pump the infuse into the infuser, using enriched items is far easier and does the same thing

You don't need to use a pipe, you can drop them in the machines as tanks. 

The reason behind giving these things as tanks is two fold: 
1) Free tank!
2) It's a learning opportunity to show people that these things can be moved around by tank, which could lead to alternate automation methods.

> * **increase amount of tubes / pipes / cables to 32 from 16**: 16 never feels like enough to be a suitable reward while 32 might leave you with some left over for next time

It's already 2 crafts worth, that's quite generous enough. This chapter is already exceedingly generous in it's rewards. 

> * **downgrade the induction cell to elite from ultimate**: with more than one player you end up with more of these than you'll ever need, so it seems fair to nerf that just a little. 

On the fence there, but I could see nerfing this one a bit. Probably just by making it rarer in the loot table, really. 

> * **add the advanced induction provider to quest rewards**: more transfer speed is always good; if induction cells can be given then why not providers?

These would quickly get into 'feels-bad-man' territory as they wouldn't be useful very quickly. One ultimate already is more than enough for most power gen in the pack. Even a single Advanced one outpaces most power gen... Having 2 wouldn't even be of use outside of people running fusion.

> * **move the elite tier installer to the epic loot box (from rare)**: you get the rare loot boxes so early on that you can't even use the elite installer for quite a while, so moving it to the epic loot box (which also can drop an ultimate tier installer) makes sense

Maybe. These being early is kind of meant as a spur to get you upgrading. You've got this shiny toy but you can't quite use it yet, so you press on. Feels worse getting these once you're already capable of mass producing them. 

> * **add basic and advanced tier installers to the rare loot box rewards**: these will be far more useful than the elite installers that were here previously and also serve as a replacement

See above. Goal isn't to give you everything, but to drive you onward.

> * **remove biomass and fungi from quest rewards**: be honest, have you ever wished you had 128 buckets of fungi in your world? it's just disappointing to get this as a reward

This was a bit of a troll reward. But, fungi and biomass can be a great way to obtain things like mycelium and podzol, which was their main intent. "Hey, what's this thing useful for? Ooooh." Void it if you don't want to use it, however. 

> 
> all changes were tested in game.

Catchears:
Ok, thanks for your opinion. I respect it and you seem to have things under control - I'll leave it to you then!

PR #2968, opened 2021-08-12
Require 16 stone bricks instead of 8 for nature altar quest

Quezler:
i prepared 8 stone bricks to build it with the schematicannon, only to be surprised that it needed 16 in total:

Screen Shot 2021-08-12 at 21 15 29


NillerMedDild:
Thank you kindly!

PR #2974, opened 2021-08-13
Alternate recipe for antimatter through powah

Quezler:
on my current playthrough server we used powah this time around instead of one of our basemates setting up mekanism reactors, we slowly made out way up to nitro & eventually i also automated all the generators from industrial foregoing in order to activate the 25m/t mycelium reactor.

during this run i also grew (extra) fond of the infinity tools, and would thus like to charge them, but at the rate of 25m per tick it would appear to take 60+ ish real life years? not really feasible.

it would therefore be nice to have a powah/other way of obtaining antimatter and/or artifact tier infinity tools, this draft accomplishes it by adding a direct chorus -> antimatter recipe in the empowerer, leaving the rest or it as-is.

(i picked chorus just because it is also purple & came to mind, i don't think its _that_ out of place here)

originally i had picked 10 billion since [100 would be needed to process 10 buckets into a full charge](https://github.com/NillerMedDild/Enigmatica6/blob/29398af2801055b711dffa09f7d04eb14b34b476/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/mekanism/nucleosynthesizing.js#L23) but that took kinda long, so i lowered it to 1 billion, in my testing setup this seems quite fair, since 15 energizing rods for powah is getting on the silly side, and a too high number would prompt for a pincushion, below are some of the numbers of my testing setup:

`15 nitro rods = 60k/t = ~14 minutes`

Screen Shot 2021-08-13 at 19 31 46

ps: i have no experience in how long it takes to get antimatter through mekanism the long way round, but i am guessing that this recipe will take longer in comparison, and tbh that sounds fair since you're skipping the mekanism reactors entirely :3

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/powah/energizing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/powah/energizing.js
@@ -44,6 +44,14 @@ onEvent('recipes', (event) => {
                     item: 'powah:spirited_crystal_block',
                     count: 1
                 }
+            },
+            {
+                ingredients: [{ item: 'minecraft:chorus_fruit' }],
+                energy: '1000000000',
+                result: {
+                    item: 'mekanism:pellet_antimatter',
+                    count: 1
+                }
             }
         ]
     };


NillerMedDild:
I'd rather not merge this tbh, I think the challenge to get Artifact tier tools is good as-is, this would make it quite a bit less challenging as far as I can see.

I appreciate the sentiment though!

TemperedFool:
To think I didn't get into GitHub to comment before this closed. I agree this seems too simple, but is there an alternate not-quite-too-easy route that uses nitro-tier materials to make polonium for an SPS? That way the general Mekanism mechanic doesn't get bypassed, but we provide an alternate route for people who don't want to do Mek fission. Just a thought.

NillerMedDild:
I'm open to the idea, but I don't think Powah has any crafting mechanics complex enough on it's own.

Quezler:
> I'd rather not merge this tbh

figured, that's why i left it as a draft, just hoped to spark some discussion that would lead to an alternate route 😉 (it worked)

Ridanisaurus:
We could look into using the Light and Void elements from Create maybe? I honestly can't remember their names
Something along the lines of 1x Light Element + 1x Void Element + 1x Uranananite in an Induction Smelter with a specific catalyst like a Nitro gem. This way we shift the complexity and progression from Mekanism Fission to Create end-game items, Powah end-game item, and Thermal. A combination of these might be somewhere near the complexity and requirements of Mekanism Fission for an alternative recipe for those that want to avoid Mekanism all together but still want their Antimatter

RobuxShooters:
> We could look into using the Light and Void elements from Create maybe? I honestly can't remember their names Something along the lines of 1x Light Element + 1x Void Element + 1x Uranananite in an Induction Smelter with a specific catalyst like a Nitro gem. This way we shift the complexity and progression from Mekanism Fission to Create end-game items, Powah end-game item, and Thermal. A combination of these might be somewhere near the complexity and requirements of Mekanism Fission for an alternative recipe for those that want to avoid Mekanism all together but still want their Antimatter

While this is a good idea, I do not think that this recipe is expensive or dangerous enough. Generating antimatter should consume quite a hefty amount of energy, be fairly complex, and somewhat dangerous and difficult to maintain. Perhaps to add onto this complexity, we can use some of the following?

- Void Worm (Alex's Mobs) drops or some Alex's Mobs equipment that uses drops from multiple different mobs.
- One of any curio from Artifacts (Artifacts to make Artifact-tier tools?) per Antimatter Pellet or 2 Pellets
- Molten Nitro Crystals instead of Nitro Crystals as well as possibly Molten Shadow Steel or Refined Radiance
- An artifact from Atum.
- Ether Gas
- Some Magical aspects to somehow condense the trace amounts of antimatter into usable form. (Ars? BM? AS?)

PR #3012, opened 2021-08-16
Bee Crafting - move to Jarred Bees

TemperedFool:
REQUIRES at least kubejs-forge-1605.3.15-build.80 and NBT-Ingredient-Predicate-1.3. I'm leaving this as a draft to start as we need to wait until we're ready to upgrade both.

Moves all bee crafting recipes using spawn eggs to use `.weakNBT()` with equivalent jarred bees. If we stick to this for bees that are crafted things should be more consistent - getting recipes for empty bee jars in JEI should bring up all crafted bees in that case since the only difference is NBT. This is possible via the weak-matching provided by the above mods, and will allow a bee picked up from the world in a jar to match despite having 45-46 tags vs. the 3 tags in the recipe. Once we've updated to Resourceful Bees 0.9.9.x we should even be able to bring the number of tags in the recipes down to 1.

Note: Industrial Foregoing's Dissolution Chamber does not appear to work with `.weakNBT()` ingredients, so I've adjusted the recipe to otherwise exclude the spawn egg and explicitly require Iron Bee products & an advanced machine frame. Should get the job done while I investigate with Max and Buuz.

Recipes now look as follows:

Elven Bee:
image

Mana Bee:
image

Industrious Bee:
image

Starry Bee:
image

Bloody Bee:
image

Terrestrial Bee:
image
and:
image

Nature's Aura Bees:
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/astralsorcery/altar.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/astralsorcery/altar.js
@@ -1,7 +1,7 @@
 onEvent('recipes', (event) => {
     const recipes = [
         {
-            output: Item.of('resourcefulbees:starry_bee_spawn_egg', 1),
+            output: Item.of('resourcefulbees:bee_jar', {Entity: "resourcefulbees:starry_bee", BeeType: "starry", Color: "#002184"}),
             pattern: ['_____', '__E__', '_DCB_', '__A__', '_____'],
             key: {
                 A: {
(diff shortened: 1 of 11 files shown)


NillerMedDild:
Nice! I've updated both in-dev, ready to merge :)

NillerMedDild (approved):

TemperedFool:
As soon as I've got the normal Terra plate recipe working I will!

TemperedFool:
Alright, everything is in. I updated the initial comment to add the now-working Terrestrial Bee recipe.

PR #3044, opened 2021-08-20
List names of mods excluded in the server installer

Quezler:
https://github.com/Yoosk/ServerStarter/blob/8a6ab8fe2592b2d5f7908f42eb1ed8124e512975/server-setup-config.yaml#L67

was digging through the serverstarter repo to try and make one for a pack that didn't have one, and i noticed that in their above example they added the names of the mods in comments after the id to better keep track.

i took the liberty to do the same to the server configuration file for enigmatica 6 to make it more easily maintainable.

(funny enough ReAuth/237701 was in there twice, thanks to this not anymore, also make it easier to see which is missing :3)

### DIFF
--- a/server_files/server-setup-config.yaml
+++ b/server_files/server-setup-config.yaml
@@ -58,15 +58,14 @@ install:
     formatSpecific:
         # optional paramenter used for curse to specify a whole project to ignore (mostly if it is client side only)
         ignoreProject:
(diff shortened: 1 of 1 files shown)


NillerMedDild (approved):

NillerMedDild:
Thanks!

PR #3070, opened 2021-08-24
[Expert] Proof of Concept - Assembly Table Circuits

MuteTiefling:
Just a single recipe to showcase how I think we can move some more complicated recipes into the Assembly table.

image

image


NillerMedDild:
I'm not 100% sold on the numbers here, but it's a cool concept 👍

MuteTiefling:
Fair! We can always tweak it as we get deeper into tech progression

PR #3099, opened 2021-08-27
Fix Base Resourceful Bees Shaped.js

### LINKED ISSUE #3089: Refined storage autocraft conflicts
### Modpack Version

0.5.11

### Describe your issue

I was not able to autocraft raw tuna from sushi go crafting with the ressourceful bee recipe.

There are multiple results from the same craft (cucumber, nori sheets and raw tuna) :
image

And the pattern doesn't allow to choose between results :
image

I think it is related to the fact the patterns don't store the result in its NBTs :
```
{
    Input_4:
    {
        id:"resourcefulbees:wax",
        Count:1b
    },
    Version:1,
    Input_0:
    {
        id:"resourcefulbees:wasabee_honeycomb",
        Count:1b
    },
    Exact:0b,
    Input_6:
    {
        id:"resourcefulbees:wasabee_honeycomb",
        Count:1b
    },
    Processing:0b
}
```

This applies to every other "multi-result" crafting recipe that I tried.
The easy fix would be to make an unique crafting recipe for every item, to make autocrafting available.

### Crash Report

_No response_

### Latest Log

_No response_

Yes

### User Modifications

**Server :** 
Installed server-side optimisation mods : 
- starlight-forge-1.0.0-RC2-1.16.5.jar
- sodium-forge-2.9.10.jar

Updated server starter to 2.1.0

**Client :**
Unmodified

Both

TemperedFool:
Fixes #3089 - for the initial RB Recipe conflicts anyway. Was an issue with vanilla shapeless recipes automatically having righthand whitespace trimmed & being mirrored left/right. All recipes should be unique now.

Also added recipe ID's to match the convention we've been discussing: `enigmatica:*mode*/*mod*/*item*_*descriptor*` It's a little wordy, but I think it will make it very easy to keep things clear; we could also consider code to automate it rather than specifying it - I'll look into that if I have some free time.

image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/resourcefulbees/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/resourcefulbees/shaped.js
@@ -16,112 +16,126 @@ onEvent('recipes', (event) => {
         {
             // Wasabee Comb Recipes
             output: Item.of('sushigocrafting:rice',3),
-            pattern: ['A  ', '  B', ' A '],
+            pattern: ['AAB'],
             key: {
                 A: 'resourcefulbees:wasabee_honeycomb',
                 B: 'resourcefulbees:wax'
-            }
+            },
+            id: 'enigmatica6:base/sushigocrafting/rice_combs'
         },
         {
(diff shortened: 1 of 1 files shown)


NillerMedDild (approved):
Looks good to me!

PR #3105, opened 2021-08-28
[Expert] Progression changes to Botania/Astral, more recipe tweaks. 

MuteTiefling:
Slight re-order to Astral/Botania progression. Elven Gateway now requires Astral to make:
image

Altered starlight costs of several items in the Astral tables as well to promote more extensive starlight arrays. Some key items now require solid starlight generation as they need nearly full bars. 

Reduced cost of Wooden Stands. Now costs half as much Arcane Gold to produce the full set, making it significantly less grindy to build with the manual Cauldron process for Arcane Gold
image

Moving Occultism lenses to PNC:
image

Results in 3 pair:
image

These find their way into Astral Lenses later on, hence the cost reduction. More lenses per shadow gem this way. 

Update Pedestals recipes for new tokens:
Wireless Energy Transmitter
image

Arrow Turret:
image

Breeder:
image

Smart Crafter:
Simply disabled the base recipe, leaving only the upgrade from the regular crafter upgrade. 

Mantle of the Stars:
image

Silky Cloth:
image

Magebloom Fiber
image

Blaze Fiber
image

End Fiber
image

Sophisticated Backpacks Upgrade Base:
image

Flipped orientation of Looking Glass craft. it was bugging me >.>
image

Replaced rods with arcane gold inlay for observatory:
image

Telescope:
image

Substantial Satchel:
image

Ender Pouch: 
image

Ender Chest:
image

Ender Tank: 
image


TemperedFool:
> Sophisticated Backpacks Upgrade Base:
image

Do we really want to gate Backpack upgrades? Or maybe more specifically, do we want to gate all of them? Seems like leaving some of them (basic pickup, maybe even the feeding upgrade, maybe a few others) as early access may be better for QoL without being particularly overpowered.

MuteTiefling:
It's a pretty light gate and they're awfully powerful. But yeah, it might make more sense to target some of the more potent upgrades.

Might be better to have it use the tier 1 ars fiber instead.

TemperedFool:
It might be silly, I just want to make sure we don't gate QoL tools too much; I'm totally fine keeping players mining and scraping by for a bit to gather resources, and making them explore (which it seems magic will do), just want to make sure they have good tools to do so.

MuteTiefling:
Yep. Definitely want folks to have QoL tools early ish. Goal here was to push it juuuust a bit outside of the 'I can craft this immediately' zone into 'this is a moderate investment and I have to work a bit to get there' zone. I do still feel the base recipe is too cheap. There should be a little bit of a struggle to get it, but I may have misjudged the struggle on this one.

TemperedFool:
Makes sense. Should we also offer a tech based recipe? My thinking is that for a few tools (backpacks and danks are the things that come to mind for me), if they're gated at all then maybe they should have a few options - let all the players who initially jump into Create have access too 😉

MuteTiefling:
this recipe requires tech and magic, as it stands. Can't make the fiber without a bit of a jaunt into Create to make a mixer. You also need some form of alloying to get the rose gold for the item frame. Kinda brings the message of 'you need to do both' very early on, I think.

TemperedFool:
Hrm, I missed that entirely. Carry on then 😅

NillerMedDild (approved):

PR #3115, opened 2021-08-28
Start server by utilizing screen for easier management

CollinHerber:
Currently when not using screen even when detaching from the console the server still shuts down. By utilizing screen you are given the ability to manage your started server easier and detach without closing down the server.

Possible Considerations

What distributions have screen preinstalled.
- [x] Ubuntu (Confirmed) 
- [ ] Debian (Unknown)
- [ ] Fedora (Unknown)
- [ ] CentOS (Unknown)

What about utilizing Systemctl instead?

### DIFF
--- a/server_files/start-server.sh
+++ b/server_files/start-server.sh
@@ -31,7 +31,11 @@ else
         echo "Neither wget or curl were found on your system. Please install one and try again"
     fi
 fi
-java -jar serverstarter-2.0.1.jar
+echo "Starting server in screen"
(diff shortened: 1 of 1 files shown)


NillerMedDild:
I do not have the required knowledge about Linux to review this properly to be honest. Additionally, I'd rather not mess with something that works, especially since accepting PRs for the startup scripts has been quite volatile in the past.

I appreciate the sentiment though :)

PR #3128, opened 2021-08-30
[Expert] Extended Initial Compressed Iron Recipe

TemperedFool:
Modifies Compressed Iron Ingot and Block gating. Rather than explosion crafting directly creating Compressed Iron, it produces Superheated Steel:

image
image

Superheated Steel can be packed in a Thermal Multiservo Press into Hot Compressed Iron:

image
image

Which can be cooled into Compressed Iron:

image
image

I'm not including any texture assets (despite the screenshots above), and will instead be begging Rid for art as usual!

This will be bypassed via the Pressure Chamber once it's constructed, which currently directly creates Compressed Iron. But... what if it didn't? I'm thinking that we have the pressure chamber convert Superheated Steel to Compressed Iron, maybe with packed or blue ice, at a better ratio than the initial gate - this would get the player a better production ratio & skip the multiservo press/water cooling steps.

Added bonus if we go this route - add a later Create Mixing superheated recipe to produce Superheated Steel. This would be much easier to automate than explosion crafting, and I would again give this a better production ratio than explosion crafting.

### DIFF
--- a/kubejs/assets/kubejs/lang/en_us.json
+++ b/kubejs/assets/kubejs/lang/en_us.json
@@ -107,6 +107,11 @@
     "item.kubejs.basic_circuit_package": "Basic Control Circuit Assembly Package",
     "item.kubejs.basic_circuit_assembly": "Unpacked Basic Control Circuit Assembly Package",
 
+    "item.kubejs.superheated_steel_ingot": "Superheated Steel Ingot",
+    "block.kubejs.superheated_steel_block": "Superheated Block of Steel",
+    "item.kubejs.hot_compressed_iron_ingot": "Hot Compressed Iron Ingot",
+    "block.kubejs.hot_compressed_iron_block": "Hot Compressed Iron Block",
+
     "item.kubejs.basic_lenses_package": "Basic Lens Assembly Package",
 
     "item.kubejs.direhouse": "DireHouse",
--- a/kubejs/assets/kubejs/lang/en_us.json
+++ b/kubejs/assets/kubejs/lang/en_us.json
@@ -110,7 +110,7 @@
     "item.kubejs.superheated_steel_ingot": "Superheated Steel Ingot",
     "block.kubejs.superheated_steel_block": "Superheated Block of Steel",
     "item.kubejs.hot_compressed_iron_ingot": "Hot Compressed Iron Ingot",
-    "block.kubejs.hot_compressed_iron_block": "Hot Compressed Iron Block",
+    "block.kubejs.hot_compressed_iron_block": "Hot Block of Compressed Iron",
 
     "item.kubejs.basic_lenses_package": "Basic Lens Assembly Package",
 
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipes/remove.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipes/remove.js
@@ -30,7 +30,9 @@ onEvent('recipes', (event) => {
         'darkutils:crafting/rune_damage_player',
         /emendatusenigmatica:alloy_dust/,
         'thermal:compat/refinedstorage/smelter_refinedstorage_alloy_quartz_enriched_iron',
-        'refinedstorage:quartz_enriched_iron'
+        'refinedstorage:quartz_enriched_iron',
+        'pneumaticcraft:explosion_crafting/compressed_iron_ingot',
+        'pneumaticcraft:explosion_crafting/compressed_iron_block'
     ];
 
     const outputRemovals = ['tiab:timeinabottle'];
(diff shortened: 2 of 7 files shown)


MuteTiefling:
> maybe with packed or blue ice, at a better ratio than the initial gate

Could make a Heat Frame cooling recipe for this too.

TemperedFool:
Hmm, would be cool (haha). Replacing the water step, rather than eliminating it you think? So final process would be Create Superheated mixing -> Pressure Chamber -> Chilling via heat frame?

TemperedFool:
Also, do we add code to make the heated items hurt the player? They ARE hot after all...

NillerMedDild:
> Also, do we add code to make the heated items hurt the player? They ARE hot after all...

Hmm maybe like half a heart every 5 seconds, but I don't think we have that kind of mojo from KJS.

I think the bypassing recipe should only use the Pressure Chamber, since the player does not have access to proper automation at this point in time. The larger, permanent production lines should wait until the player has the means to automate their way out of it, imo.

It's a cool production line though, maybe we can use it for something else down the line.

kubejs/assets/kubejs/lang/en_us.json
```diff
@@ -107,6 +107,11 @@
     "item.kubejs.basic_circuit_package": "Basic Control Circuit Assembly Package",
     "item.kubejs.basic_circuit_assembly": "Unpacked Basic Control Circuit Assembly Package",
 
+    "item.kubejs.superheated_steel_ingot": "Superheated Steel Ingot",
+    "block.kubejs.superheated_steel_block": "Superheated Block of Steel",
+    "item.kubejs.hot_compressed_iron_ingot": "Hot Compressed Iron Ingot",
+    "block.kubejs.hot_compressed_iron_block": "Hot Compressed Iron Block",
```
NillerMedDild, line 113: Should be `Hot Block of Compressed Iron` to match naming conventions for blocks of metal :)

NillerMedDild (approved):

Honhk:
> > Also, do we add code to make the heated items hurt the player? They ARE hot after all...
> 
> Hmm maybe like half a heart every 5 seconds, but I don't think we have that kind of mojo from KJS.
I mean the player can hold buckets of lava and due to iron being a conductor of heat should probably have a burnt hand, but here we are.

TemperedFool:
> I mean the player can hold buckets of lava and due to iron being a conductor of heat should probably have a burnt hand, but here we are.

Hmm, maybe we should fix that while we're at it too 😉

RobuxShooters:
Damaging the player in a set number of ticks while the player is standing on or holding something is possible with vanilla commands, though I'm not sure how feasible it is in modded Minecraft. Functions would have to be bundled with datapacks and run continuously every tick.

TemperedFool:
I'm almost positive KubeJS can check to see what a player is holding and/or what's in their inventory and fire off stuff based on that. Either way, although I would find it amusing, I'm not sure it's super player-friendly outside of a GT pack.

NillerMedDild:
> I'm not sure it's super player-friendly outside of a GT pack.

Yeah not so much :P It would have to be a trivial amount of damage.

I'm slightly worried about the performance implications of checking for the player holding a specific item every tick tbh.

Darkere:
How do I put this into perspective.... 

Every Living Entity checks every tick:
- if they are in a boat/minecart
- if they are inside a nether portal
- if they are in water
- if their eyes are in water
- if they are swimming
- if they are on fire
- if they are in lava
- if they are falling out of the world
- if they are/should be glowing

And this is without mods :D . Checking for an item every tick is nothing.

NillerMedDild:
I am amazed Minecraft works.

Thiana:
Please don't add the lava burns you thing. It doesnt add anything to the game except annoyance.

theboo:
I think we might be maxed out on buuz mods anyway

TemperedFool:
> Please don't add the lava burns you thing. It doesnt add anything to the game except annoyance.

Tempted to, I found it amusing in Omnifactory, but I am definitely in the minority. I won't add it if you don't 😉

> I think we might be maxed out on buuz mods anyway

I feel like we could fit one or two more in...

NillerMedDild:
@Thiana Would you also be against a tiny amount of damage from holding a `Superheated Steel Ingot`? :)

Thiana:
No. The problem with adding it to lava buckets is you use the stuff in a LOT of different ways and unless you add a heat mechanic where even approaching world lava causes similar damage it doesnt make sense. For survival type packs like Permafrost Eye of the Storm where it's handled consistently the mechanic makes sense but not for E6. 

For something like the ingot which isn't used constantly and is named in such a way the player has fair warning, I consider some damage to be reasonable.

;-P

NillerMedDild:
Cool, that makes sense 😀

PR #3137, opened 2021-08-30
[Expert] Textures for #3128

TemperedFool:
Courtesy @Ridanisaurus as usual...

image

I also apparently pushed my all of my quests data rather than just the Expert Tech one I updated - the rest of them shouldn't have actually changed. For Tech I added a bit on the thermal side going over the new Compressed Iron workflow.


NillerMedDild (changes requested):
This PR seems to change EOL for quest files to LF, please remove everything but the textures :)

TemperedFool:
When in doubt revert it out! Should be good now I think.

NillerMedDild (approved):

PR #3139, opened 2021-08-31
[Expert] Preliminary restrictions setup 

MuteTiefling:
Restrict soul forge to the nether. Can only be used once the hellfire forge has been crafted.

Key Blood Magic blocks restricted to Undergarden until they've crafted a master blood orb.


NillerMedDild (approved):

NillerMedDild:
Looks good to me :D

PR #3147, opened 2021-08-31
[Expert] Further Magical integrations, beginnings of Digital Storage

MuteTiefling:
New Fel Pumpkin to make people work a bit if they're gonna use these guys to obtain early blaze entities
image

Mana Infuser
image

Rune changes:
Niflheim
image

Muspelheim
image

Midgard
image

Alfheim
image

Reduced the rune needed for Luminous Crafting considering the above changes
image

Chromatic Compound:
image

Changed fertilizer upgrade for pedestals to rune of spring
image

Update Iridescent Altar
image

Update Common Tartaric Gem
image

Dimensional Storage Actuator
image

Update to Sentient tools, now need rune of helheim
image

Digital (dimensional) Storage!
image

image

Reminder of what Crystal Matrices involve
image

RS Disk Drive:
image

I've prepped for more recipes like this for larger disk drives, assuming we want to have them all assembled by machine as such. Assuming these aren't too evil :D


MuteTiefling:
64k Fluid storage part and crafting
image
image

NillerMedDild:
I think having to do an Occultism ritual per Storage Component is too evil tbh :P having to automate those should be a later task, imo.
Magic additions and Disk Drive look good, love what you did with the colors of chromatic compound ingredients!

MuteTiefling:
I was afraid you might say that for the storage disks :D 

How about a recipe that splits that matrix into 3 (or 9) Crystal Platters? Then we use that platter as the disk ingredient.

NillerMedDild:
That helps, but I still think it might be too evil. Can Demon Wills be automated yet in 1.16 BM? ^^'

MuteTiefling:
Not with BM, no. But you can do it nicely with a PNC drone holding a sentient sword and a tartaric gem

NillerMedDild:
Hmm..
Would you be willing to replace the Demon Wills for something more easily digestible for the player in the Crystal Matrices recipe? :P

MuteTiefling:
Sure, lemme give it a think on what could be suitable

MuteTiefling:
Alright. Crystal glass
image

Similarly removed the demon will crystals from the tier 1 stabilizer. Still needed for the tier 2
image

NillerMedDild:
Nice! Excellent swap.

I would still like for the Matrix to be splittable in 9 when used in Storage Parts, but I can live without that 😄

NillerMedDild (approved):

MuteTiefling:
I'll put the splitter together in a bit

MuteTiefling:
Thermal and mekanism sawing recipes for cutting the matrix down to storage crystals. Better yield from Mek
image

image

Packaging for the disk
image

Your comment on will crystals got me thinking we should probably implement something better for that as well since they're already used in some later crafts.

This allows them to be slowly grown. Any demonic crystal can be resonated down to a bit of dust. tiny dust is compacted to dust, 9:1
image

set recipe for the crystalline resonator, replacing the primitive one entirely
image


High heat and pressure to grow them back up. This is exothermic, so the player must strike a balance in heating and cooling.
image

So 3 crystals in gets about 2 powder, plus a little extra. 2 powder grows to 4 new crystals.

will flavors can then be infused
etching acid for corrosive
image
blood for vengeful
image
blazing blood for destructive
image
molten obsidian for steadfast
image

NillerMedDild:
Very very nice!!

PR #3151, opened 2021-09-01
Add basic Industrial Foregoing Ponders using PonderJS

MaxNeedsSnacks:
This PR introduces KubeJS scripts and ponders for two Industrial Foregoing concepts:

- Gathering Latex
- Resource farming using Laser Drills

I mostly chose these two because they were suggested to me by the other contributors, so just consider these more "examples" for using PonderJS rather than anything particularly useful to the average Joe (though I definitely tried to be as informative as possible :p)

Here's a preview of both ponders in action, as well as a quick look into what items can be used to trigger what ponders (1st video) and the overall Ponder Index (2nd video).

https://user-images.githubusercontent.com/12750501/131591962-736bb7d4-6bec-4b66-be74-61f4c22804a1.mp4

https://user-images.githubusercontent.com/12750501/131591966-e1a6848e-638c-4911-884f-ca7f86dd2d3a.mp4

Unfortunately, it seems some language keys didn't autogenerate correctly, so I won't be including them in this pull request (though, PonderJS **will** generate those itself when you launch the game yourself), and I also won't be including my modified minecraftinstance.json here (+PonderJS, +updated KubeJS) as that would most likely break your local instance ^^


MaxNeedsSnacks:
Quick addendum: It seems PonderJS messes up the formatting of the existing kubejs lang file, so you'll definitely want to go into the client config and change the file location to something else (maybe `enigmatica6/lang/...`?)

NillerMedDild (approved):

PR #3155, opened 2021-09-02
[Expert] Finishing out disks and processor proposal

MuteTiefling:
Raw Neural Processor
image

Alright. So making wixie charms shouldn't be too complicated to automate. Spell turret using "Projectile > Lightning > Dispel" on a villager from a villager breeder would do it. Or "Projectile > Harm > Dispel" on a witch from a spawner.

Lightning is abundant early game for manually converting villagers to witches, otherwise, and Dispel is a Tier 1 glyph. nothing difficult to reach and all things that would be available by the time a player can even make the Dimensional Storage Crystals.

These'll be used for upper tier crafters and upper tier disks. Potentially for other high tier crafts. This was basically the idea I had of capturing mobs and using them for advanced circuits, but without the need for masterful machinery yet or custom machines/items. I wouldn't mind doing something similar for some of the other processors, but I think those would need MM to do nicely.

Disks remain a tiered craft, similar to default, though now requiring increasingly higher tier circuits.
4k part
image

16k part
image

64k part
image

256k and up use the same processors and circuits
image

fluid disks follow the same pattern

Also, I'm thinking that the current Raw Processor > Processor might be neat to put into the Thermal Press with a numismatic dye like so:

image

It's ultimately not more complicated than smelting, but feels a little more techy.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/kubejs/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/kubejs/shaped.js
@@ -48,6 +48,8 @@ onEvent('recipes', (event) => {
                 C: 'mekanism:cardboard_box'
             }
         },
+
+        // Storage Parts
         {
             output: 'kubejs:1k_storage_part_package',
             pattern: ['AAA', 'BCB', 'DEF'],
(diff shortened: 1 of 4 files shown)


NillerMedDild:
Absolutely love it. We probably need to supply some guidance on the ways to achieve this automation though.

MuteTiefling:
Should be an easy clue to give in a quest, I think. When we get to filling out the expert quests, that is

MuteTiefling:
There, added a quest in the normal quest line for now. Worth it just to have a concrete example to add for the spell chalk

PR #3165, opened 2021-09-03
[Expert] Update theme to BM recipes, expertify ore processing, and more

MuteTiefling:
Since Blood Magic is initially bound to Undergarden, it seemed fitting to theme some more of the recipes around that.

Alchemy Table Anointments all now follow a pattern of Bottled Sunlight, a nugget of some sort, and an item specific to Undergarden, thus creating something by melding items from two worlds.

Honing:
image

Soft Coating:
image

Fortuna:
image

Holy Water:
image

Miner's Secrets:
image

Dexterity Alkahest:
image

Plunderer's Glint:
image

Iron Tip:
image

Slow-Burning Oil:
image

Each also has a new tooltip attached to help people discover their uses. They're covered in the BM book, but it seemed like they could use a little help.

image

Nocturnal Powder now uses ink mushrooms
image

Illumination Powder using glitter kelp instead of glow squid ink
image

Mining Reagent:
image

Blood Lamp Reagent:
image

Binding Reagent:
image

Essence of Death:
image

Makes Demon Stone/Bricks craftable in all modes. Expert uses Infused stone for base while normal uses `#forge:stone`
image

Variants available through stonecutting.

Updated Crystalline Resonator to use Demon Stone instead.
image

ARC
image

Ore Processing update:

Occultism Crusher Spirits getting more appropriate recipes compared to their power
Foliot (2x Ore)
image

Djinni (3x Ore)
image

Afrit (4x Ore)
image

Marid (6x Ore)
image

Also add NA aura to Atum and Undergarden as alternate places to set those up.


NillerMedDild:
Awesome stuff :D

PR #3167, opened 2021-09-03
Candles rework

theboo:
Quark white candle is now the base candle, with recipes to go from quark>occultism>eidolon>back to quark

Also changed the tag used in base recipe from `forge:wax` to `enigmatica:candle_materials` which i think just allows honeycomb to be used along with RB beeswax and tallow


NillerMedDild (approved):

NillerMedDild:
Good work btw :D

PR #3168, opened 2021-09-03
Switch to using Tinkers config for hiding table and anvil variants

DarkAngel22579:
Because KnightMiner told me to do it 😅
(Also hid the quark buttons from the smeltery)


NillerMedDild (approved):

NillerMedDild:
Fair enough :P thanks!

PR #3170, opened 2021-09-04
Refactor Base Shaped Crafting recipes

MuteTiefling:
All ID Replacement recipes are moved into their own script by mod.
Custom recipes are in the enigmatica folder and given new ID's matchin `enigmatica:base/modid/item_extra_description` where extra_description is added if necessary to avoid ID conflict.

Tried to clean up one off ID removals that were only there because we couldn't override at the time. Probably missed a few.


kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/enigmatica/shaped.js
```diff
@@ -0,0 +1,929 @@
...
+            },
+            id: 'enigmatica:base/morphtool/tool_filled'
+        }
+    ];
+
+    recipes.forEach((recipe) => {
+        event.shaped(recipe.output, recipe.pattern, recipe.key).id(recipe.id);
+    });
+
+    // Unification type scripts. Consider moving out?
```
NillerMedDild, line 691: I think that'd be a good idea, I'm not sure where to put them though.

NillerMedDild (approved):

NillerMedDild:
Nice work, thank you for doing this!

Issue #3173, opened 2021-09-04
Two By Two Advancement Showing Strange Results


zedekblue:
### Modpack Version

0.5.11

### Describe your issue

I have been working on the breeding & taming advancements and noticed some... strange progress in the advancements tab, which is all I really have to go off of for progress.
![Capture](https://user-images.githubusercontent.com/17040113/132106279-22b2f494-962b-4b68-9fa9-058a10025484.PNG)
As you can see, I have bred 44, yet it only counts 41. Furthermore, it shows 11 remaining, yet only 10 left until 51.

To make matters more complicated, as you can see in [The Spreadsheet I've made,](https://docs.google.com/spreadsheets/d/1xmT22Dc4zi0rOvUgPzFrbl4Zpvq8cOkdKrZ00W9d1Vo/edit?usp=sharing) I have successfully bred 7 additional mobs which did not count towards this advancement.

My question is:
Is this list autogenerated or manually created? Is this behavior intentional, with some mobs left out?
Can you confirm if those 7 mobs _should_ have counted or not?
Considering this is an official e6 challenge quest I am having a difficult time finding anyone else who has any information on completing it.

Regardless, I thought you should know this is happening in case it needs to be fixed.
Thank you!

### Have you modified the modpack?

No

### User Modifications

Server has been updated from 0.5.9 -> 0.5.10 -> 0.5.11 in case that may be a contributing factor.

### Did the issue happen in singleplayer or on a server?

Server

NillerMedDild:
Thanks for the report! @ochotonida do you have any light to shed on this? :)

ochotonida:
Thanks for the very detailed report, the spreadsheet helped a lot :).

The lists of mobs that count towards the taming & breeding advancements are manually maintained, so there's always a possibility I missed something. From the 7 mobs you couldn't get to count, cockroach is the only one that should have counted. Tarantula hawk breeding isn't possible to detect afaik, so I can't add that one to the advancement. The other 5 simply weren't added to the list. The book from Alex's Mobs doesn't mention breeding for 4 of these, so I'm not surprised I missed them.

The incorrect count is caused by environmental. Its mobs currently don't actually count toward the advancement despite showing up in the list, this is also fixed in my PR.

zedekblue:
> From the 7 mobs you couldn't get to count, cockroach is the only one that should have counted. 
[...]

PR #3180, opened 2021-09-05
Update animals.js

ochotonida:
This PR fixes a few issues with the breeding and taming challenges, as pointed out in #3173.
* Added deer, ducks, slabfish and yaks from environmental to the breeding advancement, and slabfish to the taming advancement. Normally, environmental already handles this, but it doesn't seem to be counting correctly with the custom advancement. Adding these manually fixed this.
* Added gorillas, grizzly bears, roadrunners and gazelles to the taming advancement. The book from Alex's mobs didn't mention these as breedable, so I assumed they weren't.
* Added rattlesnakes to the taming advancement. Not sure how I skipped over this one tbh.
* Added crows to the taming advancement, there used to be an issue where taming crows wouldn't trigger the advancement but this has been fixed in a recent Alex's mobs update.
* Removed bald eagles from the breeding and taming advancements. Bald eagles are disabled, which I missed.


NillerMedDild (approved):

NillerMedDild:
Nice work!

PR #3189, opened 2021-09-07
[Expert] Flight tweaks, custom rituals!

MuteTiefling:
Attempting to place flight options where they'll be more appropriate

Ars Flight tablets made a bit cheaper, as they're consumed, localized, and require resources to maintain:
image

Magic Feather: Limited range, requires beacons. So made relatively early.
image

Flugel Tiara
image

Magical Feathers
image
Yes, that requires an ender dragon sacrifice :D I've ensured all the blocks required for the ritual are immune to the dragon, so she doesn't destroy them as you attempt the ritual. Tested in survival a few times and it's a bit of a challenge since the ritual prevents you from right clicking (no bows, or instance). It's entirely doable, however, and I see this as a very late game challenge in any case. Good gear should be in abundance at this point.

I think this gives people a clear sense of flight progression. As their magical abilities grow, so too does their ability to fly. With the pinnacle of flight requiring mastery of essentially all magic mods.

Ars Ritual Rework
Sunrise:
image

Moonfall:
image

Binding:
image

BM time shifting alchemy array changed to match relevant ars rituals
image
image

Greedy Familiar:
image

Drikwing Familiar
image

Bat Familiar
image

Deer Familiar
image

Book: Just for the heck of it. Makes found books very valuable as crafting them requires machinery.
image


NillerMedDild (approved):

zedekblue:
Just my two cents, while magical feathers is an unlimited flight resource, it is also significantly slower than the flugel tiara. I personally removed feathers in favor of the tiara a long time ago on normal mode.

Since the tiara is only a temporary flight resource without the eye of the flugel, perhaps lock it behind additional or equivalent progression as the feathers.

MuteTiefling:
Not sure how I feel about altering the Botania Relics. The Eye just drops as part of grinding Gaia... we'd have to altar that to allow your suggestion, and that gets tricky due to how they're bound to the player. 

I know base Creative Flight is slower than other options, but as I understand it the major draw for a lot of people is that it's free to use. There's no mana to supply, no fuel. Doesn't take up a gear slot, etc. It just is on or off. The Eye/tiara combo has some nice perks to it, but at a cost of constant mana use, meaning infrastructure dedicated to it at home and a mana mirror, plus the two items. Three inventory slots and mana infrastructure kinda warrants being better, no?

I personally much prefer PNC Jetboots to either option as they suit my playstyle, for instance.

PR #3198, opened 2021-09-07
[Expert] Godforged tweaks

MuteTiefling:
Alters Godforged Blocks to look for a tag, rather than a specific item when breaking them for relics.

For normal mode, this is only applied to the Nebu Hammer, the normal tool for this. For Expert, however, I've added the Alf Shatterer and Alfsteel AIOT to the list since Nebu is used in their creation.

This just opens up a few extra possibilities for automation. May back out the previous Pharaoh summon. Thoughts?

Also updates NA quest chapter to cover some extra details concerning the Aura Field Creator.


MuteTiefling:
oh, hold up. need some jei descriptions on this

NillerMedDild:
> This just opens up a few extra possibilities for automation. May back out the previous Pharaoh summon. Thoughts?

I think the Pharaoh summoning ritual looked very interesting, I'd prefer that it stayed :D

NillerMedDild (approved):

MuteTiefling:
jei descriptions set!

NillerMedDild (approved):

PR #3199, opened 2021-09-07
Pondering Portals

TemperedFool:
Adds ponders for the Atum portal (to the Scarab) and Undergarden Portal (to the catalyst).

[Atum](https://cdn.discordapp.com/attachments/730169148294037544/884853045714894878/atum.mov)
[Undergarden](https://cdn.discordapp.com/attachments/730169148294037544/884895214894989352/undergarden.mov)


NillerMedDild:
Nice work!

PR #3206, opened 2021-09-09
[Expert] Rune re-shuffling and other tweaks

MuteTiefling:
Update to Blood Magic Runes. Didn't like the simple crafting recipes, so now they're made in the Runic Altar

Blank
image

Speed
image

Sacrifice
image

Self Sacrifice
image

Displacement
image

Capacity
image

Augmented Capacity
image

Orb
image

Acceleration
image

Some of these require runes that aren't craftable from Corundum, so I've added recipes to combine other Quark Runes to the colors that were missing:

image

Similarly, all of the runes crafted from corundum are moved to the Runic Altar
image

Also swaps in some BM Slates where appropriate. These are made from otherstone, essentially, so some previous recipes using otherstone now simply use slates of varying quality.

image

image

Shifted Arcane Stone to a clay base
image

Infused Rock now made from Arcane Stone instead of Firmament
image


NillerMedDild:
I think those recipes are better too, but the amount of runes given per recipe might need to change, depending on playtesting.

MuteTiefling:
Sounds good. Could maybe double em up.

PR #3212, opened 2021-09-09
[Expert] Concrete Curing

MuteTiefling:
Concrete Curing
image

Bit of a bonus for keeping it cooler. Get it down to 20c for a 10% increase in yield.
image
That's not particularly cold, but concrete doesn't usually like cold weather curing either. So just a little bonus if they take the time to cool things further. 

And JEI update for concrete cooling in world:
image
The default IE cooling happens faster than this though, so just for the sake of consistency.


NillerMedDild:
Perfect, thank you :D

PR #3217, opened 2021-09-11
Missing assembly recipes

MuteTiefling:
Not sure why, but the if statement for stacked item was never firing, leaving all recipes without the type. As a result, the entire script was failing silently.

Shouldn't hurt to leave it on every recipe. Also moved over to the new ID convention.


NillerMedDild:
I also ran into this issue, for me putting whether the item was stacked or not in the json helped. I like your solution better though :D

PR #3240, opened 2021-09-15
[Expert] recipe tweaks

MuteTiefling:
Turns out Hellfire forge does not like auto splitting inputs, so things that require 2 of something require 65 to be inserted at a minimum to work automatically. This should make automating these two recipes a bit less of a pain in the bum


NillerMedDild:
Nice catch!

PR #3242, opened 2021-09-15
[Expert] Adds tooltips to some custom summons and move Sentient tools to rituals

MuteTiefling:
image

image
Same cost for sentient tools, just shifted to a new mechanic

requires occultism-1.16.5-1.11.4 or higher.

also fixes a goof on the nebu hammer change as that was meant for expert only


NillerMedDild (approved):

NillerMedDild:
Very nice! Reading the Occultism changelog I did have a feeling you were involved :P

MuteTiefling:
:D

Issue #3246, opened 2021-09-16
Unable to set up a crafting recipe for Runed Marble in Refined Storage


NurikSakura:
### Modpack Version

0.5.11

### Describe your issue

As the title states, there is a problem with setting a crafting recipe for Runed Marble from Astral Sorcery. It's not that big of a deal, as it can be done through the Stonecutter (although manually) or through Create's Mechanical Saw (althoug this is quite an overkill setup fro one single block).

Here are some screenshots. On the second one, where JEI recipe is shown, the left and the right marbles loop through every type of marble, while the middle one stays Chiselled Marble all the time.

![RS Pattern Terminal](https://user-images.githubusercontent.com/408325/133528903-88c4ec46-5f85-4413-b97e-91f40b8d8450.png)
![JEI Recipe](https://user-images.githubusercontent.com/408325/133528906-b505e1c5-d96c-4ccf-92a1-4c8666ecadfe.png)

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Server

MuteTiefling:
should be all fixed up next update. thanks!

PR #3250, opened 2021-09-16
Remove secondary marble slab recipe

MuteTiefling:
Resolves conflict between runed marble
image

and marble brick slab
image


MuteTiefling:
Resolves #3246

PR #3256, opened 2021-09-16
Remove powah cables from quest rewards

MuteTiefling:
Makes powah cable quest optional. Might consider removing?

removes powah cables from quest rewards. replaced with basic universal


NeosGhost:
no expert recipe for powah cables?

Rodg88:
powah cables are disabled in expert mode yes.

PR #3266, opened 2021-09-17
Nori Sheet Ponder

TemperedFool:
Adds a 3-scene ponder for Nori Sheets, going over seaweed growth, smelting, packing, and pressing.

[Nori Sheet Ponder](https://cdn.discordapp.com/attachments/884949955410796564/888263054289158184/nori_sheets.mov)

I added an extra delay between the furnace and crafter in scene 2 after recording this video.

I also did a little reorganizing. After discussing it in the Ponders thread, I think we've settled on 5 categories - Portals, "Basics", Tech, Tools, and Magic, to mirror (somewhat) the way quests are laid out. We don't have any magic ponders yet, so no category for it yet. I've rearranged the layout for the scripts and base.js to mirror.


NillerMedDild:
Very nice :D

Issue #3271, opened 2021-09-18
Enriched Menril Log harvesting when ultimining Menril Logs


DesignatedT:
**Description**
There is a tree called Menril Tree which spawns these Enriched Menril Log blocks on the trunk. If you Ultimine(veinmine) part of the tree, those blocks aren't considered part of the regular Menril Log blocks. Is there a way to chuck those blocks together similar to how you did it with the different stones?

**Why would you like this added/changed?**
Ease of use or Quality of life.

**Screenshots if applicable.**

PR #3272, opened 2021-09-18
Allow ftb ultimine to harvest entire menril tree (excl. leaves)

Quezler:
> implements https://github.com/NillerMedDild/Enigmatica6/issues/3271

my attempt at trying to be helpful, went with the `ftbultimine:` tag namespace instead of `integrateddynamics:menril_tree` as to not confuse people where the tag originates, and to allow for future ultimine grouping to happen in the same place 🤔


MuteTiefling:
Looks good!

PR #3273, opened 2021-09-18
Allow Ultimine to break Filled and Regular Menril logs

MuteTiefling:
Allow Ultimine to break Filled and Regular Menril logs. Allows chopping a whole menril tree at once like any other tree

.

### DIFF
--- a/defaultconfigs/ftbultimine-server.snbt
+++ b/defaultconfigs/ftbultimine-server.snbt
@@ -11,5 +11,5 @@
 	
 	# These tags will be considered the same block when checking for blocks to Ultimine
 	# Default: ["minecraft:base_stone_overworld"]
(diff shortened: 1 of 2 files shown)


Quezler:
duplicate of (my) #3272 😅

MuteTiefling:
yeah, lol was just seeing that

Quezler:
(it would probably also have messed with [some recipes](https://github.com/CyclopsMC/IntegratedDynamics/blob/master-1.16/src/main/resources/data/integrateddynamics/tags/blocks/menril_logs.json) that were only intended for the normal log by the mod itself)

MuteTiefling:
nah, that's why I left it on block tags. Recipes don't use those. But yours is safer still

PR #3280, opened 2021-09-19
Hive Upgrade Type/Tier Recipes

TemperedFool:
Moves to using type/tier ingredients added for RBees 0.9.9.6 in recipes; this will allow hives upgraded in-world to be used as their tier for crafting recipes. The only downside is that this can confuse autocrafting a bit - technically, for example, a T3 Beehive that is actually T2 is a thing (although unobtainable), which makes it possible for RS to fail due to recursive recipes. This seems to only happen if it doesn't have recipes for the whole crafting chain, however.

If that happens, Refined Storage for example will pop an error like this:

image

We get enough questions/complaints about the discrepancy between in-world upgraded hives & crafted ones that I think this change is worth it, but it's not a sure thing, so I would definitely like review on it.


NillerMedDild (approved):

NillerMedDild:
Seems like I good decision to me. Can the Hive with multiple tiers have one of them removed?

PR #3286, opened 2021-09-20
[Expert] PneumaticCraft Expertification

MuteTiefling:
Remove solid fuel compressors. Fuel production will be handled by Thermal machines. Plenty of decent options here.

Primary reasoning for this is that solid fuel poses a significant problem to new players: It doesn't stop burning when you turn off the machine. Instead, it finishes that piece of fuel... which can put the machine's pressure into the danger zone quickly. Pop.

Liquid Compressors, on the other hand, stop instantly when disabled.

Bellows
image

Basic Liquid Compressor:
image

This comes with a Security Upgrade pre-installed.

Pressure Tubes:
image
Just a bit of hardened glass, and a boost to output.

Reduced cost of Tanks
image
Primary reasoning is because the new compressor recipe is very costly in terms of compressed iron. This brings it down a tad.

Removed default foundry controller (unrelated, just noticed it was still there)

Reinforced Stone:
image
Concrete Powder + molten compressed iron

Removed extra Pressure Chamber Valve recipes. Added Security Upgrade to remaining one
image

Pressure Chamber Interface:
image

Pressure Chamber Glass, now with hardened glass
image

Pressure Gauge:
image

Safety Tube Module:
image

Pressure Gauge Tube Module:
image

Capacitors
image

Transistors
image

Move the Printed Circuit Board assembly to Pressure Chamber:
image

Basic Control Circuit upgraded to use Advanced Processors, since the PCB has a basic one already
image

Advanced Compressor:
image
Also has a security upgrade, as the others.

Electrostatic Compressor, getting magical:
image

Thermal Compressor:
image

Flux Compressor:
image


MuteTiefling:
Disabled Compressors will drop some Reinforced Bricks, rather than disable the PNC Villages.


Fuelwood Heater. Alternative to the Vortex Tube for early heating. 
image

Similarly, the External Heater may be used:
image

Resistive Heater pushed back a bit as it's a bit -too- easy for PNC heat (dial it in and walk away). Can be something to work towards.
image


JEI Descriptions added for these blocks as well to let people know that they work as heat sources.

Empty PCB boards
image

Buffing Ether Gas as a coolant for PNC
image

NillerMedDild:
This is spot on what I wanted for the PNC chapter so far, brilliant work!

theboo:
Enigmotoko back at it again

MuteTiefling:
Alternate route to obtain Advanced PCBs. Having access to these early can really help with avoiding explosions and such since you can dial in Gauges to output at exact levels.
image
The alternate recipe is a bit expensive and cannot be automated since Logic Circuits have to be made by hand. So long term, the PNC recipe would likely be ideal. 

Changed Elite Centrifuge from Advanced PCB to just a regular PCB as a result of the above:
image

Turbine Blade: 
image

Increased output for programming puzzle. A small program can easily take up 100 of these things and with the increased cost of circuit boards, I don't want to discourage people looking to program some drones. 3x output per circuit. 
image

Assembly Table
Laser
image

i/o
image

controller
image

drill
image

platform
image

pneumatic dynamo
image

MuteTiefling:
More to come, but I'll do that in another PR, this one's already getting big :D

NillerMedDild (approved):

NillerMedDild:
Again, very very nice you've done here :D

PR #3294, opened 2021-09-21
[Expert] Mineral Survey Drones

MuteTiefling:
Requires pneumaticcraft-repressurized-1.16.5-2.14.3-255 or later. No official release yet, so may need to wait.

Introduces per dimension trades. As a starting point, I wanted to set up a way to get some dimension specific resources from Atum and Undergarden. I'm not 100% sure of the actual loot table, might be better to limit potential items a bit more... Perhaps to only the interesting materials from the dimension, or giving those a much higher weight. Thoughts?

As an initial balance, these trades are available 5 times per day and can only be requested from the appropriate dimension. Each one requires a steel paxel, filled air canisters, and a few other disposable items, as well as a chest everything will go in.

You have a 15% chance of getting a very damaged paxel back and 100% chance of getting the emptied air canisters back.

The chest then pulls a loot table with a selection of ores at different weights, which appear in the chest when placed.

image

https://streamable.com/isisiu


NillerMedDild:
> I'm not 100% sure of the actual loot table, might be better to limit potential items a bit more... Perhaps to only the interesting materials from the dimension, or giving those a much higher weight. Thoughts?

I think much higher weight is best, that way normal ores serve as filler 😄

NillerMedDild (approved):

MuteTiefling:
Alright, should be about 50% materials from the dimension, 50% filler from generic ores. 

Think a nether one would be suitable? Leaning towards an end one too that might be more of a botanical survey that gets some interesting plants and such.

NillerMedDild:
> Think a nether one would be suitable? Leaning towards an end one too that might be more of a botanical survey that gets some interesting plants and such.

Yeah that could be cool too, maybe it could output some of the BYG worldgen items 😄

MuteTiefling:
I think that covers all the various End and Nether flora :D

NillerMedDild (approved):

NillerMedDild:
Awesome! It's tempting to put this in normal mode too tbh.

MuteTiefling:
Certainly could. I was tempted to at least do the botanical surveys.... but honestly the rest wouldn't be out of place either

NillerMedDild:
Will you do the honours of moving it to base or should I? 😄

MuteTiefling:
nah, I'll go ahead and get it over there :D

PR #3303, opened 2021-09-23
[Expert] Thermal Augments

MuteTiefling:
Thermal Power
So it seems we can make our own augments for these, or alter existing ones, by slapping NBT on an item. So we could turn Powah Capacitors into thermal upgrades if we wanted to, or magical items like crystals or runes :D

As a test case, this gets a dynamo to max out at 4960 FE/t with a fuel efficiency of 73%.


NillerMedDild:
Very cool! ... we don't wanna alter the base ones though, right? :P

MuteTiefling:
Well, if we co-opt items from other mods or register our own, then we've got the problem of people not thinking to look for them.

If we use the thermal items and tweak them (which appears to override them fine) then folks will always know where to look to get stuff.

The big problem is future tweakability though. since it's all in NBT, we can't modify existing items. If someone ends up with a super augment and we want to remove it, well.. it's still there. It also doesn't show in JEI nicely unless we register a new item and explicitly show the item with NBT in JEI...

Not sure how best to handle this all, really.

NillerMedDild:
I feel like making dedicated KubeJS items with very similar names to the Thermal ones would be best, that way we can remove them if needed.

MuteTiefling:
Sounds good. What sort of output numbers would you like to see?

Base being 640 FE/t with resonant integral components and 3 Aux Reaction Chambers

Maybe 2-3 more tiers that work up to 4960? 200% augment, 500% augment and 1000% augment?

NillerMedDild:
> Maybe 2-3 more tiers that work up to 4960? 200% augment, 500% augment and 1000% augment?
>
> Maybe some Machine speed upgrades could be good too?

That sounds like a solid start, 3 tiers. 

I'm all for Machine speed upgrades too 😄

MuteTiefling:
No clue what's going on, but I'm having a hell of a time getting the Kube items with NBT to show in JEI. Gonna keep poking at this tonight to see what I can do though.

NillerMedDild:
Alrighty, could you perhaps commit what you have now and I can help spot the issue?

MuteTiefling:
done!

NillerMedDild:
Looks right to me tbh, I've never gotten the filled Morph-o-tool or Akashic Tome to show either, we might have a bug here.

MuteTiefling:
hrm, was hoping it would be as simple as your Tetra tools. Anyway, I've asked in latcord already, so whenever they get a chance to look at it.

MuteTiefling:
Ok, well, they're visible now at least!

image

MuteTiefling:
Also added in some fun catalysts:
Fire Rune for the Induction Smelter
image

Earth Rune for the Pulverizer
image

MuteTiefling:
So, some numbers. 3x MK4 speed gets you up to this:
image

2x speed and 1x efficiency gets you here:
image

Efficiency might be a bit high, but I was thinking of having these high tier efficiency upgrades require some tougher materials, like ether gas and maxwellian heat exchangers.

As for Machines, they get pretty fast with these upgrades, but also require the player to mix and match. You can't just toss 3 mk4 speed upgrades and expect them to work. Instead, you'll need to balance them out with something like a High-Flux RF Coil or efficiency upgrades to work well. In fact one of these with 2 efficiency upgrades seems to be the best combo. 

https://streamable.com/laq05u

Naturally this much speed would have to be very late game to craft. But with upgrades like these, Thermal machines could keep pace with mek machines for speed, while being pretty expensive to operate. 

There's some extra numbers in the comments of the code. The numbers on the upgrades start to look a bit silly (+729000% Base Power, for instance) but we can pretty well tune this however we want in terms of Processing time and Energy cost.

NillerMedDild:
Those are some insane augments, nice work 😄 Did you talk to Lat about hiding the non-NBT item?

BooBeerb:
I wonder if it's possible to do something similar with mek machinery. You can't add custom factory upgrades / speed upgrades, can you?

NillerMedDild:
Not afaik :)

Darkere:
You would need to copy the entire machine to make a second one with the upgrade that is faster.

MuteTiefling:
> Those are some insane augments, nice work 😄 Did you talk to Lat about hiding the non-NBT item?

Yeah. There was a bank and forth between him and max about possible someone they could add to kubejs, but right now it doesn't seem like it's possible.

MaxNeedsSnacks:
https://github.com/KubeJS-Mods/KubeJS/commit/ab950564080e61cb3a0a841f64333907c8baa9e8

:^)

PR #3333, opened 2021-09-29
KubeJS Powah Coolants Script

MagisterXero:
Added a script to allow the adding of solid coolants for Powah's Reactors
Please customize this as you'd like, these are really just reference points to use.


MagisterXero:
Also to note, this script is required on the Client side in order for players to see the coolant list, and not have some weird glitchy stuff when inserting/removing coolant items from reactors.

kubejs/config/common.properties
```diff
@@ -1,7 +1,8 @@
 #KubeJS Common Properties
-#Sat May 01 23:14:14 CEST 2021
+#Sat Sept 28 21:45:14 CEST 2021
 exportVisualizerData=false
 packmode=default
 hideServerScriptErrors=false
 announceReload=true
 serverOnly=false
+invertClassLoader=true
```
NillerMedDild, line 8: This setting change worries me to be honest 😄 
@TemperedFool did you create something that relied on this being false?

NillerMedDild (changes requested):

NillerMedDild:
Looks good to me, thanks for your contribution! :D I just gotta make sure the settings change doesn't interfere with anything else before I merge, sit tight 😄

kubejs/config/common.properties
NillerMedDild: I asked on the KubeJS Discord and am no longer worried 😄

PR #3335, opened 2021-09-29
[Expert] Blood Magic Updates

MuteTiefling:
Drop Occultism crusher outputs for expert to 1x, 2x, 3x, and 4x

More BM tooltips added and other helper text

Update recipes for new anointments. Made enhanced versions cheaper to promote use.
New Archer's Polish
image

Upgrades follow similar format of tau oil for the extended duration and saturated tau for the improved power
image

Removed new default blood shard recipe in expert

Vitreous Pickaxe brought forward a bit.
image


MuteTiefling:
Add clay from water sigil recipe to all modes:
image

Intermediate Cutting Fluid buff:
image

NillerMedDild (approved):

PR #3341, opened 2021-09-29
Make Morph o Tool work with Compact Machine Shrink Device

MuteTiefling:
Make Morph o Tool work with Compact Machine Shrink Device

Came up in #suggestions and seemed reasonable. Seems to work without issue.


LakeThomason:
Sounds awesome

Issue #3337, opened 2021-09-29
Two Issues: 'Meet Your Fight' Swampjaw Crash / I have no clue but Server Crash


greenking13:
### Modpack Version

E6-0.5.14

### Describe your issue.

So, the first boss from Meet Your Fight, Swampjaw seems to summon TNT to fight you while it is flying. This causes a ticking entity crash if you don't kill him fast enough. Also, I have no clue if it despawns if you de-render it, so it might still exist on my server. 🤷 
Crash Log: https://pastebin.com/0Gy4r5fV

Second one just occurred randomly, we thought it might be related to the first, but after closer inspection, 1. I was not anywhere near the chunk the previous crash occurred and 2. Computercraft seems to be mentioned. I don't believe anyone on my server has even attempted Computercraft and when this crash occurred, I was with only one other person who I know doesn't know how to use Computercraft.
Crash Log: https://pastebin.com/Vp6qCzvq

I've attached the latest log from this last crash, if any other details are needed, just ask.

-Green

### Latest Log

https://pastebin.com/32pQuwvd

### Have you modified the modpack?

Yes

### User Modifications

Only changes should be to the waystones configuration, which I'm pretty sure reverted when I did a pack update, so no.

### Did the issue happen in singleplayer or on a server?

Server

greenking13:
So, slight update, I just learned that villages can have computercraft in them. Huh.

Darkere:
Yes. Also your first crash looks more like an interaction between bees and a target dummy?

greenking13:
Dope, that fixes the second issue I'm guessing. Any ideas revolving around the first and how fighting Swampjaw correlates to bees and a target dummy? (The first of which could have been a wild, but the second is impossible without crafting I'm assuming.)

NillerMedDild:
@MaxNeedsSnacks it seems we have a RR crash here:

> So, the first boss from Meet Your Fight, Swampjaw seems to summon TNT to fight you while it is flying. This causes a ticking entity crash if you don't kill him fast enough. Also, I have no clue if it despawns if you de-render it, so it might still exist on my server. 🤷
Crash Log: https://pastebin.com/0Gy4r5fV

PR #3342, opened 2021-09-30
Disables computercraft village structures

theboo:
nobody likes these, and apparently it is advanced peripherals that adds them


NillerMedDild:
I've never seen one, do you have a screenshot or something for reference? Why are they not liked?

NillerMedDild:
Ah this fixes #3337 I see, nice

PR #3346, opened 2021-10-01
refreshes byg config, fixes missing biomes

theboo:
refreshes byg config... is the reason we were missing so many biomes, the .toml files had no change, only the .json


NillerMedDild:
Did you check whether this change causes chunk border issues?

theboo:
I do not have any old worlds to check this in

PR #3368, opened 2021-10-03
Add Blazing Blood as fuel for the Magmatic Dynamo

MuteTiefling:
Produces the same amount as diesel in the compression dynamo.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/magmatic.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/magmatic.js
@@ -1,0 +1,18 @@
+onEvent('recipes', (event) => {
+    var multiplier = 10;
+    var data = {
+        recipes: [
+            {
+                fluid: 'tconstruct:blazing_blood',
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/magmatic.js
```diff
@@ -0,0 +1,18 @@
...
+        recipes: [
+            {
+                fluid: 'tconstruct:blazing_blood',
+                energy: 1000000
+            }
+        ]
+    };
+    data.recipes.forEach((recipe) => {
+        event.recipes.thermal.magmatic_fuel({
+            type: 'thermal.magmatic_fuel',
```
NillerMedDild, line 13: does this work? seems like a bit of a mix between KubeJS Thermal and event.custom

MuteTiefling: Oh yeah good point. These are all a little wonky. I think they predate the thermal kubejs add-on and I just copied. I'll clean em all up here

PR #3388, opened 2021-10-06
Fixes missing valid apiary tags 

### LINKED ISSUE #3384: Missing valid_apiaries tag list.
### Modpack Version

Enigmatica6-0.5.14

### Describe your issue.

When placing a Tier 1 apiary, it does not register any blocks as valid to be used in an apiary. Looking further, when going into the config files for the modpack (Enigmatica 6 - E6 - 1.16.5\config\resourcefulbees\resources\data\resourcefulbees\tags\) there is no valid_apiary.json file to be found, which should be present based on the default installation of Resourceful Bees.

### Crash Report

_No response_

### Latest Log

_No response_

No

### User Modifications

_No response_

Singleplayer

theboo:
fixes #3384 

left the glassential blocks in as they are still not registered by RB's collision test

removed the `removeAll()` from the block and item tag scripts

tested in game and all seems to be working as it should


NillerMedDild (approved):

NillerMedDild:
Good work!

PR #3391, opened 2021-10-06
[expert] fix compressors dropping compressed bricks in normal mode

MuteTiefling:
[expert] fix compressors dropping compressed bricks in normal mode


MuteTiefling:
Oh, lol. Here I was thinking I just goofed it in the original pr. No worries :D

PR #3397, opened 2021-10-08
[Expert] Refined Storage Components

MuteTiefling:
Clean up some of the BM scripts.

Thoughts for RS components:

Destruction Core:
image

Construction Core:
image

Exporter:
image

Importer:
image

Cables:
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/extrastorage/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/extrastorage/shaped.js
@@ -68,7 +68,7 @@ onEvent('recipes', (event) => {
             key: {
                 B: '#forge:circuits/ultimate',
                 D: 'thermal:upgrade_augment_3',
-                E: ['extrastorage:iron_crafter', 'extrastorage:gold_crafter']
+                E: 'extrastorage:gold_crafter'
             },
             id: 'extrastorage:diamond_crafter_upgrade'
         }


MuteTiefling:
Raw Processors:
image

image

Looking to reduce our heavy dependence on iron... otherwise I might have stuck with something that fits the color better, like invar... dunno. 
image

NillerMedDild:
I think coins would be 👌

TemperedFool:
The process & materials are still annoying enough that they can probably be more generous? Maybe like 2x for the regular and 3x for the Enderium ones?

MuteTiefling:
Also, external storage

image

MuteTiefling:
Attempt at crafters:
Base
image

Iron:
image

Gold:
image

Diamond:
image

Netherite:
image



Note that these are not tiered crafts as it stands here, unlike default. Felt it was kinda unfair to have tiered circuits/integral components inside an already tiered craft. So each one is simply an upgrade from the base crafter, except netherite, of course.

theboo:
makes sense you wouldnt want to shove the double tier down their throats, but maybe there should still be some sort of tiered upgrade path in case they wanna upgrade their crafters? So maybe it could accept any crafter rather than just basic?

TemperedFool:
> I could see discounting the processors/iron during an upgrade, like you're just clumsily swapping out the control circuit and integral parts, destroying the old ones in the process. Similar to what we did for Powah in Normal.
> 
> As for mechanical crafting, I almost set these up that way... but so far we've kinda left that to one off type crafts, or things you're not going to be using a lot... and I kinda feel that's the best way to treat mechanical crafting because a fully smart/dynamic single-crafter-for-all-recipes is a major pain to set up. So most likely anything on repeat craft is going to end up with a very dedicated setup for that one recipe.

Yeah, totally makes sense; the only reason it even came to mind was that the upgrade path likely won't be done too frequently, and probably not as part of auto crafting. Easy enough to pop some hoppers on a few mechanical crafters to batch upgrade some. Keeping it simple makes sense though.

PR #3401, opened 2021-10-08
[Expert] Misc recipe changes

MuteTiefling:
Couple of minor changes. more thematic than anything.

IE Conveyors. Increases output from 8 to 12
image

Create Chain Drive:
image


MuteTiefling:
boy that had a lot more in it than I intended

PR #3432, opened 2021-10-13
update configswapper format

Darkere:
this requires configswapper 2.1 to work :)


Darkere:
They wouldn't do anything but sure is cleaner that way :)

PR #3472, opened 2021-10-17
[Expert] Integrated Dynamics continuation

MuteTiefling:
Logic Director:
image

Variable Store:
image

Logic Programmer
image

Output Variable Transformer
image

Input Variable Transformer
image

Menril Glass now uses Crystal Glass only
image

Same for Chorus Glass
image

Crystallized Chorus and Menril Blocks
image
image

Thermal recipes for menril/chorus glass and blocks now relegated to normal only. 

Proto Chorus:
image

Mono-Directional Connector
image

Crafting Interface:
image

Energy Interface
image

Item Interface:
image

Fluid Interface:
image


MuteTiefling:
I think this pretty well covers everything. The remaining items that could possibly use a tweak have at least had sub crafting components modified (like the logic director used in omni-directional connectors)

PR #3476, opened 2021-10-17
Blacklist powah from mekanism cardboard

Quezler:
Started to look into it when someone on my server tried to move a player transmitter with cardboard and lost it:

- the reactor just straight up vanishes when you cardboard one block of it (furnator, a single block, seems to be unscaved)

- since the transmitter also consists of 2 "actual" blocks it can get derpy as well (top and bottom can both be cardboarded)

- and energizing rods pop off without dropping if their cable is cardboarded (the cable itself does stay intact & reconnects)

i suppose this is enough to qualify for the entire mod being blacklisted for cardboard instead of just a few blocks 🤔


NillerMedDild:
Nice fix, thanks 👍

PR #3496, opened 2021-10-19
Suggesting a third magic feather recipe

Quezler:
For the two current magic feathers i have to go out of my own way in order to gather some of the missing parts.
Therefore i would like to suggest a new/third recipe to obtain them, which uses items from different mods:

Screen Shot 2021-10-19 at 15 53 38

(table is not in the same order as the picture, but in the order listed in the code)

| comparing | magic | tech | shiny |
| ----------- | ---- | ------ | ----- |
| has leaves  | yes | yes | yes |
| unique items | 6 | 6 | 6 | 6 |
| hardest | enderium | atomic alloy | refined radiance |
| overworld | roadrunner | blitz | aquamarine |
| mild effort | mana fiber | hazmat fabric | nebu ingot |
| goal | performing magic | setting up electric machines | multiblock contraptions |
| unsymetrical | top & bottom | top & bottom | left & right | 

honestly i cannot think of any good things to put in the table, so just ignore that, i'll write some text instead:

this recipe is not particularly hard, the most challenging item is probably refined radiance, i started the pr with that item in mind since it requires a beacon to make, and this item (without upgrading) will be needing a beacon to function at all.

i was planning to put some more items in there from the create mod, but i wanted to avoid using too much of it in order to give some of the other mods attention as well:

- nebu ingot from atum, will require a diamond for the portal & some exploration of another dimension
- two items from integrated dynamics, players will need to obtain a sapling for a tree, and produce slimeballs
- aquamarines from astral sorcery, these feel a bit out of place but i needed to fill the spot with something shiny

whilst most effort will go into the production chain for refined radiance it will nudge the players into exploring integrated dynamics & atum (and to a lesser extent astral sorcery), this doesn't feel quite done yet so i'll be leaving it as a draft (spent several hours thinking about it, would be a shame not post it at least somewhere)

feel free to comment any feedback/alterations, or just close it if it's too much out of place :)

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/normal/recipetypes/enigmatica/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/normal/recipetypes/enigmatica/shaped.js
@@ -32,16 +32,16 @@ onEvent('recipes', (event) => {
         },
         {
             output: 'magicfeather:magicfeather',
-            pattern: ['CEC', 'BDF', 'AEA'],
+            pattern: ['CFC', 'BDB', 'AEA'],
             key: {
-                A: 'atum:nebu_ingot',
-                B: 'integrateddynamics:variable_transformer_output',
-                C: 'create:refined_radiance',
-                D: 'byg:leaf_pile',
-                E: 'astralsorcery:aquamarine',
-                F: 'integrateddynamics:variable_transformer_input'
+                A: 'create:refined_radiance',
+                B: 'create:crafter_slot_cover',
+                C: 'create:electron_tube',
+                D: 'create:precision_mechanism',
+                E: 'create:chocolate_glazed_berries',
+                F: 'create:honeyed_apple'
             },
-            id: `${id_prefix}magicfeather_shiny`
+            id: `${id_prefix}magicfeather_create`
         },
         {
             output: 'losttrinkets:magical_feathers',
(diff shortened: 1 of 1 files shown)


RobuxShooters:
Maybe for the aquamarine, it can be replaced by its attuned version? Attuned aquamarines (I think that's what they are called, I forgot) are relatively late into AS however...

Sometimes, I feel like the magic path to craft the magical feathers is much more inexpensive compared to the tech path. The tech path requires cured rubber, which requires sulphur which can be somewhat difficult to acquire. It also requires drops from 2 different Alex's mobs as opposed to only one for the magic path. However, I reckon that refined obsidian ingots might be better suited to fit the ingot theme in the crafting recipes as opposed to ultimate alloys.

Perhaps the source fibre (is that the new name?) in the magic path could be exchanged for something tougher to acquire? Glyphs would be somewhat more difficult considering that source isn't as easy anymore as it used to be (placing volcanic accumulators over lava). The magic path could use another drop that comes from another animal in Alex's Mobs.

Quezler:
> > I have to go out of my own way in order to gather some of the missing parts
> 
> This was kind of the intent with these. :D That you'd have to do a little more than dig in the ground for a bit to gather the materials.

i didn't mean it like that: yes the feather requires some digging around, but not with the mods i wanted to play with, the feathers push you down a certain path wile the normal mode is supposed to be a kitchen sink, it would be fairer if there were some other combinations of items in different groups of mods so that you can unlock the feathers "at some point along the tech tree" instead of having to gather a few items of mods you have no intention of touching (yourself, e.g. splitting tasks/mods on multiplayer)

NillerMedDild:
Myeah maybe it's a bit too much, more difficulty in some form is what I'm after 😄

Quezler:
i'll try to come up with some revised recipes today/tomorrow 🤔 

some food for thought:
- _should_ each feather recipe use leaves?
- it came to my attention that ars has its own creative flight ritual, so would a feather recipe with ingredients from that mod make sense?

PR #3519, opened 2021-10-21
[Expert] Mekanism and other Tweaks

MuteTiefling:
PNC Speed upgrades, remove default recipe. Only crafted through the glycerol method:
image

RS Speed Upgrades get a similar treatment:
image

Stack upgrades:
image

Network Transmitter/Receiver
image

Dimensional Cells
image
image
image

Power Core Medium
image

Power Core High
image


MuteTiefling:
Charged Porter
image

Advanced Charged Porter: 
image

PR #3559, opened 2021-10-24
Mention the `quark:dragon_scale` doesn't drop from the first

Quezler:
The text is up for debate, but it would be nice if it was mentioned ingame 🤔


NillerMedDild:
`Only drops from Ender Dragons after the first one has been killed.` ?

NillerMedDild (approved):

PR #3599, opened 2021-10-28
[Expert] IE Energy tweaks

MuteTiefling:
| Machine                     | Baseline | Proposed |
|-----------------------------|:--------:|:--------:|
| Metal Press (plates)        |    40    |   1024   |
| Arc Furnace (per item x12)  |   2048   |   5120   |
| Excavator                   |   4096   |   10240  |
| Crusher                     |    240   |   6144   |
| Industrial Fermenter        |    640   |   1024   |
| Industrial Squeezer         |    640   |   1024   |
| Refinery                    |    80    |   2048   |
| Distillation Tower          |   2048   |   6144   |
| Sawmill                     |    120   |   3072   |
| Pumpjack                    |   1024   |   10240  |
| Blast Furnace Heater        |    32    |    256   |
| Assembler                   |    80    |   1024   |
| Diesel Generator Max Output |   4096   |   10240  |
| Cloche                      |     8    |   4096   |
| Auto Engineers Workbench    |    256   |   2048   |

So, the aim here is to really start ramping up power costs in the mid game, but also make these machines more worthwhile in terms of speed. Several of them have had speed values adjusted to be much faster, given their cost (crusher is flying).

The speeds on the squeezer/fermenter allow you to run a nice ratio of 2 refineries from 1 squeezer/fermenter pair. 
The diesel burn rate has increased up to 16384 (4x base) so one setup should keep a single generator running. Fuel density will go up to be closer to what we have for Thermal as currently a bucket is only worth 512k FE. I'd like that higher, I think. There's currently a ticket open over there to have fuel values be made into datapacks, so we'll see when/if that happens. 

Note: None of this actually works right now, sadly. Darkere is looking into that. IE is reading the config before configswapper sets them and re-writes what it had on world save, deleting whatever configswapper set...


MuteTiefling:
Note to self: Testing was done with Thermal's values for FE/bucket of diesel. 

When the diesel gen is updated to use datapacks for fuel, I'd like to bring it in line with the compression dynamos, allowing it to burn all the same fuels.

MuteTiefling:
https://github.com/BluSunrize/ImmersiveEngineering/issues/5062#issuecomment-955706016

djmattyg007:
Could I make the suggestion that these changes should be pushed back to Enigmatica 7? I don't think this kind of change is what people expect to find when upgrading from (for example) 6.0.5.21 -> 6.0.5.22. Some of these changes will completely break peoples' existing setups.

BooBeerb:
That is the most confusing sentence I have ever read. First of all, E7 will not exist. Second of all, you dont push changes from one modpack to an entirely different modpack. Second of all, this is for expert, not for regular.

NillerMedDild:
This is indeed for expert mode, so it shouldn't mess anything up, since that hasn't been "officially" released :)

waleed2244:
when is gonna be the officially released

MuteTiefling:
When it's ready.

YashTFEM:
> When it's ready.

Maybe as a Christmas or new year's gift or something?

NillerMedDild:
Ready to merge?

MuteTiefling:
Umm... gotta check the Immersive Petroleum thing. We still don't have the IE update so the diesel gen is toast for now, but I think we can go ahead with this once I check IP

config/serverconfigupdater-common.toml
```diff
@@ -10,7 +10,7 @@
...
+	filesToDelete = "defaultconfigs/immersiveengineering-server.toml,kubejs/assets/mythicbotany/lang/en_us.json,kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/occultism/spirit_fire.js,kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/tools.js,kubejs/data/tetra/loot_tables/actions/geode.json,kubejs/assets/enigmatica/tips/paraglider.json,kubejs/data/minecraft/loot_tables/chests/end_city_treasure.json,config/configswapper/expert/randompatches.txt,config/configswapper/normal/randompatches.txt,config/configswapper/normal/occultism.txt,config/configswapper/normal/immersiveengineering.txt,config/configswapper/normal/astralsorcery.txt,config/configswapper/expert/occultism.txt,config/configswapper/expert/immersiveengineering.txt,config/configswapper/expert/astralsorcery.txt"
```
Darkere, line 13: why this?

MuteTiefling: All of that's moved to the config swapper configs, so it's to avoid having the normal values reset the expert ones after an update

MuteTiefling: On second thought, I guess it's not actually needed

MuteTiefling: removed

PR #3621, opened 2021-11-01
[Expert] Mekanism, Refined Storage, and RFTools Quarry Energy Adjustments

MuteTiefling:
| Assuming   10x speed/energy |             |
|:---------------------------:|:-----------:|
| Machine                     | Rate (FE/t) |
| Digital Miner               | 80,000      |
| Digital Miner (Silk Touch)  | 480,000     |
| Sawmill                     | 640         |
| Smelter                     | 640         |
| Crusher                     | 640         |
| Enrichment Chamber          | 640         |
| Metallurgic Infuser         | 5,120       |
| Modification Station        | 5,120       |
| Pump/Plenisher              | 200         |
| Purification Chamber (3x)   | 2,560       |
| Chemical Injection (4x)     | 5,120       |
| Dissolution Chamber (5x)    | 10,240      |
| Laser (Max Usage)           | 204,800     |

Alright, so in an attempt to create some more power progression as people get into tier 3 tech, I've increased a good portion of mek machine costs.

Things I attempted to avoid: Anything used heavily for energy production. For example, isotopic centrifuges are mostly only for the Fission process. Increasing their cost only hurts the energy return from fission. So it was left alone. 

Ore processing machines see a heavy increase in energy required to run. The Digital Miner is similarly very expensive to run. 

Things like the sawmill and furnace remain cheap, but see an increase. These numbers mostly keep them in line with Thermal, but Mek gains an advantage in space.

For example, a single fully upgraded Thermal furnace runs about 320 fe/t and smelts about 2x as fast as Mekanism. The cost to run a fully upgraded Mekanism furnace is 640 fe/t. This may seem high at first, but consider that adding a single factory installer allows it to process 3x more items per tick. This is now faster than Thermal in a single block space, though the cost is also tripled. So you conserve space, but trade that off for more energy usage. 

The increase to the Laser cost is more of a QOL change. By increasing this, a single laser can pass more energy into an amplifier faster, making it easier to trigger a Fusion reactor once that point has been reached. 

Lastly, the pump, plenisher, and assemblicator had their costs reduced as there's nothing much special about them. For the pump, this really just serves to remove a bit of the cost from Fusion fuel production.


MuteTiefling:
RFTools Builder
Increase cost by roughly 3x
Rates for Fortune and Silk Touch similarly adjusted. Silk touch is 6x the base rate, similar to Digital Miner

Range reduced from 512x512 to 128x128

Increased power it can hold/receive per tick to accommodate increased operational cost.

MuteTiefling:
Significant increases to RS power usage across the board, including adding per cable cost, controller cost, and disk drive cost.

Max buffer increased to accommodate

MuteTiefling:
Serverconfigs are not currently being updated in a timely manner configswapper 2.2

NillerMedDild:
> Serverconfigs are not currently being updated in a timely manner with configswapper 2.2

Well that sucks :/

Darkere:
> Serverconfigs are not currently being updated in a timely manner with configswapper 2.2

What do you mean by this?

NillerMedDild:
I think the RS Controller increase is too much, perhaps 512 would do? The rest looks great to me 👍

NillerMedDild (approved):

MuteTiefling:
> > Serverconfigs are not currently being updated in a timely manner with configswapper 2.2
> 
> What do you mean by this?

Same as was happening with the IE Configs the other day. When I load into the game, I see the values are changed in the configs, but it's like they changed after the mod read them. They then revert back when closing the game.

MuteTiefling:
Alright, dropped the controller to 256

Also sneaking in Powah reactor buffs here. 

Aproximate FE/t for the various tiers. 

| Powah Reactor   (Basic)  | 16242.1  |
|--------------------------|----------|
| Powah Reactor (Hardened) | 29110.5  |
| Powah Reactor (Blazing)  | 54801.1  |
| Powah Reactor (Niotic)   | 108806.7 |
| Powah Reactor (Spirited) | 228873.9 |
| Powah Reactor (Nitro)    | 512533.4 |

PR #3636, opened 2021-11-02
Add Tinkers Construct Quest Line and Loot Tables

mcilhargey:
Add a Tinkers quest line to educate/reward building in the tinkers line. This quest line has it's own loot tables based on items in the mod itself.


NillerMedDild:
Looks good from what I've read so far 👍 
I would however like the slime island quests to not be checkbox based, perhaps they could be merged to be one per dimension, and just require the saplings and such?

theboo:
got home late from work today, ill look through these tomorrow

theboo (commented):
no description in "puny smelting" task, as i dont know tinkers 1.16 well at all, unsure if this is intended
image

checkmark quest for tinkers upgrade, is there an advancement for your first applied upgrade that might be able to be used instead?
image
same question here:
image

If one of the books has info on what the various frames' effects are, you might mention that here?
image

KnightMiner:
> checkmark quest for tinkers upgrade, is there an advancement for your first applied upgrade that might be able to be used instead?

You an easily create an advancement that checks for a tinker tool having one of a list of modifiers. The syntax is rather verbose, but look into the TiC "ability master" advancement or the "upgrade slots" advancement.

MuteTiefling:
If they're advancements, we can actually just watch for those.

KnightMiner:
The advancements don't do exactly what you want the ones I mentioned are for applying all modifiers from a particular set, not applying just 1

theboo (commented):
looks good to me!!

NillerMedDild:
Woohoo \o/

PR #3678, opened 2021-11-07
[Expert] Thermal expertification

MuteTiefling:
Tinker Workbench
image

Vacuumulator
image

Igneous Extruder. Pushing this back a ways so as to promote more creative stone farms earlier on. 
image

Aqueous Accumulator. Similarly pushed slightly back. Hopefully not too far on this. 
image

Nullifier
image

Similarly changed trashcans to be in a similar cost range:
image

Thermal Saw Blade:
image

Sawmill
image

Fluxsaw
image

Fluxbore
image

Noticed IE press wanted Thermal gear press for gears but couldn't accept it. Swapped to IE mold. 

Induction Smelter
image

Environmental Kiln:
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/immersivepetroleum/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/immersivepetroleum/shaped.js
@@ -1,0 +1,36 @@
+onEvent('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
+    const id_prefix = 'enigmatica:expert/immersiveengineering/';
+    /*
+        ,
+        {
+            output: '',
+            pattern: ['', '', ''],
+            key: {
+                A: ''
+            },
+            id: ''
+        }
+    */
+
+    const recipes = [
+        {
+            output: 'immersivepetroleum:flarestack',
+            pattern: ['ABA', 'CDC', 'AEA'],
+            key: {
+                A: '#forge:plates/steel',
+                B: 'tconstruct:flint_and_bronze',
+                C: 'thermal:redstone_servo',
+                D: 'immersiveengineering:fluid_placer',
+                E: 'immersiveengineering:fluid_pipe'
+            },
+            id: 'immersivepetroleum:flarestack'
+        }
+    ];
+
+    recipes.forEach((recipe) => {
+        event.shaped(recipe.output, recipe.pattern, recipe.key).id(recipe.id);
+    });
+});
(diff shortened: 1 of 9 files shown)


MuteTiefling:
Pulverizer
image

Magma Crucible
image

Fluid Encapsulator
image

Blast Chiller
image
Intentionally pushed far back. There are other less obvious ways to farm ice early on if needed. Would rather have people explore those. One being suggested in the recipe itself.

MuteTiefling:
Centrifugal Separator
image

Pyrolyzer. Hopefully not breaking anything with this one. Already required coal coke, so I just expanded that up to blast bricks.
image

Fractionating Still
image

MuteTiefling:
Multiservo Press
image

MuteTiefling:
Insolator pushed back a fair bit
image

Sequential Fabricator as well:
image

Cleaned up recipe input replacement since further changes were made to each of those recipes above.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/immersivepetroleum/shaped.js
```diff
@@ -0,0 +1,36 @@
+onEvent('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
+    const id_prefix = 'enigmatica:expert/immersiveengineering/';
```
NillerMedDild, line 5: Typo, immersiveengineering -> immersivepetroleum I think 🙂

MuteTiefling: Ah good catch

MuteTiefling:
Dynamos!

Stirling
image

Magmatic with light progression into PNC
image

Compression needs a bit more progress into PNC
image

Magical alternatives:
Numismatic requiring a bit of early alchemy

Lapidary with a touch more magical progress
image

MuteTiefling:
Redstone Furnace
image

MuteTiefling:
Think that's about it other than upgrades... which I really need to do more testing to figure out where to put them. Might not move them at all, however, considering the Integral Components are the main gate there.

NillerMedDild:
I think the Stirling Dynamo is a bit on the complex side considering it's the first generator one can make.

Other than that, really nice recipes! I am a bit nervous they're too expensive/micro-crafty, but we'll have to gauge that in playtesting.

MuteTiefling:
Hrm, drop the andesite alloy from stirling maybe?

NillerMedDild:
Yeah that ought to do the trick :)

MuteTiefling:
Alright, simplified it back to more vanilla materials:
image

NillerMedDild (approved):

NillerMedDild:
Nice! Thank you :D

PR #3720, opened 2021-11-13
[Expert] Storage Drawers

MuteTiefling:
Starting off with a tweak to Mek Auto crafter. This give a semi early auto crafting option that's easily paired with Pretty Pipes to help with pre RS automation. I went with this over the Thermal one for early game as it is slower and upgrades are very far away. So it's just a slow little crafter for early game wheras Thermal or RFtools become available later and are faster. 

Crafter
image

Recipe patterns
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/enigmatica/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/enigmatica/shaped.js
@@ -493,8 +493,6 @@ onEvent('recipes', (event) => {
         event.shaped(recipe.output, recipe.pattern, recipe.key).id(recipe.id);
     });
 
-    // Unification type scripts. Consider moving out?
-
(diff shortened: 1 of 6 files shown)


MuteTiefling:
Upgrades
image

image

image

image

image

Framed Drawers:
image
image
image


Regular Drawers
image
image
image

Upgrade Templates
image

Void
image

Compacting
image

Controller
image

Slave
image

NillerMedDild:
Very nice recipes so far 👌

MuteTiefling:
Cool thanks. Thought I was going to have more for this. Only other thing would be possibly adjusting the total capacity down. But I think they're probably fine where they are with the new cost.

NillerMedDild:
> Cool thanks. Thought I was going to have more for this. Only other thing would be possibly adjusting the total capacity down. But I think they're probably fine where they are with the new cost.

Agreed 👍

PR #3736, opened 2021-11-16
[Expert] IF Progression and IE Part Two

MuteTiefling:
Unrelated: Switch Wax in for comb. More appropriate to the intended theme.
image

Disable IF Mob Slaughter Factory

Pathway into IF through Astral. This is intended as a one time craft with a more normal tech based craft available afterwards. That recipe is not yet determined but will still use an Advanced Machine Frame (and therefore refined pink slime).
image


MuteTiefling:
Simple Machine Frame:
image

Advanced Frame:
image

Supreme Frame:
image

MuteTiefling:
Pull the Chemical Washer forward a bit to put mek 5x and IF 5x on about even footing:
image

MuteTiefling:
Laser Drill
image

Laser itself brought forward to be obtainable here
image

Fluid Laser Base
image
Using Advanced frame, accessible only after the Astral recipe completes and some pink slime obtained

Ore Laser Base
image
Gated a bit later behind the Supreme Machine frame, requiring deeper magic progress (iesnium)

MuteTiefling:
This gets progression moving again. Will continue to plan the inevitable machine shuffle in another PR. Have a rough plan for moving some of the machines around to be accessible earlier or later.

MuteTiefling:
Decided to stick the basic IE progression in here as well.  

Currently looking at 2 tiers. Anything unnamed would simply fall into more standard gating with IE part 1 and recipes.
First tier would maybe open up with with PNC Circuits and include the following:

- pumpjack for better access to oil
- excavator primarily for access to some of the custom nodes like corundum and nebu. IF ore lasers being a bit later. On the fence on this one, really. Would like it to be earlier than ore lasers though.
- Mixer and sawmill just because. Honestly not sure to what extent these will even be useful... Sawmill possibly needs something special added and not just speed.
- Squeezer to prepare for Mek batteries using graphite
- Lightning rod as a potential mid game power source
- metal press as an upgrade to Thermal 

Second Tier would open up a bit later into IF progression and includes the following:
- Auto Workbench (allowing automation of key ingredients, such as perfect electrodes)
- Arc Furnace
- Assembler
- Crusher (plans to buff as discussed)
- Diesel Generator (also big buff planned)
- Refinery/fermenter to fill out requirements for diesel production
- coker unit / distillation tower / and hydro treater (not quite sure what to do with these yet)

NillerMedDild:
Good plans so far 👌 I agree with you about moving Excavator to be available sooner, I forgot about the custom veins, those should help with the grind.

MuteTiefling:
Blueprints
image

image

NillerMedDild:
> Probably need better art for the outputs so they look different...
Combine the above to create a final Engineer's School Project. This roughly represents materials for a 50x50 building for 20 students. It is... a bit costly? My hope is that it's not actually too bad though. Like you need 2500 copper shingles, but you get 4 per copper plate. So it's a bit less than 10 stacks of copper.

Totally an acceptable cost for this stage of the game :)

MuteTiefling:
Heavy Machinery:
Computer Package:
image

Fluid Drill Package
image

School Upgrade Package
image

Heavy Schematics:
image

MuteTiefling:
Alright. Think that wraps this for real.

NillerMedDild (approved):

NillerMedDild (approved):

PR #3738, opened 2021-11-16
Fix Mythic Botany Alfsteel Axe

GalaxySynth:


NillerMedDild:
Thanks for the fix, but would you mind targetting the develop branch instead? :)

GalaxySynth:
whoops, my bad! will do

PR #3742, opened 2021-11-18
[Expert] IE Crusher buff

MuteTiefling:
IE Crusher outputs Create Crushed Ore instead of dust. Results in about 4.7x ore processing if washed. 
image


MuteTiefling:
Update Occultism Crusher to use Crushed Ore too

image

Expected rates with washing:
Foliot Crusher = 1.39x
Djinni Crusher = 2.79x
Afrit Crusher = 4.17x
Marid Crusher = 5.56x

Crushing Speed increased as well.

MuteTiefling:
After further thought, I'm not so happy with the occultism crusher here. I like the idea of adding more processing to it, but I think the afrit crusher rate is probably as high as it should go... Might have to drop the multiplier again.

NillerMedDild (approved):

PR #3769, opened 2021-11-23
[Expert] Industrial Foregoing Power Pass

MuteTiefling:
Generally increases the cost to operate all IF machines. Lower end machines are kept cheap to run, while the bulk of the boost ends up in things like enchanting, ore processing, and the ore/fluid laser.

Ore Processing: 

The below ratios assume maximum upgrades in each machine and demonstrate what's needed to keep all of the machines running at the same rate. 

2x
1 Washing Machine
1 Fermentation Station running 1/4 the time
1 Sieving Machine running 1/4 the time

Approx power: 600 FE/t

3x
1 Washing Machine
1 Fermentation Station running 1/2 the time
1 Sieving Machine running 1/2 the time

Approx power: 800 FE/t

4x
1 Washing Machine
3 Fermentation Station 
3 Sieving Machine

Approx power: 2800 FE/t

4x
1 Washing Machine
5 Fermentation Station 
5 Sieving Machine

Approx power: 4400 FE/t


MuteTiefling:
Bio Fuel and Bioreactor

Bumps biofuel up to the same rate as other bio fuels for Thermal Dynamos.

Bioreactor takes 2 million FE over 7200 ticks to make fuel. So about 278 FE/t before upgrades. 

Biofuel generators run at 1000FE/t, producing 10 million FE per bucket.

This means the setup can be net negative if the bioreactor is run at low efficiency (only providing one type of input).
However, with full upgrades (speed, efficiency, and processing), a single bioreactor can run 9 generators at a net gain of 15.28 million FE at a total of 9k fe/t.

NillerMedDild:
> 2x 1 Washing Machine 1 Fermentation Station running half the time 1 Sieving Machine running half the time
> 
> Approx power: 40k FE/t
> 
> 3x 1 Washing Machine 1 Fermentation Station 1 Sieving Machine
> 
> Approx power: 60k FE/t
> 
> 4x 1 Washing Machine 3 Fermentation Station 3 Sieving Machine
> 
> Approx power: 140k FE/t
> 
> 4x 1 Washing Machine 5 Fermentation Station 5 Sieving Machine
> 
> Approx power: 220k FE/t

I feel like this is very expensive considering how little the player can generate with IF at this point, without scaling the number of generators up massively.

MuteTiefling:
Good point. I can certainly lower the cost of the 2x and 3x processing, as those would realistically be the only modes they'd be able to access early on. Once they're getting into pink slime/ether gas needed to run 4x and 5x though, they're similarly getting into Powah and the more expensive machines of late game.

NillerMedDild:
Sounds good :)

MuteTiefling:
Thinking about this more. I can't actually reduce the 2x/3x power per tick without similarly impacting the higher tiers. So I'm just gonna bring it all down a fair bit and focus on making the laser drill more expensive to run. That'll be the main cost of moving into 4x/5x here.

MuteTiefling:
new values above

NillerMedDild:
Looks good 👍 perhaps even a little too cheap 🤔 would 2x the current be acceptable?

MuteTiefling:
I'm good with doubling it

MuteTiefling:
Ore/Fluid Drilling

Laser Drills
Base cost: 5k FE/t
Slowed down moderately compared to default as they're very fast and can be stacked a lot. Requires 100 ticks of progress rather than 50.

Stasis Chamber
Base Cost: 25k FE/t

A spirited Powah reactor, for reference, outputs about 225k FE/t. So while costly, this should be very attainable at this level. A setup with a single stasis chamber and 4 drills would run at a cost of about 45k FE/t. 

A spirited thermo gen similarly outputs about 86k fe/t running on ether gas. A laser setup could therefore be self powered, and does still produce a net gain on ether gas even without any upgrades in the laser. Not a lot, but some upgrades in the lasers go a long ways to making that very good. 

Ore Laser Base
Further slowed from 100 ticks of progress per operation to 200. Makes Ore mining more costly without further impact to liquid mining.

MuteTiefling:
IF Dissolution chamber runs at 25k FE/t now. 

Enchantment factory at 50k FE/t

Applicator and Extractor at 5k FE/t

Mob Crusher runs at 50k FE/t

Mob Duplicator runs at 2.5 million per operation, which amounts to 50k FE/t without upgrades. 

Similarly, the Wither Builder runs at 500k per operation, effectively 5k FE/t without upgrades.

PR #3778, opened 2021-11-25
Frosty Json

Master2Mind88:
Here is frosty json file updated that is updated so that it has Centrifuge Data and use the secondary output as what the Thermal Pulverizer will give you if that is correct.


TemperedFool:
Thank you - not sure what I did to not include that...

PR #3788, opened 2021-11-26
Apotheosis Tome Towers and Astral Shrines loot tweaks

MuteTiefling:
Introduces a bunch of 'magical' type reagents as extra loot in Apotheosis tome towers. The original loot table rolls first, then the new ones as extra, so this shouldn't take anything away or make them rarer.

Still want to add some more caster tomes to this, only have 3 so far. Just need ideas for spells. 

Similarly, I've added a 'breeding pair' of animals to the Astral Shrines. Each chest has a 50% chance to have 1 pair of Shrink bottles with animals in them.


MuteTiefling:
Gonna call this good for now. Can always come back and add more books

NillerMedDild (approved):

PR #3807, opened 2021-11-29
Move Alloy Bees to Induction Smelter

TemperedFool:
Moving from block->egg mutations to Induction Smelting for Alloy Bees. Originally wanted to do it this way, works now apparently. This resolves a progression break in Expert for Steel and other alloys.


NillerMedDild (approved):

TemperedFool:
Awesome, I want to just make sure I failed to mess any of this up, then I'll merge it.

TemperedFool:
Remembering why I didn't do this last time - Thermal doesn't seem to care about NBT on inputs. I've swapped all these recipes over to Honeycomb Blocks, and when possible an empty jar too. Lumium/Signalum/Enderium all have 3 inputs, so they get no jar.

PR #3821, opened 2021-12-01
[Expert] Pyrolizer back behind coke dust

MuteTiefling:
This was originally behind coke dust as an upgrade to the IE coke oven, but that seems to have been lost along the way. Just puts it back there by using blast brick in the recipe. 

image


NillerMedDild:
I'm gonna playtest this soon to see how the Create-detour feels like when you're going Thermal first, hopefully this is good though 😄

PR #3823, opened 2021-12-01
[Expert] Smol Tweaks

MuteTiefling:
Cake transmutation
image

Don't think Bucket of Slime is gonna work out (too much confusion since tinkers green slimes cannot currently be bucketed) So, a slight change to the scanner
image

Mud unification for all modes
Byg mud was conflicting with environmentals mud. Unified to environmental and made stonecutter recipes for converting mud and mud bricks.


MuteTiefling:
ready for review

NillerMedDild (approved):

NillerMedDild:
Looks good :D

PR #3826, opened 2021-12-01
Little Buzzy Tweaks

TemperedFool:
Fixes up loot tables for RBees quests, removing all the extra NBT and making sure gifted bees don't break progression. Almost all of the bees given are world-gen, and they come in pairs. Tables have 2 pools, one for a bee pair, and one for other useful bee-related stuff.

Bump up spawning bees honey/wax output - both can be a bit hard to come by, and are required for progressing in RBees.

Cobalt secondary makes it strictly better than Cobalt bee; blizz powder isn't as good as Blizz bee though.


NillerMedDild (approved):

NillerMedDild:
Nice work!

PR #3838, opened 2021-12-03
[Expert] Fix uncraftable hardened jetpack

MuteTiefling:


NeosGhost:
Modpack Version Expert 0.5.24

Hardened jetpack still uncraftable

Rodg88:
It is craftable, the bucket needs to be named the correct name. You can ether find a #MLG-YOLO bucket as loot or make one yourself by naming it when you put an infinity book onto a water bucket.

PR #3845, opened 2021-12-05
Documentation and tweaks

MuteTiefling:
Simplify Arcane Gold
image

More Info tabs:
image

image
Removed the easier vacuum tube as a result of this. 

image


MuteTiefling:
Lowered number of rolls for loot in atum crates. Left nuggets, but at much lower rates than before. The idea is that these guys are very successfully pillaging surrounding areas and, as such, are rather rich in raw materials.

PR #3856, opened 2021-12-06
[Expert] IE recipes, configs, and more

MuteTiefling:
Due to progression changes for some modes, I'm removing the 'gating' in the patchouli books. This enables the players to see all entries from the start. While this might normally cause some confusion as to how to progress, this being expert and the fact we have quests, I think this is acceptable to prevent entries being locked out unintentionally. 

Engineering Blocks
image
image
image

LV Capacitors
image
MV Capacitors
image
HV Capacitors
image

Ranged Pumps pump
image
Honestly people would be a bit silly to use this for lava over Create, but hey :D

Storage Scanner
image
Brought forward a bit by including some magic. 

IE windmill/waterwheel brought forward
image
image
image

Dynamo
image

image

Radiator Block
image

Portable Generator
image

Generator Block:
image

Cloche:
image

New way to get Fish Bones, rather than fishing for trash, just get any fish and convert to bones with NA
image


MuteTiefling:
Removing Compressors and Charging Stations from the PNC Villager homes. This means two things:
1) No early trades for PNC stuff. Once you're able to make a charging station, trades are available
2) No access to Compressed air prior to making real compressors. Despite breaking when attempting to pick them up, they could still be used in situ, which wasn't ideal.

TemperedFool:
can we plz get this in soon? I want the storage scanner ^_^

MuteTiefling:
Should be set for review.

NillerMedDild (approved):

Issue #3849, opened 2021-12-05
Impossible to make Enderium tools due to different liquids


wowdud:
### Modpack Version

0.5.21

### Describe your issue.

Melting down Enderium in any way, or alloying it in any way produces Molten Enderium from Tinker's Construct. However, this version of Molten Enderium can't be used to make tools; only the Emendatus Engimatica version can be used for that. That version of Molten Enderium doesn't have a recipe, though.

### Crash Report

N/A

### Latest Log

N/A

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Singleplayer

NillerMedDild:
Thank you for the report, I will investigate the issue.

MuteTiefling:
found the issue here, fix incoming

PR #3860, opened 2021-12-06
Fix uncraftable Tinkers parts using Thermal materials. Clean up unused materials.

MuteTiefling:
Resolves https://github.com/NillerMedDild/Enigmatica6/issues/3849

Also re-adds thermal hardened glass making in expert via tinkers and cleans up some unused materials from EE (ruby/sapphire/thermal molten metals)


kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/tconstruct/materials.js
```diff
@@ -1241,7 +1241,7 @@ onEvent('server.datapack.low_priority', (event) => {
                 {
                     type: 'tconstruct:material_fluid',
                     fluid: {
-                        name: 'emendatusenigmatica:molten_enderium',
+                        name: 'tconstruct:molten_enderium',
```
KnightMiner, line 1244: If you use a tag here instead of a fluid, it would work with both Tinkers and Emendatus Enigmatica fluids. Most of the native casting/alloy recipes use the forge tag.

PR #3874, opened 2021-12-09
[Expert] Tinkers and Mek armor update

MuteTiefling:
Tinkers Travelers Gear
image
image
image
image

Plate armor moved to sequenced assembly
image


MuteTiefling:
Meka Suit moved forward a bit.
image
image

No longer needs Fission to craft. However, the power use has been tweaked accordingly. All forms of blocked damage are now considerably more expensive. Meaning the armor can be used early for it's utility without making the player unkillable. 

The limiting factor here is mostly charge rate. Until FluxNetworks becomes available, remotely charging the armor fast enough to offset heavy incoming damage will be very difficult. Tested against 2-3 wither skeletons and witches, they were able to kill me pretty quickly if I just let them beat down my shields. 

The idea for the upgrades will be to move a few of the more utilitarian ones forward, like the jetpack and upcoming elytra. Jump height, run speed, night vision, etc. More powerful upgrades like energy (increases shield duration and charge rate), creative flight, and others will remain gated behind fission/antimatter where appropriate.

YashTFEM:
I don't see the point of this armor then. If it can't defend against just a few wither skeleton shots, then what's the use? Other than utility.

MuteTiefling:
You upgrade the shields and you get pretty unkillable very quickly. And I did not say you die in a few hits. It takes them a bit to chew through your shields still, but you _will die_ if you just stand there and take it. Much like you will with any armor set.

MuteTiefling:
Meka Suit upgrades
Solar
image

Inhalation Purification
image

Nutritional Injection
image

Energy Unit. Fairly expensive and locked behind Fission
image

Radiation Shielding
image

Vision Enhancement
image

Electrolytic Breathing
image

MuteTiefling:
MekaTool


Silk Touch
image

Farming:
image

Excavating
image

Teleportation
image

Vein Mining
image


Attack Amplification
image

MuteTiefling:
Removed pink slime alloy and refined obsidian alloying recipes from tinkers

MuteTiefling:
That... should just about cover this. Moves meka tool forward a bit too, somewhat in line with other tech tools. This, unfortunately pretty much guts all the uses for polonium >.> Gonna have to fix that in end game recipes

PR #3881, opened 2021-12-11
Update blasting.js

GalaxySynth:
Fixes missing blasting recipes, instead they are doubled up in smelting currently


NillerMedDild:
Nice catch ty!

PR #3884, opened 2021-12-11
Fix freeze between RS and polymorph

Darkere:
Seen it quite a few times now and just happened on the dev server.

No idea how or why tbh. The chunk with the RS system was already loaded in this case.
Spark report in case we want to report this at some point: https://spark.lucko.me/KHuSv2JuXzhttps://spark.lucko.me/KHuSv2JuXz


NillerMedDild:
Aw too bad, but thanks for the fix.

PR #3898, opened 2021-12-14
[Expert] Tier 3 magic quests and progression tweaks

MuteTiefling:
Soul Scribe
image

Blood of Kvasir
image

Attunement Altar
image

Ethereal Slate
image

Iridescent Altar
image

Mead of Kvasir
image


YashTFEM:
Why is magic crossing over into tech? Should magic stay in magic progression, and tech stay in tech progression?

YashTFEM:
Namely, why does the mead of Kvasir need the thermopneumatic press?

MuteTiefling:
Removing the ability to manually mine livingrock and other abundant 'free' resources in Alfheim. These should, _hopefully_ continue to work as intended in other dimensions. More testing needed. Currently set to drop some random junk instead. Ideally, would like to make a custom 'illusory' block as a the real drop here.

NillerMedDild:
Is it possible to start a Gaia fight with an Occultism pentacle in it?
Also, depending on how late Alfheim is, I think it could be a nice reward to be able to mine living rock and such freely, or would it be too easy / bypass progression?

MuteTiefling:
You have to put the pentacle under the gaia ritual for this one to work, but it does totally work. Have tested that.

As for the living rock etc... I mean, yes we're pretty late in progression at this point. It just feels so... Odd to go mine it up after all this time of crafting it. Dunno. I've been thinking about changing the loot tables for a bit now since we couldn't modify the dimension. Dually just decided to give it a try.

I'm still a bit torn honestly, so I'll defer to your judgment here, but I was thinking that maybe the illusory 'junk' could actually be given a use as we get into end game recipes.

MuteTiefling:
Magic quests complete. Ready for review.

NillerMedDild:
Let's go with changed loot tables for now then, see how it plays 👍

NillerMedDild (approved):

PR #3907, opened 2021-12-15
[Expert] Progression fixes

MuteTiefling:
Occultism bowls moved forward. Were too late where they were to be useful
image

Blank runes now de-coupled from Runic Altar, allowing building up to a simple tier 3 blood altar before getting into botania mana generation
image

Ease up restrictions on Low Tier Occultism Stabilizers
Tier 1
image

Tier 2
image

Tier 3
image

Tier 4
image

Major refactor of ritual script since my brain finally started understanding maps better.

Renewable Virulent Mix
image


MuteTiefling:
Cheaper Solegnolia. 
image

MuteTiefling:
Occultism configs have changed a lot and our default is out of date. 

Moved full occultism server config to config swapper. 
Config now has a global multiplier for the time of rituals. Using that now instead of re-creating every single recipe just to reduce the time. So a lot of recipes in Base/Normal are just no longer needed. Updated accordlingly. 

Similar for Expert, Base recipe times were increased to match the decrease in configs.

MuteTiefling:
Ook.
image

NillerMedDild (approved):

PR #3932, opened 2021-12-19
Fix pneumaticraft expert quests dependencies

awebneck:
Currently, the expert gates pneumaticraft quest depends on RFTools having been completed - but this is reversed: the RFTools Machine frame requires advanced Pneumaticraft machines.


Darkere:
I'm pretty sure this is fixed in dev.

MuteTiefling:
It has indeed been fixed in Dev already, but thank you just the same. :D 

Also, if you're looking to contribute in the future, be sure to target the dev branch rather than master. <3

awebneck:
@MuteTiefling @Darkere D'oh, will do. Thanks!

PR #3936, opened 2021-12-21
[Expert] Update Engineering Student Supplies

MuteTiefling:
Was just using tools instead of inserting them.
image
Fixed typo in name too

New Tinker's Workbench, same reason, was using tools:
image


MuteTiefling:
Was going to include IE fuel in this, but I'm having trouble getting it to work, unfortunately.

https://github.com/BluSunrize/ImmersiveEngineering/issues/5128

NillerMedDild (approved):

PR #3937, opened 2021-12-21
Fix players not getting waystones

landretk:
I wondered why I didn't get a waystone and neither do many players in my server. The math in this function gives an array index in the range of [1,3]. 3 is out of bounds so returns nothing. Corrected to be in the range [0,2] so it's within the array bounds.

random() -> [0,1) -> *3 -> [0,3) -> floor() -> [0,2]

Double checked and this gives a uniform distribution too.

Before:
image

After:
image


NillerMedDild:
D'oh thank you ^^'

PR #3943, opened 2021-12-23
atum tweaks

Rodg88:
no fog, less patrols, and kangaroos

not sure if you'd want to disable the fog entirely, as it kind of makes it 'easier'. But the fog is pretty horrible...


Rodg88:
an example of fogless atum topside
image

RobuxShooters:
I'm not sure how much of an effect raising 1000 to 1600 provides on the patrol spawn rates but I agree that the spawn rates should be decreased by quite a lot. This is what happened in my world when I AFK'd in Atum for around an hour:
image
Because of this, it was quite laggy when I returned after AFKing in the dimension. Ideally there should be a mob cap stopping the patrols from spawning when the number of mobs goes over a certain value, but it does not seem like it.

MuteTiefling:
Hrm, lack of cap is worrying. There are spawners out in the open that could allow endless spawning as mobs move away from it. It's there any chance you were near one?

RobuxShooters:
I believe I was AFKing in a ruined structure but I broke the spawner that was in it beforehand (it was spawning Bonestorms). I do have another image that shows the situation after I AFK'd in the area (keep in mind this is only in one general direction):

Luckily I had the MekaSuit by that time so the arrows were more of an annoyance than threatening but I can see that it will be quite unfair for players due to the endlessly-spawning waves of patrols, even more so after AFKing in the dimension.

NillerMedDild:
I think we need to ask the Atum devs whether a cap exists, and decrease patrol spawn rate. 

I don't want to disable the fog, it's a big part of the atmosphere, albeit annoying :P

Rodg88:
just closing this as looks like it might need work on the mod's side instead

NillerMedDild:
Alrighty, thanks anyway!

sscards55:
One of the main ways to deal with the fog is to get armour that increases the visibility, those being, the desert armour and the artifact headpieces.
As for the spawn rate  I will look a bit into it, it's something that comes up often. One thing you could do if you want to reduce the patrols could be to increase the  weight of other mobs, neutral mobs or the stone guards, as the stone guards only spawn underground.

Rodg88:
Maybe upping the spawn weights for the alex's mobs animals would be a good idea then? (and adding kangaroos too)

PR #3958, opened 2021-12-29
Make AIOT tools use SWORD affixes

telemenar:
By default AIOT Tools only take PICKAXE affixes. I filed
https://github.com/Shadows-of-Fire/Apotheosis/issues/499

Suggesting that more than one affix type should be possible for AIOT
tools given that they have multiple tool types added to them. (ToolTypes
in the API not referring to the recipe)

The dev says there that this won't happen for 1.16 but will be
considered for 1.18

However the PICKAXE affixes are not especially useful for botania tools
(Placing torches, expanded mining area, improved mining level)

So I propose that they are recategorized as SWORD for the purpose of
affixes.

I thought about changing Paxel's as well but they don't include a sword
in their recipe so I decided against it.


NillerMedDild (approved):

NillerMedDild:
Nice change, thanks for your contribution 👍

PR #3966, opened 2022-01-02
added quark cactus paste as a source of green dye

Rodg88:


MuteTiefling (commented):
Is cactus paste obtainable ? I couldn't sworn there was no way to craft it currently.

Rodg88:
i crushed cactus and got it

NillerMedDild:
I don't really see the need to add it as a dye source, what do we gain from it here?

Rodg88:
making it work with rgbee combs for making more green dye mainly

NillerMedDild:
Err.. sure :P

PR #3991, opened 2022-01-06
Enable Chiselling between glass types

### LINKED ISSUE #3968: Add Connected Glass variants to stonecutter list
### Describe your suggestion

connectedglass:borderless/clear/scratched (plus panes and coloured) glass variants should probably be added to the stone cutter, since chipped glass is stonecuttable, and it's just a simple crafting recipe to swap between the types anyway.

It might also be worth putting tinker's clear glass in the stonecutter script. However it's not just a plain crafting recipe to get it, needing the smeltery, so not sure if you'd want to let people bypass that requirement. It is just cosmetic afaik so might not be an issue

Only Chipped glass gets the stonecutter treatment currently

MuteTiefling:
Will resolve #3968

TODO: colorless glass, add untagged glasses to proper tags.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/glass.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/glass.js
@@ -1,0 +1,11 @@
+onEvent('item.tags', (event) => {
+    colors.forEach((color) => {
+        event.add(`forge:glass/${color}`, [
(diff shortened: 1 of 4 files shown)


MuteTiefling:
ok, think this one's ready, but I'm not sure how you feel on iterating on items in tags >.>

NillerMedDild:
I'm good with it, seems like the cleanest solution👍

PR #4000, opened 2022-01-07
Birthing altar updates

MuteTiefling:
Expert: Reduced cost to summon wilden:
image

All modes: Add yak
image

Slabfish
image


NillerMedDild (approved):

fajcyk:
It seems like slabfish recipe got lost somewhere, I don't see it in files.

MuteTiefling:
>.>

PR #4004, opened 2022-01-07
Ensure shell scripts/hooks are executable

Blacksmoke16:
Ensures the git hook is executable such that it runs after pulling changes within `update-server.ps1` so that the mods folder exists so that client mods can be removed. Also makes the two shell scripts executable as well.

Ref: https://discord.com/channels/195234321836146688/544589452492931076/929003233068863488


NillerMedDild (approved):

NillerMedDild:
Thank you kindly!

NillerMedDild:
Would you mind applying the same fix to the update.sh recently re-added? :)

PR #4026, opened 2022-01-09
stop if not run from root

Darkere:
I guess this works?

Figuring out the current folder is a nightmare since you can run from another working directory. But checking for mods folder seems easy enough.


NillerMedDild (approved):
Perfect thank you :)

PR #4032, opened 2022-01-10
Disable unobtainable food items, update Spice of Life food blacklist

fajcyk:
1) Disables clam food item and other food items that use clam as ingredient that are unobtainable, since clam generation is disabled in Project: Vibrant Journeys config:
https://github.com/EnigmaticaModpacks/Enigmatica6/blob/bdcca524fc9bc57d695d466ea6a2451f6dacc98e/config/projectvibrantjourneys-common.toml#L97

2) Removes some food items that are disabled, but still show up in Spice of Life food book

There is also Meat Ingot, that at the moment is only used in expert and can't be obtained in normal, maybe it should be added to SoL blacklist as well.


NillerMedDild (approved):

NillerMedDild:
Thank you for the contribution 😊

PR #4034, opened 2022-01-10
Bee Fixes

TemperedFool:
Fixes the collision between the new Dev Bee and our existing Starry Bee.

Updates the [Expert] recipe for the T2 Apiary in the Enchantment Apparatus, to use Resourceful Honey Blocks in lieu of Honey Buckets. This will prevent clogging the pedestals.


NillerMedDild:
Hmm, did anything actually change in the config?

TemperedFool:
> Hmm, did anything actually change in the config?

You need to hide whitespace changes to see it:
image

PR #4044, opened 2022-01-12
Force IPv4 ping in start-automated-server.bat

### LINKED ISSUE #4043: start-automated-server.bat doesn't support IPv6
### Modpack Version

d3638970

### Describe your issue.

Starting the server won't work on Windows if you have a public IPv6 address.

The script makes a `ping` to `minecraftforge.net` and tries to look for `TTL` in the response. Apparently, if the server returns an IPv6 address, the `TTL` is not included:

image

As I personally have a public IPv6 address, the server responds with IPv6 as well, causing the script to think, that I don't have internet.

### Crash Report

_No response_

### Latest Log

_No response_

No

### User Modifications

_No response_

Server

algorythm:
Ping does not respond with a TTL if the server responds with an IPv6 address. This essentially means that the script will think the user does not have internet, if they have a public IPv6 address. Forcing IPv4 in
the ping command, ensures TTL gets returned, and the startup can proceed.

As far as I could see, the bash script did not have that same check, so I reckon it should not be necessary to make a fix there.

Closes #4043

### DIFF
--- a/automation/start-automated-server.bat
+++ b/automation/start-automated-server.bat
@@ -420,7 +420,7 @@ IF NOT %MC_SERVER_IGNORE_OFFLINE% EQU 0 (
 )
 
 REM Ping minecraftforge before attempting download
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Thanks for the fix, but I've decided to remedy the issue by removing the check entirely 👍

PR #4046, opened 2022-01-13
[Expert] Masterful Machines - Advanced Assembly Table

MuteTiefling:
Introduces an advanced assembly table which operates at high speed. Select recipes run at 32x speed. This is largely to help with massive circuit requirements late game as some playtesters are running large arrays of assembly tables to keep up with the demand. Also reducing crafting time on the IE Auto Workbench

 

The primary goal with this is to keep the lovely animations of the original assembly table. So the structure wraps a fully functional table. Air comes in in two places, one for the MM, one for the basic table (both below. The MM then takes items in, crafts them to an intermediary, which outputs inside the glass area, accessible only to the assembly table's input. 

This finally crafts the intermediate item into the final item and outputs to the back. The back inventory is also positioned in such a way that normal crafts can input/output through there.


MuteTiefling:
Controller
image

Pressure Valve
image

Input
image

Output
image

NillerMedDild (approved):
Awesome idea and design 😀

PR #4047, opened 2022-01-13
sequenced assembly armors use the base armor

FieldArmada:
Currently the material ingot is the transitional item, which feels visually odd (chainmail -> ingot -> finished armor).

Switch the transitional item to the base armor. The custom NBT ingredient makes sure only non-partially assembled recipes can be used as an input ingredient. Otherwise, all the recipes using identical transitional item and base armors conflict.

Capture
Capture2
Capture3


Rodg88:
It was probably changed to the ingot in order to be unable to accidentally undo all in-progress steps by equipping them

MuteTiefling:
I had it this way originally and had to switch to the current method due to weirdness, yes. If I remember right I couldn't actually craft anything but the first recipe using a given armor base.

They only good way to do this would be to create dummy items for every single armor type to use as an intermediary. That was a lot more work than I was willing to put in for a visual, so using the ingot is the next best compromise.

MuteTiefling:
Yeah, actually looks really good. just tested. Very clever solution :D

PR #4048, opened 2022-01-13
Saner default rankds

hvdklauw:
Gives the admin default access to the commands, and thus would allow singleplayers to actually get access to the command through `/ftbranks add @p Admin` instead telling  people to edit a file.

### DIFF
--- a/defaultconfigs/ftbranks/ranks.snbt
+++ b/defaultconfigs/ftbranks/ranks.snbt
@@ -4,22 +4,31 @@
 		power: 1
 		condition: "always_active"
 		ftbranks.name_format: "<{name}>"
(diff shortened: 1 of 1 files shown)


Rodg88:
If someones in single player, wouldn't just enabling cheats allow access to the commands? Allowing everyone access to the commands might just mean players on servers overdo force loading?

hvdklauw:
Nope, because the ftb ranks just disables them, cheats enabled or not don't matter. If you are with cheats enabled the command won't work either.

Rodg88:
Oh my bad, you mean the /back/home commands. I thought you meant chunk loading commands

NillerMedDild:
Good change, I thought OPs had access already tbh.

PR #4051, opened 2022-01-13
[Expert] QoL Tweaks

MuteTiefling:
Oxidizer no longer requires sink to make it auto crafting friendly
image

Sea Shelf uses regular prismarine to make it craftable :D
image


MuteTiefling:
Blood shards are currently very slow to craft due to the high mana cost of the incense stick. This brings the mana cost down per shard and gives the incense stick a nice bonus roughly equivalent to it's mana cost compared to the other two. 
image

MuteTiefling:
Adds missing undergarden items to the market.

MuteTiefling:
Convert apiaries over to honey blocks as per Tempered's notes.

image
image

TemperedFool:
Oooooh thank you. I only thought to do T2 as it actually jammed, but this is better for consistency.

MuteTiefling:
Removed custom Syrup bottle. Using Thermal's instead (kept our texture). 
Updated Rune of Spring recipe using it to also use tulips instead of any sapling
image

Moving recipes out of thermal filler, as it's bugged right now (all _custom_ recipes cost 1000mb of fluid, no matter what)
Reinforced Stone:

MuteTiefling:
increase yield on stardust when processing the ore through Thermal Pulverizer. Gives it a nice boost when using better catalysts.

MuteTiefling:
Add magma cream crucible recipe
image

MuteTiefling:
Think that's it for now.

NillerMedDild (approved):

PR #4056, opened 2022-01-15
Some Atum QOL suggestions

FieldArmada:
In commit order, feel free to take none or any of these:

* Silk touch is borderline a downside in Atum because you can't process Bone, Relic, or Khnumite. Add smelting recipes for bone (usually drops 1-3) and khnumite (usually drops 2-4). I don't know what to do about Relic Ore; I don't think there's a sensible way to average that loot table.
* Atum gold / iron tools and armor are now recyclable like their vanilla counterparts. Makes the bandit and chest trash slightly less annoying at least.
* Add `forge:ores` to Bone Ore and Relic Ore so that they scan properly. Also make Bone Ore scan as common.
* Flax seed is, unlike every other seed, not compostable.
* `Quandry` -> `Quandary`.
* Add the dye QOL that exists for normal wool and carpet to linen and linen carpet. I was thinking about doing this for ceramic and crystal glass too, but it seems like there's already something in place for ceramic and there are so many versions of glass.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/compote/composting.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/compote/composting.js
@@ -24,7 +24,9 @@ onEvent('recipes', (event) => {
                 { item: 'sushigocrafting:wasabi_root', chance: 0.65 },
                 { item: 'sushigocrafting:soy_bean', chance: 0.65 },
                 { item: 'sushigocrafting:cucumber', chance: 0.65 },
-                { item: 'sushigocrafting:rice', chance: 0.65 }
+                { item: 'sushigocrafting:rice', chance: 0.65 },
+
+                { item: 'atum:flax_seeds', chance: 0.3 }
             ],
             remove: [],
             change: []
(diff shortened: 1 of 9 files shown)


MuteTiefling:
> I don't know what to do about Relic Ore; I don't think there's a sensible way to average that loot table.

It should probably just melt down to gold ingots, to be honest ;D

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/compote/composting.js
```diff
@@ -24,7 +24,9 @@ onEvent('recipes', (event) => {
                 { item: 'sushigocrafting:wasabi_root', chance: 0.65 },
                 { item: 'sushigocrafting:soy_bean', chance: 0.65 },
                 { item: 'sushigocrafting:cucumber', chance: 0.65 },
-                { item: 'sushigocrafting:rice', chance: 0.65 }
+                { item: 'sushigocrafting:rice', chance: 0.65 },
+
```
NillerMedDild, line 28: Found a wild newline😄

NillerMedDild (approved):
Nice changes you've made, thank you very much for contributing :D

I think Relic Ore -> Gold Ingot works, would you mind adding that to this PR?

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/compote/composting.js
NillerMedDild: Oh nevermind, I see it follows the convention of that file ^^'

NillerMedDild:
Nice, thanks!

PR #4058, opened 2022-01-15
Add Sawing recipes for Ancient Wood and Bark

MuteTiefling:
Adds sawing recipes for ancient planks and bark
image


MuteTiefling:
this has no stripped variant, so would be a royal pain to put in the normal unification.

PR #4068, opened 2022-01-16
manually add glass to stonecutter tag

MuteTiefling:
Seems they're not getting added to the tag in their new home, causing a bunch of warnings in the logs


NillerMedDild:
Oh right, my bad ^^' forgot to use the right event.

MuteTiefling:
No trouble. Wasn't quite sure if this was the right place with its new home. If there's a better way just let me know.

PR #4072, opened 2022-01-17
Fix tc quests giving wrong molten metals

FieldArmada:
These two quests give the wrong fluid, which can't actually be used for anything.


MuteTiefling:
Looks good, thanks for your contribution :D

PR #4075, opened 2022-01-17
[Expert] Machine Speed and Ore Processing tweaks

MuteTiefling:
Speeds up crusher more to compete with Create crusher a bit better. 

Speeds up IE Metal Press by allowing it to do 4x recipes. This keeps the animation speed less insane, but ultimately makes it faster than the Thermal press. 

Buffs Mek solar. Still less than Powah solar and lower than other options at that stage, but this gives a little bit of reward for the high cost of crafting them early game.

Speeds up custom magical ore processing due to mana/aura throughput limitations by reducing required mana/aura. Also removes some consumables from the process and increases output rates.

### DIFF
--- a/config/configswapper/expert/config/Mekanism/generators.toml
+++ b/config/configswapper/expert/config/Mekanism/generators.toml
@@ -1,0 +1,95 @@
+
+#Mekanism Generators Config. This config is synced between server and client.
+[generators]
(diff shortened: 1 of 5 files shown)


MuteTiefling:
Gonna call this one good here

NillerMedDild (approved):

PR #4077, opened 2022-01-18
[Expert] Buff power output of Modular Routers

MuteTiefling:
Aim here is to make Routers viable again for powering IF Laser Drill arrays. With this buff, a router can pretty comfortably power a 3x3x3 cube of lasers, assuming it has full power/speed upgrades. 

Also alters the recipe to be later game as a result. 

image

### DIFF
--- a/config/configswapper/expert/config/modularrouters-common.toml
+++ b/config/configswapper/expert/config/modularrouters-common.toml
@@ -1,0 +1,167 @@
+
+[Module]
+	#Should Breaker modules show particle effects when working?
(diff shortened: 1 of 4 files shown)


MuteTiefling:
Gonna take that Thumbsup as OK to merge as it's going to be easier to fix my derp of committing this to my dev branch instead of a separate branch :(

NillerMedDild:
Hehe yes, all good to merge, I was just waiting for the draft status to change before reviewing.

PR #4079, opened 2022-01-18
Make nbt extracters not require the programmer

Rodg88:
The programmer is a lot more expensive in Expert, so using it as a crafting ingredient for a QOL item is maybe a bit much. In normal mode, the programmer is fine as an ingredient as it's trivial to make, and that's what the vanilla nbt extractor recipes are balanced around. Replaced with an input variable transformer, as the block 'reads' the card you insert.

Could instead be a 'network reader' to add some cables to the cost, and may be more thematic?


NillerMedDild:
Could you perhaps add screenshots of the changed recipes? 😀

Rodg88:
Logic Programmer updated recipe:
image

NBT Extractor:
image
Extractor Remote:
image

MuteTiefling (approved):

PR #4084, opened 2022-01-19
Fluid Storage and Automation quests

FieldArmada:
I found that the Storage quest chapter is an excellent enumeration of every way to store items, but not so much for liquids or item transportation.

Would this be useful? Wanted to check before I put some time into writing descriptions :P

Capture


telemenar:
Not a pack dev but I love the idea. Though it is very hard to be complete. 

For item transfer could also toss in:
Blood Magic Nodes
Botania Corporea
If the Omni-Hopper represents PNC as a whole it could call out Logistics/Drones
Ars has the Carby
Occultism has some spirits as well. 
Create also has the Arm.

jlfree12:
You might also want to mention the ender chests as cheap wireless cross dimensional item transport.

NillerMedDild:
Ooh yes please :D

FieldArmada:
Did a first pass of descriptions.

config/ftbquests/quests/chapters/storage.snbt
```diff
@@ -1166,5 +1166,373 @@
...
+				id: "6A4A37F812BE7CA6"
+				type: "xp"
+				xp: 100
+			}]
+		}
+		{
+			title: "Ender Tank"
+			x: 10.0d
+			y: -3.0d
+			description: ["Ender Tanks will share their inventories with other tanks using the same color code. Try them for pipeless fluid transport."]
```
NillerMedDild, line 1290: I think you should mention that you can set the color code network to privat by clicking with a diamond.

NillerMedDild (approved):

NillerMedDild:
The rewards aren't great, but everything else is nice, good work! 
Could you supply a screenshot of the chapter please?

FieldArmada:
What's usually done for rewards? Is it just the `_ Delight` random food items?

NillerMedDild:
We do that often yeah, but we generally prefer items that match thematically with the quest - Resources you would need to craft more of the item in question, upgrades, that sorta thing.

Rodg88:
I'm wondering if we should add this to a new chapter under Logistics, "Automation"? Maybe tanks can stay in storage, but move transport into it, and also put stuff like redstone control in there? Could be good for Routers too, which probably don't need a whole chapter?

As a quick example:
image

NillerMedDild:
> I'm wondering if we should add this to a new chapter under Logistics, "Automation"? Maybe tanks can stay in storage, but move transport into it, and also put stuff like redstone control in there? Could be good for Routers too, which probably don't need a whole chapter?
> 
> As a quick example: image

Good idea 👍

FieldArmada:
Made the quest rewards the random food items.

Made the automation chapter.

automation
Capture

FieldArmada:
Also does anyone know how to reload FTBQuests faster than full-restarting the game? I must have wasted an hour or two doing that :(

NillerMedDild:
Looks good to me 👍 thanks for the contribution 😊 

As for reloading, unless you edit in-world a full reload is required 😞

Rodg88:
Now this is in dev, I'm going though the quests and touching some up etc, looking good thanks!

PR #4086, opened 2022-01-19
Add pre-genned NBT to thermal cells, fixes RS autocrafting with them.

Rodg88:
Adding the NBT from a newly crafted thermal cell to the default output for thermal cells stops RS autocrafting getting really confused. Just re-encode your patterns for the energy and fluid cells after updating.

This is in the kube expert branch as I don't think normal mode has anything that uses crafted thermal cells as inputs? Could go in kube base branch instead though without hurting anything, if the `if (global.isExpertMode == false)` conditional was removed.


NillerMedDild (approved):
Looks good to me, we may need to change the values if we ever reconfigure how much they can hold though.

PR #4097, opened 2022-01-21
[Expert] Stellar Neutron Activator MM

MuteTiefling:
Credit to Telemenar for the basic structure and inspiration.


MuteTiefling:
Controller:
image

Power In
image

Fluid In
image

Fluid Out
image

Mana In
image

Not the greatest display here, but Water, power, and mana go in, tritium comes out:
image


This takes 4 pools worth of mana all at once, then burns 100k fe/t and 64k mb/t of water to produce 640 mb/t of liquid tritium. Each burst burns for 4000 ticks, so it comes down to 1000 mana per tick to generate tritium. 

The intention with this is to reduce some of the multiblock lag from massive mekanism setups, especially on servers where each individual has a tendency to build a fusion reactor and all the machines to go with them. The evap towers are particularly rough. One of these should be sufficient to run a Fusion reactor at full speed (max injection + max d-t injection).

MuteTiefling:
A note on why this outputs liquid tritium: Simply put, MM hasn't updated to Mek's 10.1 API yet and won't until they finish their re-write. Gases can be produced, but get voided upon exporting them from the machine. Not very ideal :D

config/masterful_machinery/controllers/stellar_neutron_activator.json
```diff
@@ -0,0 +1,32 @@
+{
+    "controllerId": "stellar_neutron_activator",
+    "textureOverride": "kubejs:block/hempcrete",
```
NillerMedDild, line 3: Can we not reference the original texture? I don't really like copying mod assets here 🤔

NillerMedDild (approved):
Looks awesome 👌

config/masterful_machinery/controllers/stellar_neutron_activator.json
MuteTiefling: Not that I could make work, no. IE puts it in a sub folder and MM couldn't access it :(

MuteTiefling: just had another thought I'll try.

MuteTiefling: Yeah, I was derpin, gave up too quick

NillerMedDild (approved):

config/masterful_machinery/controllers/stellar_neutron_activator.json
NillerMedDild: Nice!

PR #4099, opened 2022-01-22
Added Quark deepslate and new Ars stones to stonecutter

Rodg88:
Added Quark deepslate and the new Ars "smooth" arcane stones to stonecutter:
image
image

Also minor info tab fix for corundum, which was getting applied to irrelevant items.

### DIFF
--- a/kubejs/client_scripts/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/item_modifiers/jei_descriptions.js
@@ -191,7 +191,7 @@ onEvent('jei.information', (event) => {
             text: ['Place next to living coral to infuse.']
         },
         {
-            items: [/quark:\w+_crystal$/],
+            items: ['#quark:crystal'],
             text: [
                 'Will grow up to four blocks tall if placed deep underground. Will emit particles while growing.',
                 ' ',
--- a/kubejs/client_scripts/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/item_modifiers/jei_descriptions.js
@@ -191,7 +191,7 @@ onEvent('jei.information', (event) => {
             text: ['Place next to living coral to infuse.']
         },
         {
-            items: ['#quark:crystal'],
+            items: [/quark:\w+_crystal$/],
             text: [
                 'Will grow up to four blocks tall if placed deep underground. Will emit particles while growing.',
                 ' ',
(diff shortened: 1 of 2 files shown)


telemenar:
Technically desperate shouldn't stone cut into cobbled. Unless Stone cuts into Cobblestone.

Rodg88:
> Technically desperate shouldn't stone cut into cobbled. Unless Stone cuts into Cobblestone.

Other stones can be cut into cobble variants, eg create scoria, or even diorite etc. Just not plain vanilla grey stone.

Rodg88:
Added fix for https://github.com/EnigmaticaModpacks/Enigmatica6/issues/4101

Not changing slabs, as there currently isn't any recipe for turning slabs to full blocks, and adding blanket rules for that would start to interfere with some chiseled block recipes

kubejs/client_scripts/item_modifiers/jei_descriptions.js
```diff
@@ -191,7 +191,7 @@ onEvent('jei.information', (event) => {
             text: ['Place next to living coral to infuse.']
         },
         {
-            items: [/quark:\w+_crystal$/],
+            items: ['#quark:crystal'],
```
NillerMedDild, line 194: Big nope, you cannot use tags in client scripts sadly :(

NillerMedDild (changes requested):

kubejs/client_scripts/item_modifiers/jei_descriptions.js
Rodg88: seems to be working? 
image

NillerMedDild: Only in singleplayer sadly, client scripts usually load before server scripts when joining a server, causing tags to be unavailable/unbound.

Rodg88: Ah I see, well the regex is slightly beyond me so I'll revert for now

NillerMedDild: It'll tag anything that begins with `quark:` and ends with `_crystal` :)

NillerMedDild (approved):

NillerMedDild:
Thanks for the fixes yet again :D

kubejs/client_scripts/item_modifiers/jei_descriptions.js
Rodg88: Yeah, it just needs to also exclude the waxed variants, and myalite crystals, which currently get caught by that regex. I could just add the 9 corundum to a item list, might be the simplest option lol

PR #4116, opened 2022-01-23
Add more BYG mushrooms to the market

mcilhargey:
Add some mushrooms from BYG to the market place, as many produce either shroomlights or materials for crafts but are missing since nether biomes were removed.


MuteTiefling:
Looks good, thanks!

PR #4117, opened 2022-01-23
[Expert] Cleanup, bring ID crafting/inventory access earlier, buff IE wires

MuteTiefling:
IE Wires get a 4x boost on power transfer stats and can stretch twice as far per wire. 

Reduce the cost of Corporea sparks
image

Engineers are a thirsty lot. Needs more tea
image

Change RS cables to use ID cables as a base
image

Variable Transformers made cheaper
image
image

Static Light Panel
image

Display Panel
image

Storage Terminal
image

Alt Item Interface for early access
image

Item Filter/Import/Export
image

Fluid
image


MuteTiefling:
This should allow people to use the ID Item stuff earlier than the rest, including using it as a mini crafting/storage interface similar to Pretty Pipes or Refined Storage. Autocrafting with ID is still gated, as are fluid and energy capabilities.

Rodg88:
that's 3x more redstone wires/connectors per rs cable now, they gave 12 per before.. oof

MuteTiefling:
Yeah. Same ratio to quartz infused iron though, which felt like the most expensive part of the craft

NillerMedDild (approved):

PR #4161, opened 2022-01-30
Automation quests and other minor quest fixes

Rodg88:
WIP, but the chapter is hidden. Pretty Pipes quests are in the chapter now so might be confusing if it gets released as is, as the storage chapter references them but they won't be accessible.

### DIFF
--- a/config/ftbquests/quests/chapters/automation.snbt
+++ b/config/ftbquests/quests/chapters/automation.snbt
@@ -33,8 +33,8 @@
 		}
 		{
 			title: "Logistical Transporters"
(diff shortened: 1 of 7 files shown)


Rodg88:
Should be ready now!
image

Rodg88:
Also some additions in Storage 
image

And one more in Tinker's for the filtered ducts.

Rodg88:
Also included fixes to the IF quests (they had some weird dependencies), and one or two spelling errors.

MuteTiefling:
Cool thanks!

PR #4163, opened 2022-01-30
Change patchouli to be an all-modes config

Rodg88:
Disables advancement locks on all modes.


Rodg88:
I hope this is the correct way to change from using config swapper to not using it for a general config file

NillerMedDild:
It is correct gj 👍 
However, you also need to force delete the old files through the serverconfigupdater config, as launchers don't typically remove files during modpack updates.

Rodg88:
gotacha, done.

NillerMedDild:
Thank you kindly 😀

PR #4165, opened 2022-01-31
[Expert] RFTools Machine Frame Models/Textures

TemperedFool:
Adds the textures (by @Ridanisaurus of course) to finish up the RFTools Machine Frame crafting process:

image

I've also adjusted the names (and ID's) to better match the process: Rough Machine Frame Top, Coated Machine Frame Top, Rough Machine Frame.

WARNING: THE ID CHANGES ARE BREAKING. Expert players probably don't have the intermediate items lying around in quantity, but this will mess with existing crafting patterns. We've also seen all of the patterns in a crafter get wiped when a recipe becomes invalid - @Darkere thinks this is related to alloy recipes that are registered as crafting recipes, but I would advise being cautious and removing those patterns prior to upgrading.


MuteTiefling:
Looks good to me

PR #4170, opened 2022-02-01
[Expert] Change final step to use the blood altar instead of runic altar

MuteTiefling:
just didn't make a lot of sense to spend a full living rock on each ingot.

image


NillerMedDild:
Is the final output meant to be a nugget?
Also, I don't see the removal of the Runic Altar recipe?

NillerMedDild:
Ah right, thanks for clearing that up 👍

PR #4184, opened 2022-02-03
Industrial Deuterium Plant

TemperedFool:
Adds a MM to centralize Deuterium production. Output matches the Solar Neutron Activator MM. Structure:

image

Structure REQUIRES water source blocks below the platform, and surrounding the pipes that drop down through the deck. Water isn't actually supplied as part of the recipe - just molten sulfur, PNC air, Create rotation, and FE - so this is "how the structure is supplying water" for the reaction. Things like Kelp and Seaweed where the structure expects water source blocks will disrupt the structure - this is by design.

Production Recipe:

image


NillerMedDild (approved):
Damn that looks cool :D
Would you perhaps consider renaming it `Industrial Deuterium Plant`? I'm not too fond of the scale in there currently :P

TemperedFool:
No problem - I'll change the name when I commit recipes for the ports & controller.

TemperedFool:
Recipes for controller and ports:

Controller:
image

Power Input:
image

Fluid Input:
image

Fluid Output:
image

Rotation Input:
image

Pressure Input:
image

TemperedFool:
Also removing the standard Deuterium and Tritium recipes. This will require Expert players use the MM's for progression into Fusion power.

TemperedFool:
Still need to remove the unused ports; can do that later, we can merge this for testing if I don't get to it first.

PR #4197, opened 2022-02-06
Fix Stasis Chamber power use.

Rodg88:
Makes the Stasis Chamber use the intended 60k rf/t, as the cycle time is actually 25 ticks, not 50.

https://user-images.githubusercontent.com/2611674/152673720-ebae7bd7-e3a9-4766-863a-ccd14f32e1e2.mp4


MuteTiefling:
Ah thanks. Silly configs.

PR #4208, opened 2022-02-07
Add Irradiated Occultism Miner for fissile fuel production

theboo:
Adds occultism mineshaft miner for fissile fuel materials.

Also boosted the rolls per operation of the current djinni ore miner.


theboo:
just noticed the config seems to override the recipe for the ore miner output, forgot the config existed tbh... Can switch to that unless we plan on bumping that miner up to marid tier. I'll fix that in either direction tomorrow

MuteTiefling (approved):
Looks good. Just needs that rate increase we discussed.

theboo:
should be good to go now, I think? also added damage:0 with weakNBT to the iesnium pick in the irradiated recipe as i saw it being used elsewhere while scrolling through

MuteTiefling:
Yeah, good catch with the pick.^_^

PR #4244, opened 2022-02-16
[Expert] PNC updates

MuteTiefling:
Add Security upgrade to default Vacuum Pump craft and update components
image

Night Vision Upgrade
image

Inventory Upgrade
image

Entity Tracker Upgrade
image

Item Life Upgrade
image


MuteTiefling:
Update to tinker's armor as well, as it appears to have changed:

image
All pieces use Cobalt now.

NillerMedDild (approved):

PR #4254, opened 2022-02-18
[Expert] Disable Eidolon crucible and move all recipes to other crafts

MuteTiefling:


ZephyrWindSpirit:
Why?

novaexe:
amazing change 👍

PR #4332, opened 2022-03-04
disable random jukebox upgrade at mobs

xILikepuddingx:
mobs wearing a backpack have a certain chance of getting a jukebox upgrade
this change prevents that


NillerMedDild:
Thanks for your contribution, but I like the Jukebox Backpack feature, and want it to remain :)

Issue #4333, opened 2022-03-05
Add Storage Parts to Advanced assembly table.


Kevin-Marsh:
### Describe your suggestion

At the moment only the base crafting components are available for the advanced assembly table. I am suggesting that you add the storage parts in too since it is an extremely long craft at the higher levels for RS.

### Why would you like this added/changed?

Save time making large disks.

MuteTiefling:
well, it's a ton of extra items in JEI, but it wasn't too bad to do.

PR #4334, opened 2022-03-05
[Expert] Add storage parts to advanced assembly table

MuteTiefling:
Same as the other advanced assembly recipes. Just batches up the craft to make it faster. 30x per run, so 30x faster

image

image

image


MuteTiefling:
Resolves #4333

PR #4336, opened 2022-03-05
Add new main menu panorama

theboo:
Was also planning on using config swapper to have a different packmenu.cfg for rid's expert logo, but either config swapper doesn't work with .cfg files, or the menu assets are loaded before the file is swapped because it does not seem to work.

new panorama:
image


Darkere:
.cfg is a custom format. Only toml is supported right now. I guess I could think about building in a fallback that copies and pastes the entire file.

PR #4338, opened 2022-03-05
Blacklist BM and Occultism from cardboard box. Expand restrictions to chalk and BM altar runes

MuteTiefling:
Blacklist BM and Occultism from cardboard box. Expand restrictions to chalk and BM altar runes


Quezler:
having to move each of the 4 dusk runes by cart is a pain :D

MuteTiefling:
... Why are you moving them? Just chalk them in with the ritual builder wand.

Quezler:
when you don't have dusk unlocked yet you can find these while exploring, just enough for an early well of suffering ^-^

PR #4339, opened 2022-03-05
Raise sea level in Terraforged preset up to y=62

theboo:
Raise sea level to 62 (vanilla default), terraforged default is 63, and we are at 58 right now. Tested with a few worlds and getting plenty of surface caves and surface water at sea level 62. This should help with some of the ocean structures looking kinda jank. When we first lowered the sea level, I had to go below 60 to start seeing surface water/caves so maybe something was fixed on TF's side since then as the change was made in our preset quite awhile ago.

This will not affect current worlds, only new worlds will get the new sea level.

image


NillerMedDild:
Boooo... Why does this PR have a menu panorama in it... 😄

theboo:
lmao yeah, i messed up and didnt put the panorama PR in its own patch... if the panorama PR isn't accepted ill re-do this one

NillerMedDild:
No worries I just need to approve your PRs faster :diy-catjam:

PR #4360, opened 2022-03-08
Add packmenu config to config swapper for menu logo

theboo:
Config swapper will need updated for this to swap the .cfg file. Thanks to Darkere for both updating configswapper and making the kubejs startup script to have the packmenu config get reloaded after swap.

Tested changing to expert and back to normal mode, works perfectly

### DIFF
--- a/config/configswapper/expert/config/packmenu.cfg
+++ b/config/configswapper/expert/config/packmenu.cfg
@@ -1,0 +1,132 @@
+# Configuration file
+
+"forge info" {
+    # The anchor point for this element. [default: FORGE]
+    S:"Anchor Point"=FORGE
+
+    # The X offset for this element. [range: -1000 ~ 1000, default: 0]
+    I:"X Offset"=0
+
+    # The Y Offset for this element. [range: -1000 ~ 1000, default: 0]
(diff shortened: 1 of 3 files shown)


Darkere:
note: the normal config file needs to exist already for this to work.

NillerMedDild (approved):

NillerMedDild:
Nice work you two!

NillerMedDild:
Have you tested what happens in the following flow?

- Launch game in normal mode
- Join world
- Use `/mode expert` command
- Leave world

Which icon appears?

Darkere:
The wrong one, since this script only runs on launch. 

I still want to reload powah configs too on mode change. Would put this in there too. It's next on the todo list.

gabriel-andreescu:
Hi, this change introduces a crash @world load:
[https://pastebin.com/raw/QQSwJdyH](https://pastebin.com/raw/QQSwJdyH)

After further investigation I found this:
`(only configs in TOML format are supported at this time)`
source: [Config Swapper](https://www.curseforge.com/minecraft/mc-mods/config-swapper)

NillerMedDild:
Alright 👍

PR #4361, opened 2022-03-08
Update Terraforged Preset to include all the new structures

### LINKED ISSUE #4349: With the inclusion of when dungeons arise, the terraforged config preset of enigmatica needs to be updated for structures.
### Describe your suggestion

Basically what it says on the tin; the new dungeons arise isn't included on the terraforged config. 

Certain structures should be limited in size.

theboo:
This wont affect anything in current or new worlds as it updates itself whenever someone starts a world and sees new structures that it doesn't have, we just haven't updated it on the repo in a minute

fixes #4349

### DIFF
--- a/config/terraforged/presets/Enigmatica.json
+++ b/config/terraforged/presets/Enigmatica.json
@@ -73,6 +73,228 @@
         "salt": 10387313,
         "disabled": false
       },
+      "dungeons_arise:abandoned_temple": {
+        "spacing": 30,
+        "separation": 20,
+        "salt": 43726252,
+        "disabled": false
+      },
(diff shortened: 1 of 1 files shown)


NillerMedDild (approved):

Thiana:
When you say it wont affect current worlds? Do you just mean existing chunks or does that include new chunks in worlds generated before the change? It sounds like the latter which leads to the question is there any way to force terraforged to update for an existing world? Thanks.

theboo:
This won't effect new worlds or old worlds in any way. All of this was already auto generating in people's preset files with these same settings when they created their worlds, now we're just shipping it like this for visibility.

Issue #4433, opened 2022-03-18
Gold duplication using powered rails + smeltery


Hamderber:
### Modpack Version

1.0.0

### Describe your issue.

Creating Powered Rails using the Transport mod's method (using 6 gold rods) creates 24 Powered Rails, but Tinkers' Construct still lets you melt each Powered Rail into a Gold Ingot. This results in a very simple loop that can quadruple gold quantity per iteration for the simple cost of one redstone, stick, and a bit of lava for a smeltery.

### Crash Report

N/A

### Latest Log

N/A

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Both

NillerMedDild:
Thank you for the report, I will investigate the issue.

TemperedFool:
I removed the smelting recipes for regular and powered rails (pending PR acceptance); that addresses this loop on both.

Issue #4423, opened 2022-03-17
Partial coke & blast bricks can be placed, wiping NBT


Officialy:
### Modpack Version

1.0.0

### Describe your issue.

Upon placing a partial coke brick or blast brick, you wipe the NBT data required for the Sequenced assembly to work/continue working upon them. A solution to this would be to make it an item, with the same block model. Making it look the same, however just not placeable.

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Both

MuteTiefling:
I'm on the fence here. Honestly, I'd rather see this setup be automated, so you're not manually moving the block around.

I agree it sucks losing that progress though.

TemperedFool:
I'm wondering why placing them wipes the NBT - is that normal?

Honestly I think we should just make them items, rather than blocks - I agree we want players to automate, but there's also not really any reason for people to place them?

I think we can even tag them with `create:upright_on_belt` to get them to sit up on belts even. I can test later.

MuteTiefling:
most blocks that retain their NBT on placement/pickup are tile entities with special handling by the mod. a simple block doesn't have anywhere to store that nbt and retrieve it when it's picked up.

TemperedFool:
Makes sense. I'll test swapping them to items, see if it breaks anything; I don't see why it would tbh.

PR #4435, opened 2022-03-18
[Expert] Sequenced Assembly Blocks -> Items, [Base] Remove Rail Melting Recipes

TemperedFool:
For expert: changes the Partial items for IE multi block bricks from blocks to items, preventing placement and fixing #4423. New textures and models graciously provided by @Ridanisaurus as always. Because the ID is not changing, it appears this will have no effect on in-flight crafts on update - I did test this with one of the partial bricks.

Beautiful new brick pile models:
image

In hand:
image

On a Belt:
image

Rid also had include some updated textures for bricks and blend:
image

And Basalt Powder:
image

For Base: While I was at it I also addressed #4433 - I removed the melting recipes for both regular and powered rails, preventing duplication of gold and iron ingots via smeltery melting.


NillerMedDild (approved):

NillerMedDild:
Looks good to me!

PR #4443, opened 2022-03-19
Reload powah and randompatches configs on mode change

Darkere:
Haven't actually tested on a server but since the networking is also used in SP, it should work just fine on a server too.


NillerMedDild (approved):
Nice work!

Issue #4479, opened 2022-03-23
Server - Locking when structures generate above build height


YoItSoul:
### Modpack Version

Enigmatica6-1.0.0.zip

### Describe your issue.

When players are exploring, it appears that the server soft crashes (never stops responding) when structures such as the air ship are generated above the world height. Players cannot move and console in unresponsive to commands other than killing the java PID.

This is simply speculation but some of my players have connected when/where the issues occur.

The only error I could catch from the logs appear as: 

```[Server thread/WARN] [minecraft/Chunk]: Tried to load a block entity for block Block{minecraft:air} but failed at location BlockPos{x=-8424, y=261, z=-2747}```

(*no crash report, since it never stopped entirely, just locked up*)

### Latest Log

https://gist.github.com/YoItSoul/ad41863bf827c3b2acd954b25677f317

### Have you modified the modpack?

Yes

### User Modifications

The only modification we have to the modpack is a "welcome" quest and Discord integration.

### Did the issue happen in singleplayer or on a server?

Server

NillerMedDild:
Thanks for the report - I found it reported on their tracker here: https://www.curseforge.com/minecraft/mc-mods/when-dungeons-arise/issues/38

theboo:
Any chance you could grab a screenshot of the structure at that location?

YoItSoul:
> Any chance you could grab a screenshot of the structure at that location?

Literally can't. Server locks up when loading the chunk.

YoItSoul:
Requires myself to edit the playerdata position in order to unlock the server.

theboo:
Damn, alright... Well I'm going to disable these structures for the next update, was hoping to see what this one was specifically... After disabling you'll prolly just need to delete those chunks and should be good I would think

YoItSoul:
@theboo So it appears that the issue is primarily coming from the airships as they tend to be the only structure that can spawn x high above an already high world gen.

YoItSoul:
Multiple sources have confirmed that the "cursed chunks" as my community calls them, contain air ships. - Specifically the ones that may have phantom mounted enemies?

theboo:
yup yup, ty... i spoke with the dev of dungeons arise and was told which structures to disable... this wont help those chunks tho, after next update when those structures are disabled, u will have to delete those chunks to purify the curse xD

PR #4487, opened 2022-03-24
Disables flying structures and infested temple

theboo:
These structures have a possibility of partially generating above build limit causing watchdog crashes in terraforged worlds


MuteTiefling:
#4479

Issue #4526, opened 2022-03-29
[Expert] Astral Sorcery Quest Step Inconsistency


carloverso:
### Modpack Version

1.0 Expert

### Describe your issue.

I can just barely pull off E6E with questbook and google, but i have found an issue with Astral Sorcery quest.

Currently it's like this:
https://i.imgur.com/K8bNEP8.png

You craft the Luminous crafting table, but it says linking tool is required. But if you don't know AS, you don't know how to get AS rocks. You can't make linking tool, so you can't make the table.

After googling, searching discord etc. I realised how to find rocks with Resonating wand, made linking tool, made table, quest completed, but suddenly.

**There's Rock Crystal quest**, which shows all that. But its after Crafting Table one? And it gives you linking wand? Do you need 2 linking wands?

I suggest it be like this:
https://i.imgur.com/NkWdDlc.png

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Both

PR #4530, opened 2022-03-30
Explicitly state in the Expert quests how to obtain rock crystals. Remove superfluous Linking Tool from Astral Quests. 

MuteTiefling:
Explicitly state in the Expert quests how to obtain rock crystals. Remove superfluous Linking Tool from Astral Quests.


MuteTiefling:
Resolves #4526

PR #4542, opened 2022-04-01
add tconstruct:slime_crystal tag to slime crystals

zifnab06:
Tags tinkers slime crystals so they can be used in attribute filters. 

image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/tconstruct/crystals.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/tconstruct/crystals.js
@@ -1,0 +1,3 @@
+onEvent('item.tags', (event) => {
+    event.get('tconstruct:slime_crystal').add(/tconstruct:\w+_slime_crystal/);
+});


NillerMedDild:
I don't really see the point of this tbh? 😄

zifnab06:
My bulk blasting setup looks like this: image

There's 3 options I could do here: 
- run the belts slow enough the item smelts by the time it makes it to the funnel on the far side
- run the belt at a normal speed, use a funnel + item filter, add every smelted item to the filter
- (what I'm doing) run the belt at a normal speed, use a funnel + attribute filter, add tags to the filter like this: image

Without this change, the only two options available for slime crystals (which are used in RS crafting recipes in expert mode) also appear on sky/ender slimeballs (being, tinkers general items + added by tconstruct), so there's no way to filter them with an attribute filter.

MuteTiefling:
Attribute filter should have a "is blast able" filter. Or "is smelt able"

zifnab06:
it does, once again there's an easier way to do things than what i'm doing. thanks, i'll close this.

PR #4575, opened 2022-04-05
Allow more items to stack to 64

MuteTiefling:
Mostly QoL. These ones shouldn't break anything.

### DIFF
--- a/kubejs/startup_scripts/better_stacks.js
+++ b/kubejs/startup_scripts/better_stacks.js
@@ -1,0 +1,31 @@
+onEvent('item.modification', (event) => {
+    const increaseStackSize = [
+        'minecraft:ender_pearl',
+        'rftoolsbase:infused_enderpearl',
+        'rftoolsbase:infused_diamond',
+        'minecraft:snowball',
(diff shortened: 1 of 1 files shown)


NillerMedDild:
Do the IF upgrades not depend on NBT?

MuteTiefling:
They have NBT on the, yes. But it's all the same for a given tier of upgrade. They already stack to 16 as is, this just brings it up.

NillerMedDild:
Alright 👍

Discomanco:
Could put in requests for more items, like empty Buckets and Honey Bottles

MuteTiefling:
honey bottles seem kinda powerful, given their potion effects/high saturation.

TemperedFool:
Maybe just vanilla honey bottles?

theboo:
Oreos stack to 64 and they kinda destroy any honey lol

MuteTiefling:
That's fair. Still harder to get than most honey. But I think we can do honey too

Discomanco:
Signs of all kinds
Eggs of probably multiple kinds

Issue #4040, opened 2022-01-11
Shapeless recipe for Source Berry Roll can't be crafted via Create mixer


juh9870:
### Modpack Version

269a02559f907eb207390aa92b414f5e915d90aa (expert)

### Describe your issue.

Unlike other shapeless recipes, recipe for Source Berry Roll can't be crafted via Create shapeless mixing

### Have you modified the modpack?

Yes

### User Modifications

Added https://www.curseforge.com/minecraft/mc-mods/when-dungeons-arise

### Did the issue happen in singleplayer or on a server?

Singleplayer

NillerMedDild:
Thank you for the report, I will investigate the issue.

PR #4583, opened 2022-04-06
Temporary fix for missing Source Berry Roll recipe

nddragoon:
Issue #4040 has been open and under investigation for a few months now, and while this doesn't fix the underlying issue cause I can't for the life of me figure out what's disabling the normal shapeless mixing recipe, this gives a way to easily automate source berry rolls in the meantime.


NillerMedDild:
Thanks for the fix! :D

PR #4587, opened 2022-04-07
Fix inconsistencies and sort modlist

H4ckerxx44:
I saw that the modlist was unsorted (terrible for searching if you don't know `ctrl+f` and thought a sorted list would suite better.
Also fixed broken names where there was just a version number.


NillerMedDild (approved):
Thanks!

PR #4610, opened 2022-04-10
Swap the caged heart and apprentice blood orb

Quezler:
Since you need the heart to craft the blood orb & people tend to read from left to right.


MuteTiefling:
Fair enough, thanks :D

PR #4644, opened 2022-04-14
[Expert] Remove dependency on IE coke oven

MuteTiefling:
It's slow, you use it once, and then you're done forever. No point.

Pyrolizer will take it's place instead: 
image

### DIFF
--- a/config/ftbquests/quests/chapters/expert__tier_1_wip.snbt
+++ b/config/ftbquests/quests/chapters/expert__tier_1_wip.snbt
@@ -234,7 +234,7 @@
 				id: "45F596F7AFA8BFFF"
 				type: "item"
 				item: "immersiveengineering:cokebrick"
-				count: 27L
+				count: 17L
 			}]
 		}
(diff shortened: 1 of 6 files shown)


MuteTiefling:
Pyrolyzer recipes made roughly 3x slower. Coal blocks may be processed, however, processing 9 coal in the time that 8 would run as single items.

Should still be plenty fast with upgrades.

image


Tinker's Crystals added to IE Blast Furnace, process in 100 ticks as opposed to 400 in vanilla blast furnace.

image

NillerMedDild (approved):
I reluctantly approve 😄
Did you update the quests to reflect the change?

MuteTiefling:
I think I got everything, could use a second pair of eyes though.

NillerMedDild:
Alrighty,  I'll merge and have a look before next update 😀

NillerMedDild:
I added an extra line to the Coke Oven quest :)

PR #4659, opened 2022-04-17
[Expert] Buff Mek Energy Cubes compared to Powah cells

MaxNeedsSnacks:
Given the relatively high resource cost of Mekanism Energy Cubes in Expert mode, I figured they were due for a buff, especially because Powah cells currently absolutely blow them out the water otherwise :p

I decided *against* buffing them linearly respective to their "relative" Powah tiers, and instead balancing them around each other, as well as around the relative resource cost that goes into making the corresponding Mekanism alloy (which, in my eyes, get exponentially more difficult the higher you go), so their efficiency compared to the Powah cells varies from tier to tier.

Current rates are as follows (capacity / I/O):
Basic — (16MFE / 32kFE), around (1.6x / 1.6x) as good as a Blazing tier Powah cell
Advanced — (32MFE / 64kFE), around (1.28x / 1.28x) as good as a Niotic tier Powah cell
Elite — (128MFE / 256kFE), around (2.13x / 2.56x) as good as a Spirited tier Powah cell
Ultimate — (1.024GFE / 2.048MFE), around (7.3x / 5.12x) as good as a Nitro tier Powah cell

Two notes: 
- The Advanced Energy Cube seems a little bit weak compared to the Elite counterpart especially, however I think this is fine as the entry gate for the alloy used to make it is just a Metallurgic Infuser and Energised Steel, while Reinforced Alloy requires a whole Lithium setup.
- The Ultimate Energy Cube might appear a little ridiculous, however given the *exorbitant* cost of Atomic Alloy, I think it's more than justified that it is this powerful ^^

### DIFF
--- a/config/configswapper/expert/config/Mekanism/tiers.toml
+++ b/config/configswapper/expert/config/Mekanism/tiers.toml
@@ -1,0 +1,264 @@
+
+#Tier Config. This config is synced from server to client.
+[tier]
+
+	#Chemical Tanks
+	[tier.chemical_tanks]
+		#Storage size of Basic chemical tanks in mB.
+		#Range: 1 ~ 9223372036854775807
+		basicStorage = 64000
+		#Output rate of Basic chemical tanks in mB.
+		#Range: 1 ~ 9223372036854775807
+		basicOutput = 1000
+		#Storage size of Advanced chemical tanks in mB.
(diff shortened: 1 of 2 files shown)


NillerMedDild (approved):
Excellent change, thanks 👍

PR #4663, opened 2022-04-17
Remove tags from Atum crystal glass/panes in expert

theboo:
With Atum crystal glass/panes and its stained variants being more expensive in expert and used in specific recipes, it makes sense to not have them in the glass tag where it might get used instead of cheaper glass in recipes. Like when using JEI autofill in occultism storage.

Removed from all tags because one of the other tags it has is being added to forge:glass causing it to get re-added after removal.


NillerMedDild (approved):
Noice 👌

PR #4669, opened 2022-04-18
Recipes to drain resourceful honey bottles to thermal centrifuge

thraaawn:
By default the Thermal centrifugal separator can drain honey from vanilla honey bottles.
This change makes it possible to also drain the resourceful honey bottles in it.

The Create drain is already capable of doing this without using power, so this shouldn't break expert mode balancing and makes honey post processing more convenient - no flying glass bottles.


MuteTiefling:
Looks good, but could you please resend and target develop instead of master?

NillerMedDild (approved):
Nice change, thanks!

PR #4671, opened 2022-04-18
Flip the default recipe for the redstone flux coil

Quezler:
> this pull request flips the default recipe shown in order to have the gold nuggets & rod line up with the item's texture.

before: 
Screen Shot 2022-04-18 at 18 24 09

after:
Screen Shot 2022-04-18 at 18 24 31

you could already do it this way (jei just didn't suggest it), now that the recipe is flipped the old one isn't suggested & still works (aka: backwards compatible with all prior automated craftings, this is not an automation breaking change 😗)


NillerMedDild (approved):
Very nice 👌

PR #4687, opened 2022-04-21
[Expert] RAM Chip Improvements

MuteTiefling:
Introduces new advanced RAM Chip

image
@Ridanisaurus Please save me from my 'art' :D

These can be used in place of RAM Chip 8E in many recipes

image
image
image

Notably, this means a less resource intensive and faster (hopefully) craft for Basic Memory Chips.

Un-nest CPU crafting recipes. Each tier is now made directly on an unassembled PCB
image
image
image

Lastly, introduces alternate recipes using Magic to enhace the output rate of making circuits:
image

image

image

Overall these changes are aimed at reducing the grind in circuit making, while also rewarding players who have progressed deep in both tech and magic.


NillerMedDild (changes requested):
Excellent changes, good idea to use slates instead of runes in the extra circuit recipes.
I think you forgot to commit that part though, I can't spot it in the changes 😅

MuteTiefling:
ah, yep, good catch!

NillerMedDild (approved):

PR #4698, opened 2022-04-23
Update Apotheosis configs to new JSON format

MuteTiefling:
Puts our old caps back in place for the new version, but also updates the items used for various upgrades:

Normal Mode:
image
image
image
image

### DIFF
--- a/config/apotheosis/enchantments.cfg
+++ b/config/apotheosis/enchantments.cfg
@@ -167,7 +167,7 @@
 
 "apotheosis:earths_boon" {
     # The max level of this enchantment - normally 3. [range: 1 ~ 127, default: 5]
-    I:"Max Level"=5
+    I:"Max Level"=3
 
     # A function to determine the max enchanting power.  The variable "x" is level.  See: https://github.com/uklimaschewski/EvalEx#usage-examples [default: ]
     S:"Max Power Function"=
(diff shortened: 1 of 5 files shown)


Shadows-of-Fire:
I might suggest you add new "higher tier" modifiers that have extended caps (specifically for the delay modifiers)
You could also add "super" modifiers that provide buffs to multiple stats at once, which might be neat™

MuteTiefling:
Expert Recipes: 
image
image
image
image
image
image
image

MuteTiefling:
Limit Boon of the Earth (drops ores when breaking stone) to Level 3 and overworld or chunks.

NillerMedDild (approved):
Very nice 👌 
If I hadn't looked at the json I wouldn't know to dual wield the required items, any chance you could have a look at making JEI more explicit here @Shadows-of-Fire? :)

theboo:
there is a book now!! i believe it explains this in detail!

MuteTiefling:
It actually shows it quite nicely in JEI, just not in the pictures above:

image

image

image
image


The limits are also displayed there
image

Shadows-of-Fire:
Yep, it's all explained as well as it can be between both the book and JEI.

NielsPilgaard:
Oh lovely, my bad 👍

Shadows-of-Fire:
I want to note (before moto forgets) that this omits the Silent modifier, added in 4.8.99B, so that'll have to be added to the expert changes (for consistency).

MuteTiefling:
Thanks for the reminder :D Easier to keep track of things here

Issue #4700, opened 2022-04-23
Improve Discoverability of Amadron Survey Tools


fistsh:
### Describe your suggestion

The **Amadron Survey Tools** item seems like a very cool way to trade for certain rare / annoying-to-get resources without needing to spend ages mining for them. However, I am fairly certain I don't speak just for myself when I say that **I likely would never have discovered them** if I hadn't been directly told of their existence by an Enigmatica 6 developer.

The existence of this item and its usefulness could be increased by doing any/all of the following, among other things:

- Increase quantity/consistency of references to it in the quests: As explained below, only relying on the description of one specific quest to impart information like this is insufficient, particularly in Expert mode.
  - Have both Amadron quests refer to the Enigmatica-specific trades, instead of just the one quest.
  - Potentially even add this info to other, related quests (e.g. in Adventure -> Godforged Blocks, "If you need more Nebu, check out the Amadron tablet trades!" ??)
- Add a Tooltip (lore) line to the Amadron Survey Tools, describing their use to obtain a cache of other-dimensional resources. IIRC, this can be done with something like as follows:
  
  In a KubeJS startup script (probably `minecraft\kubejs\startup_scripts\item_registry.js`, though **some refactoring seems necessary** to allow for this):
  `event.create('amadron_survey_tools') /* etc. */ .tooltip(NBT.toTag('{"translate":"item.kubejs.amadron_survey_tools.tooltip","color":"aqua"}'));`
  
  And in the lang files (e.g. `minecraft\kubejs\assets\kubejs\lang\en_us.json`), a line like this:
  `"item.kubejs.amadron_survey_tools.tooltip": "Used to trade for a variety of materials from other dimensions!",`

### Why would you like this added/changed?

- The Amadron Survey Tools are primarily discoverable by players in the manner of reading the description of the Amadron quest - and **not the Expert Amadron quest**, but the one under the PNC questline. However, this is not an ideal way of imparting information to players - there are literally hundreds of quests in Enigmatica, and more in Expert mode, and players will not want to stop and fully read each and every task description on the _off chance_ that there's something modpack-specific there.
  - This goes double for Expert mode, where a lot of trades have an Expert-questline duplicate, and in this case, the Expert-questline Amadron quest _does not_ sufficiently mention the E6-specific trades, only saying "perhaps you'll be surprised" (which is vague enough to be useless in this respect).
[...]

PR #4702, opened 2022-04-23
Amadron Improvements

MuteTiefling:
Give free survey tools for Amadron quest
Add tooltip to Survey Tools to draw more attention to them.
image


MuteTiefling:
resolves: #4700

PR #4715, opened 2022-04-26
Make pink enderslime use increments of 250

kirjorjos:
Pink endersliime doesn't have an ingot version and every recipe but 2 uses increments of 250.


MuteTiefling:
Thanks for the submission. Needs to target develop instead of master, however. Could you switch it over?

NielsPilgaard (approved):

PR #4716, opened 2022-04-26
Computercraft disks preloaded with recipes

Quezler:
in our current e6e playthough i have automated both the enchanting apparatus and mechanical crafters with computercraft.

for the enchanting apparatus it was rather easy, i just had to put the 3x3 pattern in by hand into an rs pattern terminal and then export it twice (one for in the rs crafter, and another one in a chest nearby which computercraft could read the recipes. (and their reagent item) from)

for mechanical crafting recipes its a different story, both the rs pattern terminal and the large pattern are tedious to define recipes in, the default once since it scrolls and you only have a width of 3 to work with, and the bigger one since you'll have to find and drag all the recipe items from jei into it one-by-one. (doesn't really give the automation vibe)

so after some digging & help from various people i managed to export all possible mechanical crafting recipes into a format that can be copy-pasted into computercraft (aka: gave computercraft access to jei), the player (e.g. me) will still have to write his own code for which recipe it is based on the input items & put the right ones in the right slots, so this is not spoonfeeding.

whilst this is a computercraft disk, it can probably still be used by other mods, for example a pneumaticcraft drone interface, or printing it out on paper with a printer & reading that nbt via integrated dynamics. (or just using the two generated json files and reading/modifying it externally for whatever purpose that suits your needs)

the two json files containing the recipes generate after kubejs is done loading custom recipes, this way they're mode independent & update themselves automatically to new releases, this means minimal maintenance for pack maintainers and players.

the output format for each recipe type varies slightly, based on how easy it is to access/transform into a readable format:
[enchanting_apparatus_recipes.json.txt](https://github.com/EnigmaticaModpacks/Enigmatica6/files/8565252/enchanting_apparatus_recipes.json.txt)
[mechanical_crafting_recipes.json.txt](https://github.com/EnigmaticaModpacks/Enigmatica6/files/8565253/mechanical_crafting_recipes.json.txt)
(enchanting apparatus has an useless type field & items are _#'d and not an array, whilst the mechanical crafting ones use arrays for alternatives where tags are at play, but hey at least you have the recipes in a readable format ^-^)

the `recipes.lua` file in each disk basically acts as a `.gitkeep` since `JsonIO.write` cannot create the directories themselves, and gives the user some convenience since they only have to use `local recipes = require("disk/recipes")` in order to access it as a lua table.

i do not expect many people playing this pack to be experienced enough with computercraft to have some use out of this, it will not get in the way of them (apart from maybe some extra milliseconds when loading), this is aimed at technical ones.

leaving this as a draft since there are still things to sort out (e.g. replacing the 2nd if with an ifelse, possibly unifying the recipes into more identical formats, adding disks for other crafting methods that could benefit from a preloaded list, etc)

even if this never gets merged it will still act as a nice reference for how to generate a recipe json file locally to then manually import into a single player run or wgetting on a server.

🥔


Quezler:
Due to my desire to re-use this for arc furnace input splitting i fell deeper into the rabbit hole than ever before, after 3 extensive threads on latcord it just seems impossible to access the json of the recipes frm within kubejs. (for 1.16)

This draft pull request has now been transformed into something that makes no sense to merge as is:
It basically now depends on the output of `/kubejs export` + the server being started with kubejs common debug true.

There are some [github actions](https://github.com/Quezler/Enigmatica6/actions/runs/2245630483) going on that will try to launch the server in both modes and generate a list of all recipes for each, my hope was that its artifacts could be pulled & inserted into the curseforge release, but this json file alone (unminified) is just a few mb's shy of what is uploaded to curseforge, so that ship has sailed.

So i transformed it even more in my other branch: into a script that could also be ran somewhat standalone:

1) set `debugInfo=false` to `debugInfo=true` in `./kubejs/config/common.properties`
2) launch the server
3) run `/kubejs export` when you can (ignore the ~66 kubejs errors, that's kubejs debug mode talking)
4) shut down the server
5) set `debugInfo=true` to `debugInfo=false` in `./kubejs/config/common.properties`
6) put the contents of `treasure_disks.js` in a file in the server's root & run it with node
7) apply the changes made to `./kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/create/mechanical_crafting.js`

or as an alternative to option 7, just spawn the item in for yourself:
`/give @p computercraft:treasure_disk{"Title": "all recipes by ID", "SubPath": "kubejs/recipes", "Colour": 11829822}`
(ps, do option 7 before 2 if you want that recipe addition itself to be part of the disks generated after as well)

this should be enough instruction for anyone wanting access to all the recipes with computercraft to apply it to their own instance, in its current stage bundling it by default in the modpack makes just zero sense and can't really be justified. 🕹️

NielsPilgaard (commented):
Jesus christ, you've been busy :P cool stuff!

Quezler:
Moving the code to [its own repository](https://github.com/Quezler/kubejs-computercraft-recipes) so it can be [run](https://github.com/Quezler/kubejs-computercraft-recipes/actions/runs/2301343555) independently of whatever branch it is on. (curseforge server files)
Feel free to close this one to clear clutter. (i planned to do that now, but with 5 hearts and counting it seems beloved) ❤️

PR #4721, opened 2022-04-27
[Expert] Introduce some late game alternative recipes for Mechancial Crafts

MuteTiefling:
The Mechanical crafter is OK as a gate, but it seems like once you've done it and progressed a little further, making more of those same machines probably shouldn't require it. 

Alternate Portality Controller:
image

Alternate Disk Drive
image

Alternate Dissolution Chamber
image


MuteTiefling:
Assembly Table Alternates
image
image
image

Which are then all assembled on the Assembly table
image

As well as cheaper Furnace Engines on Assembly Table:
image
image

MuteTiefling:
Alternate Pyrolyzer

image

NielsPilgaard (approved):
Looks good to me 👍 nice work!

Issue #4749, opened 2022-05-02
Increase liquid starlight yields from fluid laser base


Ratlanta:
### Describe your suggestion

Liquid starlight yield from fluid lases base should be scaled up.

### Why would you like this added/changed?

Currently the fluid laser base yields 10mb of starlight per completed operation. Even when scaling up drills with speed upgrades, the output is still abysmal and probably less than a single lightwell running on resonating gems. 

In late game, most players will probably want to scale up their lightwell generation to fuel a gaia reactor. The primary method of starlight generation (aquamarine to lightwells) is not very server friendly. The fluid laser base makes sense as a viable late game alterative for starlight generation.

OneEyeMaker:
Also what about alternative ways of obtaining liquid starlight? Maybe a "fluid" bee (requiring Celestial bee in recipe/mutation)?

MuteTiefling:
I do like alternates. It won't be a bee, but I've been giving mulling over some other ideas as alternatives.

NielsPilgaard:
I think ramping up the output from the Fluid Laser would be good too, even if there are alternates.

Quezler:
> Status: Fixed In Next Release?

PR #4763, opened 2022-05-03
[Expert] LP and Liquid Starlight Improvements Part Two

MuteTiefling:
image

Adjust Slate to LP to be about 1.5 buckets per slate. This is a highly variable process, but should result in more than the base 500LP you get by popping these by hand. 

Greatly increases the duration that Resonating gems last in the lightwell, as well as boosting their output.  Given their cost to make, this makes them pretty beefy at making liquid starlight.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_fluid.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_fluid.js
@@ -20,22 +20,6 @@ onEvent('recipes', (event) => {
             id: `${id_prefix}oil`
         },
         {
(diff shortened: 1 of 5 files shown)


MuteTiefling:
Crafting recipe for Slate Ampoule: 
image


Otherwise attainable via bees (late, needs Tier IV Altar) or manually killing things with syringes. This opens it up to an automation process, and well in advance of the Well of Suffering. 

I'd still expect the Well or a Fluid Laser to be the ultimate source of LP, but this should reduce the early game grind by allowing the player to bank LP without having to do so manually.

MuteTiefling:
Increases Starlight output from Lasers. Now requires the End.

MuteTiefling:
Completes #4749

NielsPilgaard (approved):

PR #4781, opened 2022-05-05
A way to upgrade powah reactors

Quezler:
It would be neat if powah reactors were (losslessly) upgradable, on my expert multiplayer run players insisted rushing for a `niotic` tier reactor while still running only on refined fuel dynamos. (eventually they managed to even rush to `nitro`)

If the reactor tiers were upgradable (or dismantleable) it would not have been such a waste of the (non-powah) reactor ingredients since afterwards its still usable, but just a waste of materials. (basically the same sales pitch as if black hole)

My goal with this idea is to find a recipe type that makes sense & works, aka being able to take several ingredients and several outputs, originally i had the arc furnace in mind but of course those ingredients don't stack:

Screen Shot 2022-05-05 at 20 54 24

```
    powahTiers.forEach(function (this_tier, i) {
      if (['starter', 'nitro'].includes(this_tier)) {
          return;
      }

      function casingMaterial(tier) {
        switch(tier) {
          case 'basic':    return '#forge:storage_blocks/lead';
          case 'hardened': return '#forge:storage_blocks/energized_steel';
          default:         return `#forge:storage_blocks/${tier}`;
        }
      }

      let next_tier = powahTiers[i + 1];

      data.recipes.push({
        input1: Item.of(`powah:reactor_${this_tier}`, 36),
        secondaries: [
          Item.of(`powah:energy_cell_${next_tier}`, 2).ignoreNBT(),
          Item.of(`powah:thermo_generator_${next_tier}`, 3).ignoreNBT(),
          Item.of(`powah:furnator_${next_tier}`).ignoreNBT(),
          Item.of(casingMaterial(next_tier), 6)
        ],
        outputs: [
          Item.of(`powah:reactor_${next_tier}`, 36),
          Item.of(`powah:energy_cell_${this_tier}`, 2),
          Item.of(`powah:thermo_generator_${this_tier}`, 3),
          Item.of(`powah:furnator_${this_tier}`),
          Item.of(casingMaterial(this_tier), 6)
        ],
        time: 100 * 36,
        energy: 51200 * 36,
        id: `${id_prefix}upgrade_${this_tier}_reactor_to_${next_tier}`
    })
```

Attempt two was using a pressure chamber (code for that is this pr), it appears to work as intended and just swaps out the blocks through suction:

Screen Shot 2022-05-05 at 22 15 54

The recipe does look a bit ugly and unclear in jei, gathering the items by hand is a bit of a pain, but it does however work:
Screen Shot 2022-05-05 at 22 29 00
(output barrel ^)

Alternative ways could be that the pnc chamber just straight up returns all the ingredients of the reactor (what about the nbt for the maxwell spirit?), or a sort of kubejs rightclick event listener on a set of mechanical crafters that just instantly puts items in their slots, honestly idk it sounds weird.

Of course the 36 blocks could also be upgraded one-by-one via sequenced assembly or other means, but that will require some calculations and balancing in regards to crystals needed for the next tier and the power usage the energizing orb will have taken.

I wouldn't advice merging this as-is, it appears that only one of each unstackable is consumed:
Screen Shot 2022-05-05 at 22 34 14
(notice the items in the hotbar after breaking the chamber ^)

Leaving this here as a place to discuss potential ways of dismantling or upgrading reactors in a fair way. ^-^ 🥔

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/pressure_chamber.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/pressure_chamber.js
@@ -540,11 +540,52 @@ onEvent('recipes', (event) => {
             id: `powah:crafting/ender_gate_${tier}`
         });
     });
+    
+   powahTiers.forEach(function (this_tier, i) {
+      if (['starter', 'nitro'].includes(this_tier)) {
+          return;
+      }
+
+      function casingMaterial(tier) {
+        switch(tier) {
+          case 'basic': return {
+            tag: 'forge:storage_blocks/lead',
+            item: 'emendatusenigmatica:lead_block'};
+          case 'hardened': return {
+            tag: 'forge:storage_blocks/energized_steel',
+            item: 'powah:energized_steel_block'};
+          default: return {
+            tag: `forge:storage_blocks/${tier}`,
+            item: `powah:${tier}_crystal_block`};
+        }
+      }
+
+      let next_tier = powahTiers[i + 1];
+
+      recipes.push({
+        inputs: [
+          { item: `powah:reactor_${this_tier}`, count: 36 },
+          { item: `powah:energy_cell_${next_tier}`, count: 2, unstackable: true },
+          { item: `powah:thermo_generator_${next_tier}`, count: 3, unstackable: true },
+          { item: `powah:furnator_${next_tier}`, count: 1, unstackable: true },
+          { tag: casingMaterial(next_tier).tag, count: 6 }
+        ],
+        pressure: -0.75,
+        results: [
+          { item: `powah:reactor_${next_tier}`, count: 36 },
+          { item: `powah:energy_cell_${this_tier}`, count: 2 },
+          { item: `powah:thermo_generator_${this_tier}`, count: 3 },
+          { item: `powah:furnator_${this_tier}`, count: 1 },
+          { item: casingMaterial(this_tier).item, count: 6 }
+        ],
+        id: `${id_prefix}upgrade_${this_tier}_reactor_to_${next_tier}`
+      });
+    })
 
     recipes.forEach((recipe) => {
         let ingredients = [];
         recipe.inputs.forEach((input) => {
-            input.type = 'pneumaticcraft:stacked_item';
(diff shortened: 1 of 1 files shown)


AlphaP2712:
I think in different way, in expert we need 196 crystal of that teir you wan to craft  and normal mode we need 144. so we need around  4 - 6 crystal per 1 reactor block , for my suggestion (in expert atleast) just have only basic teir recipe for reactor, and now you need to craft like upgrade teir form buttom to upper with recipe like this for every teir
image
So that ,every teir upgrade you will need 144 crystal of that teir to upgrade ,So you can't rushing too much because of resource you need for next teir , I think this make sense in both expert and normal , because if have some revert recipe like you suggest ,So now we  have  last teir furnance , magmator and energy cube ,that.. to make new reactor? Naa... we still need more crystal to upgrade them whatever.
Maybe in normal mode can be skip teir form basic to Nitro ,like Other powah upgrade,But for Expert. Progression craft form basic to hardend to blazing.....  to Nitro seem more makesense. while current recipe of nitro can craft without need any spirited teir. (and yes because of that I just rush to Nitro when I get nitro crystral )

MuteTiefling:
Progression crafts aren't coming back. They were removed in normal, kept out in expert. That's just far too much cost to take every single block up through every single tier.

The only exceptions being energizing rods and the player charger.

All other generators have direct upgrade recipes from any lower tier to the one you want.  It makes sense that reactors would too.

Personally, I like the idea of using the pressure chamber for this, but like the other upgrade recipes, you shouldn't get the old crystals back. There needs to be some loss.

Sucking out the components, minus the crystal blocks, could be a decent way. Then you can manually upgrade the components when existing recipes, and rebuild by injecting it all back in with high pressure.

AlphaP2712:
hmm... ok, I respect in your decision. Just feel like something that size 3x3x4 that can produce like 500kFE with not that much resource and very safe to operate is a little bit powerful.

MuteTiefling:
Gonna mark as won't fix because I've taken a slightly different route. Thanks for the idea though :D

Quezler:
Well technically this was just a idea + code and not a way, and since you made a pr it should be labeled as accepted ^-^

MuteTiefling:
eh, either way

NielsPilgaard:
Marking as won't fix so it won't show up in the changelog for next version :)

PR #4791, opened 2022-05-07
Add `blazing blood` to the `fluid laser`

Quezler:
Expert mode is rather heavy on blazing blood, it would be neat if it was obtainable without large quanties of blaze entities:

While of course many expert recipes require large amounts of drops an thus large mob spawner setups with e.g. drygmys or looting blazing blood can only be obtained from melting either live blazes or their heads, both of those require entities.

This pull request merely allows you to use a fluid laser with a yellow lens over a single blaze to gain blazing blood.

(ps, this was edited online on github and was not yet tested/confirmed to work ingame, but it felt better than a suggestion)


NielsPilgaard (approved):
Looks good to me, thanks for making the PR! We'll fix it if it breaks 😄

PR #4794, opened 2022-05-08
[Expert] Reduce cost of Ethereal Slates

MuteTiefling:
This rate is proving nearly impossible to match. Brings the creation cost more in line with the other tiers of slates.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/bloodmagic/altar.js
@@ -206,10 +206,10 @@ onEvent('recipes', (event) => {
         {
             input: 'bloodmagic:demonslate',
             output: 'bloodmagic:etherealslate',
-            syphon: 200000,
+            syphon: 25000,
             altarLevel: 4,
-            consumptionRate: 1000,
-            drainRate: 1000,
+            consumptionRate: 40,
+            drainRate: 40,
             id: `${id_prefix}etherealslate`
         }
     ];


XenonDB:
Well, seriously? I thought this recipe is designed for expert mode and designed an automation system:

What you need is just an altar with 58 Rune of Sacrifice, 32 Rune of Capacity, 14 Rune of Augmented Capacity and continuing supplying 20~24 mob to sacrifice.
The capacity of altar will be increased to nearly 200k and the speed of blood increase in altar is fast enough to fill altar during making a tier 4 slate from tier 1 slate.
Since the altar is filled when the process of making a tier 4 slate is done, there's enough blood in altar to make tier 5 slate.

Just make sure to supply slates which tier is maximize 1 to let altar have enough time to prepare blood.

thatGuyCalledMax:
not opposed to them being cheaper though, they are quite a lot more then the others

MuteTiefling:
When they were first set up this way, they were only used for like.. one thing. Now they're part of making circuits cheaper and the cost of them is out of line with that use.

Ratlanta:
I thought crafting my first couple of ethereal slates for an iridescent was a fun and satisfying problem to solve. Couldn't you just make ultimate circuits use demonic slates if that is the purpose of this change?

NielsPilgaard:
> I thought crafting my first couple of ethereal slates for an iridescent was a fun and satisfying problem to solve. Couldn't you just make ultimate circuits use demonic slates if that is the purpose of this change?

This is a pretty good point I think

MuteTiefling:
alrighty then. I'll revert and swap the circuit slates around instead.

Issue #4771, opened 2022-05-04
Remove some of the annoyance with Occultism based crafts


Swedz:
### Describe your suggestion

Items that have been crafted using Occultism pentacles currently have randomly generated names of spirits attached to them, creating unique NBT data for each item created. It would be nice if this was not the case. This can be done easily by modifying the recipes in `server_scripts/enigmatica/kubejs/expert/recipetypes/occultism/ritual.js` to use the `ritual_type` of `occultism:craft` instead of `occultism:craft_with_spirit_name`.

### Why would you like this added/changed?

Especially in Expert Mode where almost all things need to be automated, it makes it quite annoying to set up crafts with Refined Storage since each item will have a different NBT and will never fulfill the request. Not only does it prevent RS from completing crafts automatically, but it makes the items no longer stackable.

DoomSquirter:
I would go so far as to ask not just the obvious ones that shouldn't have demon names, but any that are used in recipes.  like the satchel, etc.  it would help a lot to be able to subcraft these things instead of pulling together a long list of stuff to accomodate the items.

MuteTiefling:
I've mixed feelings on this, honestly.

First, they provide a nice bit of theme to the game that really puts a finger on the fact that you've bound spirits into these items.
Second, while it does make it impossible to do 'simple' recipes in RS, the tools are there to manage this.

I set these recipes knowing full well that RS wouldn't deal with them cleanly by itself; that the player would have to get creative in how they approach them. I even went so far as to show an [example](https://wiki.enigmatica.net/enigmatica6/gameplay/how-to.../occultism-automating-rituals) of it in our Wiki to help people wrap their heads around the process. 

I fully understand that some people find this to be an annoying process, but I don't see it as all that different from many other 'off RS' automations required here, or even in other packs. It's not uncommon that you need to send a batch of items out to a chest to have some other smarter system handle the craft for you. That's all that's happening here.

NielsPilgaard:
I think it provides a nice extra challenge, albeit annoying 😄

Buuz135:
It is a problem when an output has a random name and RS keeps waiting for it to finish the crafting, like the Dimensional Matrix.

Swedz:
[...]

PR #4800, opened 2022-05-08
[Expert] Alterations to Occultism Ritual outputs

MuteTiefling:
Attempts to improve some of these crafts, to remove the annoyance of crafting.

Logic Director moved away from the Occultism crafts. Makes it available earlier as well. 
image

Remove Sophisticated Backpacks upgrades from Compacting Drawers
image

Ender Storage no longer requires satchels. Similar gating.
image
image
image

Inception Upgrade no longer uses a satchel either:
image

Modular Routers Senders/Pullers
Mk3
image

Mk2s
image
image

Enchantment Library
image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/modularrouters/shapeless.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/modularrouters/shapeless.js
@@ -5,13 +5,20 @@ onEvent('recipes', (event) => {
     const id_prefix = 'enigmatica:expert/modularrouters/shapeless/';
     const recipes = [
         {
-            output: Item.of('2x modularrouters:sender_module_3'),
+            output: Item.of('modularrouters:sender_module_3'),
+            inputs: [Item.of('modularrouters:sender_module_2').ignoreNBT(), 'integrateddynamics:logic_director'],
+            id: 'modularrouters:sender_module_3'
+        },
+        {
+            output: Item.of('4x modularrouters:sender_module_3'),
             inputs: [
                 Item.of('modularrouters:sender_module_2').ignoreNBT(),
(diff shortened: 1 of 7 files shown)


MuteTiefling:
This should resolve #4771, though not in the way requested. 

Effectively all crafts using the satchels have been shifted over to using dimensional storage crystals. This keeps the occultism aspect intact, but serves to reduce the 'annoyance' down to a single craft: the dimensional storage cyrstal itself. The process doesn't change for that, but it does reduce the overall complexity of crafting all of these items.

Similarly, the logic director was removed from occultism entirely and brought forward a touch, as ID's gating has been a constant source of discussion in Discord. 

Finally, Compacting drawers (the other common cause for this all to come up) have had their gating changed, while still maintaining the magic feel of the recipe. Hopefully squashing those issues as well.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/modularrouters/shapeless.js
```diff
@@ -22,7 +29,7 @@ onEvent('recipes', (event) => {
                 'tconstruct:sky_slime_crystal',
                 'tconstruct:sky_slime_crystal'
             ],
-            id: 'modularrouters:stack_upgrade'
+            id: `'modularrouters:stack_upgrade'`
```
NielsPilgaard, line 32: Typo I think 🤔

NielsPilgaard (approved):
Excellent changes 😊👍
Feel free to merge when you see fit.

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/modularrouters/shapeless.js
MuteTiefling: Well that's a weird place for that to happne >.>

PR #4801, opened 2022-05-08
[Expert] Recycle Powah Reactors

MuteTiefling:
In response to #4781

Allows extracting of the major components from reactors, at a loss of many internal components.

This allows upgrading the internals and re-building as a higher tier.

image

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/pressure_chamber.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/pneumaticcraft/pressure_chamber.js
@@ -519,15 +519,28 @@ onEvent('recipes', (event) => {
         }
     ];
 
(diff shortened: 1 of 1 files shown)


Quezler:
It kinda misses the point of my original idea, my main annoyance was having to throw away all the other components.

Sure, they are not hard, but when you finally scrap together the resources for your first reactor you would like those back (or transferred) when upgrading your reactor to the next tier, currently everything not circled in green is discarded:
Screen Shot 2022-05-08 at 17 28 59

Now that i'm later down the line in expert mode the dielectric casings/the 6 crystal blocks/advanced cables/etc don't feel as expensive or annoying to craft now that it's all automated and the resource stockpiles are immense.

Ideally as much as possible should be refunded or transferred, but please at least do the 6 blocks & spirit as well 🤔

NielsPilgaard (approved):

PR #4803, opened 2022-05-08
Betterend ore automation

MuteTiefling:
Add's Thallasium to the Occultism miner and all better end ores to the laser drill. This stuff is needed in larger quantities in Expert and had no means of automation prior to this.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_ore.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_ore.js
@@ -1,4 +1,6 @@
 onEvent('recipes', (event) => {
+    const id_prefix = 'enigmatica:base/industrialforegoing/laser_drill_ore/';
+
     const endBiomes = {
         type: 'minecraft:worldgen/biome',
         values: end_biomes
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_ore.js
```diff
@@ -112,20 +103,17 @@ onEvent('recipes', (event) => {
...
-            },
+            output: { item: 'emendatusenigmatica:dimensional_chunk' },
             rarity: [
                 {
                     whitelist: endBiomes,
                     blacklist: {},
-                    depth_min: 0,
-                    depth_max: 255,
+                    depth_min: 30,
+                    depth_max: 60,
```
NielsPilgaard, line 116: what's the reasoning behind this change? :)

NielsPilgaard (approved):

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/industrialforegoing/laser_drill_ore.js
MuteTiefling: Just seemed like the focused version (higher weight). Should be a narrower range now that there's actually more than one ore in the end.

NielsPilgaard: Ah makes sense 👍

Issue #4805, opened 2022-05-08
Reduce the overall cost of the excavation mastery shard [E6E]


thatGuyCalledMax:
### Describe your suggestion

To reduce the overall price of the mastery shard for excavation.

### Why would you like this added/changed?

Its understandably meant to be expensive, but its just not in the same ball park as the other the other mastery shards. Maybe 5x shards per craft would be a better number? Its not impossible to make now, its just significantly more then all the other shards

thatGuyCalledMax:
Another request in this vein would be to reduce the cost of the battle mastery shard. The offender there is mainly the 400 total energy upgrades you need

MuteTiefling:
Thanks, will have a look at these soonish.

Ratlanta:
Agree with Max on battle mastery. I think I can actually pull off excavation in it's current form although it doesn't remotely match up to the requirements for other shards. Even if excavation was 5 per craft, it would still be exponentially more expensive than all shards short of battle. 

Battle Mastery in it's current form will probably require several weeks or even longer AFK to get the mats required. That's assuming I massively scale up my entire production. I'm currently around 15 TPS so I don't know how much scaling I can realistically do at this point. The end of omnifactory was quite similar so I understand if that is the intent. 

Graphite and starlight changes I believe you already have planned is going to be a big help compared to what we have on 1.2.1 as well.

MuteTiefling:
The intent was for them to be tough, but not weeks of AFK. I'll for sure be reviewing these to bring the material requirements down. You guys are some of the first to get this far, so there was not a lot of feedback on how rough these get :D

PR #4814, opened 2022-05-10
[Expert] Mastery Rebalance

MuteTiefling:
Excavation Masteries are now crafted 2 at a time. Added Irradiated Ore Miner Marid to the list of items
image

In terms of godshards, this brings things down considerably. about 3.2k Seth godshards, 1.4k Osiris. There are still some high volume items, like blaze and niotic crystals, but I think it's much easier for the player to scale those up.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/masterful_machinery/enigmatic_tree_of_life.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/masterful_machinery/enigmatic_tree_of_life.js
@@ -221,13 +221,14 @@ onEvent('recipes', (event) => {
         },
         {
             outputs: [
-                { type: 'masterfulmachinery:items', data: { item: 'kubejs:excavation_mastery_shard', count: 1 } }
+                { type: 'masterfulmachinery:items', data: { item: 'kubejs:excavation_mastery_shard', count: 2 } }
             ],
             inputs: [
                 { type: 'masterfulmachinery:items', data: { item: 'industrialforegoing:fluid_laser_base', count: 1 } },
                 { type: 'masterfulmachinery:items', data: { item: 'industrialforegoing:ore_laser_base', count: 1 } },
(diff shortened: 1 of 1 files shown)


TemperedFool:
Uh, are those energy costs for Spirited and Nitro off by a decimal place? 😅

MuteTiefling:
Nope :D

MuteTiefling:
Induction Cell cost reduction. Each tier uses fewer of the previous tier. With the use of gaia ingots, it becomes a straight upgrade path. 
image
image
image
image

MuteTiefling:
Reduce the number of energy units needed per meka suit
image

Shield Projector now uses nitro tier instead of spirited tier. Also reduces the number of Powah inputs in the recipe. 
image

At a rough count (haven't gone through every recipe in the chain), this gets the battle mastery down to about 3.5k seth and 2.2k osiris shards in total

Targren:
Any chance of revisiting Drawer Upgrade recipes, in light of the new crystal costs? The power reqs push them a good ways deeper down the progression.

MuteTiefling:
There was some discussion on the power already and it's changed again, @Targren.

MuteTiefling:
New addition to Engineering Mastery: DIY Pressure Chamber Kit
image

Also requires advanced compressors now instead of basic
image

This one is still probaly a bit 'cheaper' than the others, but it is heavy on coal coke (lots of steel and compressed iron), as well as iron obviously. Probably OK like this, they don't all need to be the same level of challenge.

MuteTiefling:
This should alleviate the concerns in #4805 by a fair bit. Time will tell, of course, but it makes Battle and Excavation significantly cheaper, while also making Engineering just a touch harder.

thatGuyCalledMax:
Will the induction providers also get the recipe changes to make them similar to the new cells? Or will those be left as is?

thatGuyCalledMax:
Actually nevermind, they have a lot less uses

NielsPilgaard (approved):
Looks good to me, nice changes 👍

PR #4823, opened 2022-05-12
[Expert] Improve RS Storage by skipping the first two tiers of disks

MuteTiefling:
Makes the first two tiers of RS disks uncraftable, shifting all recipes down. So the new 16k storage part, for instance, has the old 1k storage part.

image

Likewise for fluid parts
image

Current trouble: How to relay this to the player. 

I don't want to hide the four removed disks, as people that may still have those would get confused. Maybe just a JEI Description on them?

### DIFF
--- a/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
@@ -30,11 +30,11 @@ onEvent('jei.information', (event) => {
         },
         {
             items: ['refinedstorage:4k_storage_part', 'refinedstorage:1k_storage_part'],
-            text: ['Crafting is disabled in Expert, begin with the 16k Storage Part']
+            text: ['Crafting is disabled in Expert, begin with the 16k Storage Part.']
         },
         {
             items: ['refinedstorage:256k_fluid_storage_part', 'refinedstorage:64k_fluid_storage_part'],
-            text: ['Crafting is disabled in Expert, begin with the 1024k Fluid Storage Part']
+            text: ['Crafting is disabled in Expert, begin with the 1024k Fluid Storage Part.']
         }
     ];
 
--- a/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
@@ -27,6 +27,14 @@ onEvent('jei.information', (event) => {
                 Text.of('Liquid Meat can be obtained by "processing" cows.').color('#6e2a2a'),
                 Text.of('Pink Slime can be harvested from a Pink Wither using a Fluid Laser.').color('#da07e6')
             ]
+        },
+        {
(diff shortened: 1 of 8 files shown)


MuteTiefling:
I don't hate it...

image

kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
```diff
@@ -27,6 +27,14 @@ onEvent('jei.information', (event) => {
...
                 Text.of('Pink Slime can be harvested from a Pink Wither using a Fluid Laser.').color('#da07e6')
             ]
+        },
+        {
+            items: ['refinedstorage:4k_storage_part', 'refinedstorage:1k_storage_part'],
+            text: ['Crafting is disabled in Expert, begin with the 16k Storage Part']
+        },
+        {
+            items: ['refinedstorage:256k_fluid_storage_part', 'refinedstorage:64k_fluid_storage_part'],
+            text: ['Crafting is disabled in Expert, begin with the 1024k Fluid Storage Part']
```
Quezler, line 37: all the other texts seem to end with a period, was this left out so it would fit on 2 just two lines, or forgotten?
```suggestion
        },
        {
            items: ['refinedstorage:4k_storage_part', 'refinedstorage:1k_storage_part'],
            text: ['Crafting is disabled in Expert, begin with the 16k Storage Part.']
        },
        {
            items: ['refinedstorage:256k_fluid_storage_part', 'refinedstorage:64k_fluid_storage_part'],
            text: ['Crafting is disabled in Expert, begin with the 1024k Fluid Storage Part.']
```

thatGuyCalledMax:
The dimension mastery shard might need to be updated to reflect this, if the 16ks are the baseline, since it has 8 16ks per shard currently

kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js
MuteTiefling: Late night hazy brain typo :D

MuteTiefling:
Already included in this @thatGuyCalledMax

NielsPilgaard (changes requested):
Instead of removing the KubeJS crafting items for storage parts, I think we ought to make recipes to convert the disabled storage parts and their packages/assemblies into their new counterparts, so people don't lose stuff when they update. 

Then we can remove the items in 2.0.0 if that ever happens 😄

MuteTiefling:
Storage part conversions. Allows immediate upgrade of existing storage parts. Transfer of items will need to be done via disk manipulator
image
image

Reverted all item removals and associated recipes for the assembly tables, so any in progress crafts will finish.

NielsPilgaard (approved):
Nice, thanks!

Quezler:
> Storage part conversions. Allows immediate upgrade of existing storage parts. Transfer of items will need to be done via disk manipulator
> image

would it not be fairer if the 4k tier converted to 64k?

NielsPilgaard:
Ah yes, good point.

PR #4839, opened 2022-05-14
Patchouli book for masterful machinery

Quezler:
A proof of concept that can render masterful machinery structures using patchouli.

It is currently in its MVP stage, its hard to look at & anchor, code needs more work, etc, etc.

The future of this draft will be decided by the reactions, so I'll wait before sinking any more time into it if at all.

Screen Shot 2022-05-14 at 12 38 19

> `/give @p patchouli:guide_book{"patchouli:book": "patchouli:masterful_machinery"}`

### DIFF
--- a/automation/nodejs/export-mm-to-patchouli.js
+++ b/automation/nodejs/export-mm-to-patchouli.js
@@ -1,0 +1,69 @@
+const fs = require('fs')
+
+const structures_path = 'kubejs/data/masterful_machinery/recipes';
+
+const unused_char = '}' // to free up the 0 with
+
+fs.readdir(structures_path, function (err, files) {
+  files.forEach(file => {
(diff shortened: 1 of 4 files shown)


Quezler:
note: this is not intended as a way to build the structure by hand (you'd be mad if you didn't use the schematic cannon or paste gadget), but rather as a way to debug a structure that failed to form. (progress bar at the top + highlights for wrong blocks)

it of course lags with the bigger structures and the placement anchors could use some finetuning, but in theory it should work.

the code will need to be able to find 1 or 2 unused chars to use in place of the zero and the underscore, so I'll have to check the allowed range. (ascii, or more?)

NielsPilgaard:
I like the concept 👍

PR #4842, opened 2022-05-14
Energizing orb recipe for uraninite blocks

Quezler:
> implements [this suggestion](https://discord.com/channels/195234321836146688/544822403230859283/975076789678403684) by `GiaruDK` on the enigmatica discord.

Screen Shot 2022-05-14 at 19 01 04

the sulfur dust can't be turned into blocks, without a compacting recipe or another way this pull is fairly useless.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/powah/energizing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/powah/energizing.js
@@ -190,6 +190,20 @@ onEvent('recipes', (event) => {
             id: `${id_prefix}uraninite`
         },
         {
+            ingredients: [
+                { tag: 'forge:storage_blocks/uranium' },
+                { tag: 'forge:storage_blocks/uranium' },
+                { tag: 'forge:storage_blocks/sulfur' },
+                { tag: 'forge:storage_blocks/fluorite' }
+            ],
+            energy: 100000,
+            result: {
+                item: 'powah:uraninite_block',
+                count: 2
+            },
+            id: `${id_prefix}uraninite_block`
+        },
+        {
             ingredients: [{ item: 'minecraft:blaze_rod' }],
             energy: 10000000,
             result: {


NielsPilgaard:
I wonder why we have no 3x3 recipes for Sulfur Dust 🤔 I guess it normally takes Sulfur "gems".
I'm a bit on the fence here, the recipe is nice, but I don't want to break conventions by doing dust -> block.

GiaruDK:
> I wonder why we have no 3x3 recipes for Sulfur Dust 🤔 I guess it normally takes Sulfur "gems". I'm a bit on the fence here, the recipe is nice, but I don't want to break conventions by doing dust -> block.

Everything is already thought out, just need increase RF consumption for emulate crusher in sphere, for common recipe uranite we need spend 10k RF, for new recipe want set 100k-120k RF and remaining 10k-30k RFs equals works crusher. Sound Logic ?)

Quezler:
> I wonder why we have no 3x3 recipes for Sulfur Dust 🤔 I guess it normally takes Sulfur "gems". I'm a bit on the fence here, the recipe is nice, but I don't want to break conventions by doing dust -> block.

strange enough sulfur dust can be melted down by tinkers & cast into a block, which in turn returns other sulfur crystals :D

NielsPilgaard:
> just need increase RF consumption for emulate crusher in sphere, for common recipe uranite we need spend 10k RF, for new recipe want set 100k-120k RF and remaining 10k-30k RFs equals works crusher. Sound Logic ?)

I.. don't follow to be honest.

GiaruDK:
Sry for my english bad. I mean powah sphere makes this recipe without crushing sulphur and fluorite in the crafting chain, but use a little more RF, as if imitating a crusher work

MuteTiefling:
Not sure this is needed, honestly. Uraninite burns slowly enough that you already barely need to craft it.

Plus, as niller pointed out, it goes against convention to have a block of dust, and removing that step isn't just an energy savings. It reduces the logistics required for the whole process.

GiaruDK:
Next suggestion add recipe Spirited crystals block with use uranite for craft ) And posibly fully changed Spirited Crystals recipe, replace uranium on uranite.

PR #4844, opened 2022-05-14
Allow upgrading black hole units/tanks

Quezler:
smithing table seems to respect the nbt, "unfortunately" it doesn't return or refund the previous tier of frame:
Screen Shot 2022-05-14 at 19 46 34

since expert mode overrides the 1st tier to not have plastic/enderpearls in the recipe it cannot be upgraded:
Screen Shot 2022-05-14 at 19 57 41

because all the tiers after it require 3 plastic and 3 enderpearls, a notable increase in cost & gating:
Screen Shot 2022-05-14 at 19 57 51

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
@@ -3,10 +3,34 @@ onEvent('recipes', (event) => {
         {
             input1: Item.of('aiotbotania:terra_aiot', '{Damage:0}').weakNBT(),
             input2: 'kubejs:quintuple_alfsteel_ingot',
-            output: 'aiotbotania:alfsteel_aiot'
+            output: 'aiotbotania:alfsteel_aiot',
+            id: `aiotbotania:alfsteel_aiot_smithing`
         }
     ];
+
+    const black_hole_types = ['tank', 'unit'];
+    const black_hole_tiers = ['common', 'pity', 'simple', 'advanced', 'supreme'];
+
+    black_hole_types.forEach(type => {
+        black_hole_tiers.forEach((tier, index) => {
+
+            // first tier has no previous tier to upgrade from
+            if(index == 0) return; 
+
+            // no 3 free plastic & enderpearls for expert mode
+            if(index == 1 && global.isExpertMode) return; 
+
+            const prev = black_hole_tiers[index - 1];
+            recipes.push({
+                input1: `industrialforegoing:${prev}_black_hole_${type}`,
+                input2: `industrialforegoing:machine_frame_${tier}`,
+                output: `industrialforegoing:${tier}_black_hole_${type}`,
+                id: `industrialforegoing:${tier}_black_hole_${type}_smithing`
+            });
+        });
+    })
+
     recipes.forEach((recipe) => {
-        event.smithing(recipe.output, recipe.input1, recipe.input2);
+        event.smithing(recipe.output, recipe.input1, recipe.input2).id(recipe.id);
     });
 });
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
@@ -1,10 +1,12 @@
 onEvent('recipes', (event) => {
+    const id_prefix = 'enigmatica:base/smithing/';
+
     const recipes = [
         {
             input1: Item.of('aiotbotania:terra_aiot', '{Damage:0}').weakNBT(),
             input2: 'kubejs:quintuple_alfsteel_ingot',
             output: 'aiotbotania:alfsteel_aiot',
-            id: `aiotbotania:alfsteel_aiot_smithing`
+            id: `${id_prefix}alfsteel_aiot`
         }
     ];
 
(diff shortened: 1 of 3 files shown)


MuteTiefling:
losing materials when upgrading is on brand at this point, so need to worry about that

Quezler:
The common tier now uses the same other 8 items as the other tiers, instead of being the odd one out:

Screen Shot 2022-05-14 at 20 41 24
Screen Shot 2022-05-14 at 20 41 41

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
```diff
@@ -129,19 +129,25 @@ onEvent('recipes', (event) => {
         },
         {
             output: 'industrialforegoing:common_black_hole_tank',
-            pattern: ['AAA', 'A A', 'ABA'],
+            pattern: ['AAA', 'BCB', 'DED'],
             key: {
-                A: '#forge:sheetmetals/aluminum',
-                B: 'thermal:fluid_cell_frame'
```
MuteTiefling, line 135: Would prefer to keep the fluid cell frame here, just for thematic reasons.

Quezler: my bad, copy pasting error 👍

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
```diff
@@ -129,19 +129,25 @@ onEvent('recipes', (event) => {
...
+                A: 'industrialforegoing:plastic',
+                B: 'minecraft:ender_eye',
+                C: 'minecraft:ender_pearl',
+                D: 'minecraft:bucket',
+                E: 'thermal:machine_frame'
```
NielsPilgaard: I think this should be `thermal:fluid_cell_frame`?

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
MuteTiefling: happens :D

NielsPilgaard (changes requested):

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
NielsPilgaard: Woops, I was looking at old code, my bad!

NielsPilgaard (approved):

NielsPilgaard:
I forgot to mention, the smithing recipe ids could be improved by using the Enigmatica-specific id format (`enigmatica:base/....`)
I always forget how it goes exactly ^^'

Quezler:
all the other smithing recipes seemed to use the `<output>_smithing` id-ing pattern, so i followed in those footsteps 👀

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
```diff
@@ -3,10 +3,31 @@ onEvent('recipes', (event) => {
...
+                input2: `industrialforegoing:machine_frame_${tier}`,
+                output: `industrialforegoing:${tier}_black_hole_${type}`,
+                id: `industrialforegoing:${tier}_black_hole_${type}_smithing`
+            });
+        });
+    })
+
     recipes.forEach((recipe) => {
-        event.smithing(recipe.output, recipe.input1, recipe.input2);
+        event.smithing(recipe.output, recipe.input1, recipe.input2).id(recipe.id);
```
MuteTiefling, line 31: For the ID's, I'd like to keep the same format as other places.

At the top of the file, there should be an id_prefix: `const id_prefix = 'enigmatica:expert/smithing/';`

and the id in the recipe would then follow the format of id: ${id_prefix} `${id_prefix}{tier}_black_hole_${type}_smithing`

MuteTiefling (changes requested):

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
Quezler: Since all the other smithing recipes don't use id prefixes either, mind if i request this to be merged as-is, and then a separate pr to prefix all the smithing recipes properly?

Quezler: and if the prefix has `/smithing/` in it, will it still need to end with `_smithing`?

TemperedFool: Imo, if anything you'd want it to be like `upgrade` or something.

MuteTiefling:
> all the other smithing recipes seemed to use the `<output>_smithing` id-ing pattern, so i followed in those footsteps 👀

That's kinda the old way of doing things. This script just never got updated as there were never any changes in it since that structure fell into place.

kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
TemperedFool: Also here's a question - is there any reason we don't just let them jump up multiple tiers like this? Seems fine to go from Pity -> Supreme...

MuteTiefling (approved):

NielsPilgaard (approved):

PR #4845, opened 2022-05-15
[Expert] Better progression for cloche, insolator, and botany pots

### LINKED ISSUE #4836: Garden cloche not worth the effort
### Describe your suggestion

The garden cloche is late game and yet it's really slow, a lot slower than the phytogenic insulator which is early game, really disappointing.

Unless the garden cloche is sped up dramatically there's no point using it.

MuteTiefling:
Cloche now upgrades the Hydroponic Bed
image

Insolator requires higher tier tech
image

Hopper Botany Pot now requires higher tier magic
image

Better resolves #4836

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/create/mechanical_crafting.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/create/mechanical_crafting.js
@@ -577,7 +577,7 @@ onEvent('recipes', (event) => {
                 A: 'thermal:lumium_glass',
                 B: 'immersiveengineering:electron_tube',
                 C: '#forge:treated_wood',
-                D: '#botanypots:botany_pots/simple',
+                D: 'industrialforegoing:hydroponic_bed',
                 E: 'immersiveengineering:component_steel',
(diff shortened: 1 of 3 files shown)


User-12691296:
I also believe the basic botany pot should be less gated, 

While it can be a lag machine (and thus should be gated behind something that takes a decently long time to make, so it can't be spammed), 

For something very much manual, that just saves a step in, and allows you to farm more stuff than, basic manual farming, it is unnecessarily gated. It has the usefulness of an earlygame item, and yet is gated in the lategame (tier 3 magic, tier 3 tech). 

Gating the botany pot behind rich soil (which takes a while to make, so it can't be spammed as easily), overgrowth seeds, and atum fertile soil (to move it slightly into the early game), should provide the right level of gating I think is needed, as now the hopper botany pot recipe will cover all the old cost of the regular botany pot. (Weak blood shards, elven gateway, dissolution chamber, godshards, 9 million FE)

MuteTiefling:
The basic botany pot is fine. I'd rather promote other forms of farming.

NielsPilgaard (approved):

PR #4852, opened 2022-05-15
Add kubejs helper function `lowerTiers`

Quezler:
Adds a helper function so we don't need to see either:
```
let lowerTiers = [],
    i = 0,
    j = powahTiers.indexOf(tier);

while (i < j) {
    lowerTiers.push(powahTiers[i]);
    i++;
}
```
Or:
```
powahTiers.slice(0, index);
```
But instead:
```
lowerTiers(<tiers>, <tier>);
```
Since the first method looks confusing & slice only a little less.

For future helper functions i have stashed 4 lines of reference on slice.

Also leaving my initial attempt here: (which i abandoned since its uncommon, and i don't know if kubejs likes it much)

```
Object.defineProperty(Array.prototype, 'less_than', {
    value: function(index) {
        return this.slice(0, index);
    }, configurable: true
});
  
Object.defineProperty(Array.prototype, 'less_than_equals', {
    value: function(index) {
        return this.slice(0, index + 1);
    }, configurable: true
});
  
Object.defineProperty(Array.prototype, 'greater_than', {
    value: function(index) {
        return this.slice(index);
    }, configurable: true
});
  
Object.defineProperty(Array.prototype, 'greater_than_equals', {
    value: function(index) {
        return this.slice(index + 1);
    }, configurable: true
});
```

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/minecraft/smithing.js
@@ -14,11 +14,9 @@ onEvent('recipes', (event) => {
     const black_hole_tiers = ['common', 'pity', 'simple', 'advanced', 'supreme'];
 
     black_hole_types.forEach(type => {
-        black_hole_tiers.forEach((tier, index) => {
+        black_hole_tiers.forEach(tier => {
 
-            let lower_tiers = black_hole_tiers.slice(0, index);
-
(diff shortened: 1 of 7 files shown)


NielsPilgaard (approved):
Looks good 👍 did you validate that most recipes changed still work?

Quezler:
I verified that the black hole smithing recipes accepted the lower tiers, as well as the powah magmator, i didn't walk through all the powah recipes since the code is pretty much identical 🤔

PR #4856, opened 2022-05-16
Putting recipe id overrides in the enigmatica namespace

Quezler:
During my quest to eliminate `:kjs_` recipe ids i ran into inconsistencies, one of them being enigmatica 6 overriding mod recipes (possibly of a different type altogether, e.g. dissolution chamber) by intentionally putting a recipe under the same id.

It was rather confusing back when i wanted to edit the common black hole unit/tank recipe since the (apparently already customized) id was set to `industrialforegoing:common_black_hole_tank`, implying that the recipe originates at that mod, which makes it more tricky to lookup amongst the kubejs files.

This pull request serves as a showcase of the option to specify the "old" recipe id where the recipe override is defined, accompanied by the "new" id that is in the proper enigmatica namespace. (and not in the kubejs file that removes recipes)

More information regarding my intentions to cleanup recipe ids can be found in [this](https://github.com/Quezler/Enigmatica6/blob/ci/nameless_crackdown/.github/CONTRIBUTING.md) work in progress markdown document.

Screen Shot 2022-05-16 at 16 43 52

(this is not intended to be merged since its only one file as an example, for discussion before deploying on a larger scale)

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/industrialforegoing/shaped.js
@@ -2,7 +2,7 @@ onEvent('recipes', (event) => {
     if (global.isExpertMode == false) {
         return;
     }
-    const id_prefix = 'enigmatica:expert/industrialforegoing/shaped/';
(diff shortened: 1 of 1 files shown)


NielsPilgaard:
While this would be super nice, I think it would cause issues with things like Patchouli that rely on recipe ids to display recipes.

PR #4860, opened 2022-05-17
Show notice for expired `//todo remove in #.#.#`’s

Quezler:
This is different territory than my other CI related pull requests, basically scripts that will warn about inconsistencies, which after their initial run their usefulness greatly reduces, for example this pull request now shows this notice:
Screen Shot 2022-05-17 at 14 47 24

But after all those files flagged for removal have been removed, it will obviously not output a list of files anymore:
Screen Shot 2022-05-17 at 14 47 45

There are a few more similar kubejs file check scripts in my pipeline, but lets first see how this one is received. 🤔 

(i have plans to make it modular & use npm to import packages to clean it up a bit, but that's for when we have multiples)

### DIFF
--- a/.github/actions/ci/kubejs-todo-remove.js
+++ b/.github/actions/ci/kubejs-todo-remove.js
@@ -1,0 +1,68 @@
+const fs = require('fs');
+const { execSync } = require("child_process");
+
+//
+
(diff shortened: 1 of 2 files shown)


NielsPilgaard (approved):
Good stuff :D

PR #4869, opened 2022-05-20
Scrape recipe ids mentioned by patchouli during ci

Quezler:
a ci script that will look through the `patchouli` pages of all mods to find recipe ids that could throw a fuss when removed.

(further cleanup if having a list is deemed useful, currently it is merely in a state where it can output the desired list)

i don't think i have missed any page types (heck, my script would throw an error if i did), but i might have flagged some types as "contains no (important) recipe id" (aka: blacklisted) by mistake, and likewise whitelisted some types that "might" not crash if their recipe id is missing (aka: properly implemented), but it will at least spit out a list of "handle with care" ids:
Screen Shot 2022-05-20 at 11 27 20

in case you want to see the list, you can either download the artifact or preview it here:
Screen Shot 2022-05-20 at 11 32 01

and to the php haters: this is simply the language **i** can get shit done in, large scale nodejs/java/etc isn't my thing. 🙃


NielsPilgaard:
To be honest I'd rather not have this as a part of the CI, for the reasons that I can't really read it, and I don't see a need for it at the moment.

PR #4870, opened 2022-05-20
List mods that have opted out of curseforge 3rd party downloads

Quezler:
This api simply showed up when i tried to check on google if this was the key i was looking for:
Screen Shot 2022-05-20 at 12 40 25

The `xmdhs.top` site seems to be host/proxy this repository: https://github.com/xmdhs/cursemodownload

Obviously since this api key does not belong us, nor did i seek out him ^ yet this should **not be merged**, this is a proof-of-concept script that checks which mods in our `minecraftinstance.json` have opted in, out, or have yet to decide:
Screen Shot 2022-05-20 at 12 38 17

So far i have only ran it a few times myself to develop the script, but once we sort things out we can get a list like this:
Screen Shot 2022-05-20 at 12 39 08

I don't know yet how the api/multimc changes will affect enigmatica, so far instancesync seems unaffected (for now) & multimc currently suggests using a curseforge certified launcher to install and then import that installation into it, but if e6 goes a different/temporary route it is probably good to know which mods have chosen to opt out. ^-^ (manual download?)


NielsPilgaard:
Very interesting, I'd be up for having this as a separate action if we had *actual* access to the API :P

Quezler:
hit me up once you have an api key 👌

PR #4872, opened 2022-05-21
Include downloadUrl in future manifest files

Quezler:
The serverstarter is currently broken since the api behind it is down, but since instancesync still worked i decided to dig a little deeper, and noticed that the starter only used the api to resolve the download links, but they are already present in the `minecraftinstance.json` niller uses to generate the server files, this pull requests adds an extra property to the json.

Naturally this means only manifests created with this additional property will benefit from it, and not any normal curse ones.

It (obviously) requires changes to the bundled server starter: https://github.com/EnigmaticaModpacks/ServerStarter/pull/3


NielsPilgaard (approved):
Good stuff, thanks for doing it!

PR #4896, opened 2022-05-25
[Expert] Buff the IP Coker

### LINKED ISSUE #4878: [E6E] Buff Immersive Petroleum Coker Unit multiblock
### Describe your suggestion

The Coker should be buffed to provide the player with an alternative way to supply their HOP Graphite Dust needs rather than the only viable option for it is to spam Pyrolyzers as even with it maxed out can still be too slow.

As the Coker is currently gated by Heavy Machinery Schematics, it will provide players another incentive to get the schematics as well as reduce the need for Pyrolyzer spamming. 

In its current form it is pretty underwhelming as it only converts 2 Bitumen to 1 Petcoke in about 3 seconds while you can get much more for how much space it takes with a couple Pyrolyzers.

Buff ideas might be to add block recipes to it like the Pyrolyzer already has and/or increase overall speed for its current recipe but increase power consumption so the player would still need to commit a fair bit of their power network if they want to run it constantly.

Pyrolyzer would still have a use since it's the only viable way to produce Coal Coke which in most cases Petcoke does not substitute for.

MuteTiefling:
Massive speed increase at the cost of extra energy. 614k energy over 30 seconds in base per recipe, vs 3.8 million over 1.5 seconds per recipe proposed.

Resolves #4878


kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/immersivepetroleum/coker.js
```diff
@@ -0,0 +1,26 @@
+onEvent('recipes', (event) => {
+    if (global.isExpertMode == false) {
+        return;
+    }
+    const id_prefix = 'enigmatica:expert/immersivepetroleum/coker/';
+
+    recipes = [
```
NielsPilgaard, line 7: missing `const` :o

NielsPilgaard (changes requested):

kubejs/server_scripts/enigmatica/kubejs/expert/recipetypes/immersivepetroleum/coker.js
NielsPilgaard: Noes, `recipes` has no `var`, `let` or `const`, unless recipes is defined somewhere else this'll fail

NielsPilgaard (changes requested):

NielsPilgaard (approved):
Nice :D approved!

Issue #4903, opened 2022-05-26
E6E Iesnium nugget and block recipes missing


Gorfeywog:
### Modpack Version

1.3.0

### Describe your issue.

The recipes to convert between iesnium nuggets, ingots, and blocks seem to have gotten lost during the update. They should probably be readded.
For instance, ingot to nuggets:
![image](https://user-images.githubusercontent.com/39035228/170571257-fde060d1-67b9-4311-9091-ac4b20a3642e.png)
Or block to ingots:
![image](https://user-images.githubusercontent.com/39035228/170571373-cb152386-4961-4622-b5b3-70d3c88847ee.png)
Nuggets to ingot and ingots to block are also gone. Recipes outside of the crafting table still seem to be there.

### Have you modified the modpack?

No

### Did the issue happen in singleplayer or on a server?

Server

MuteTiefling:
This will be fixed as part of a pr in currently working on.

Issue #4857, opened 2022-05-16
Missing Multiservo Press 3x3 Compacting Recipes


FallenRosemary:
### Modpack Version

1.2.1

### Describe your issue.

There seems to be a bunch of 3x3 compacting recipes missing from the multiservo press for resource storage blocks like source gems and lead
discovered this after trying to get ore processing running and using multiservos to compact everything down into blocks

### Have you modified the modpack?

Yes

### User Modifications

added optifine, buddycards and metal barrels

### Did the issue happen in singleplayer or on a server?

Singleplayer

TemperedFool:
I believe this should be resolved by #4724; are there actually block recipes missing? This adds nugget -> ingot recipes.

FallenRosemary:
most of the storage block recipes are missing, yep

NielsPilgaard:
Can confirm, they're still missing sadly.

Sampppppppppppppppppppppppp:
immersive engineering metal press and multi servo press have inconsistant/conflicting recipes. Servo dies can't be crafted, IE unpacking/packing dies can't be used on the metal press

MuteTiefling:
> immersive engineering metal press and multi servo press have inconsistant/conflicting recipes. Servo dies can't be crafted, IE unpacking/packing dies can't be used on the metal press

These have been unified to the extent that they can be. Thermal dies were removed (with the exception of numismatic) in favor of the IE molds. Thermal machines use the IE molds instead. 
The unpacking is surprising, however, since those should be automatic recipes.

Sampppppppppppppppppppppppp:
> 

When I tried to use the IE metal press with the IE unpacking mold with melons/pumpkins, it did not work. However, it did work when using the Thermal unpacking die on the metal press as the JEI recipe shows. In kubejs\server_scripts\enigmatica\kubejs\constants\dies.js I added "unpacking" to the list of IE dies and that seemed to fix the metal press working with the IE unpacking die.

MuteTiefling:
alright, I think I see part of the issue. The IE machine doesn't like our tags. It's picking one randomly out of the tag and going with that.

Still doesn't explain why neither machine is doing auto block > ingot recipes based on standard tags... but that's an easy fix to. Will work on this this evening.

PR #4904, opened 2022-05-27
Many fixes for packing and unpacking with Thermal and Immersive Engineering

MuteTiefling:
Primarily addresses #4857

Removes thermal packing/unpacking dies from the tags, as the IE metal press can't seem to deal with this and just picks one from the tag to work with.

Re-works nugget > ingot > block packing and block > ingot > nugget unpacking. 

Re-adds some missing 3x3 recipes for doing the same, notably iesnium (#4903), pig iron, and refined glowstone/obsidian.

Adds 2x2 and 3x3 packing/unpacking support for a lot of blocks. There's probably a fair few more that would be nice to have. Can add these as they come up.

Adds Coin and (expert only) Processor patterns to the IE Metal Press since it seems IE will work with the Thermal Dies afterall.


NielsPilgaard (approved):
Good stuff, nice work :)

PR #4910, opened 2022-05-28
`ci` download mods once

Quezler:
Download the mods just once instead of for both normal & expert, kinder on the curseforge cdn: 
Screen Shot 2022-05-28 at 09 52 43
(this does not cache/lock them to a hash of minecraftinstance.json between runs, in order to quickly spot download issues)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,8 @@ on:
   #         - 'master'
 
 jobs:
-  mods:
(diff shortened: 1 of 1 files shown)


Quezler:
Originally planned for after the #4869 merge so it didn't have to download thrice instead of twice, but still nice to have now.

NielsPilgaard (approved):
Nice change :D

PR #4935, opened 2022-05-29
`ci` list infinite fluids

Quezler:
Shows a (count &) list of all fluids that have not been blacklisted for the hose pulley, there even are 1-2 in there that should:
Screen Shot 2022-05-29 at 20 59 57


NielsPilgaard:
Looks good to me 👌

PR #4955, opened 2022-06-04
Mention that the refined silk touch doesn't drop spawners

Quezler:
Stuff like grass blocks works just fine, but it destroys any spawners, i would have rather known about this **bug** beforehand:

### DIFF
--- a/kubejs/client_scripts/item_modifiers/tooltips.js
+++ b/kubejs/client_scripts/item_modifiers/tooltips.js
@@ -171,6 +171,10 @@ onEvent('item.tooltip', (event) => {
         {
             items: ['kubejs:monster_mash'],
             text: [Text.of(`...It caught on in a flash...`).red()]
+        },
+        {
+            items: ['refinedstorage:silk_touch_upgrade'],
+            text: [Text.of('Does not work on spawners.').red()]
         }
     ];
 


NielsPilgaard:
Did you also report the bug? Perhaps to Darkere or the Apotheosis tracker? :)

Quezler:
This is an bug report & pull request in one ^-^

Quezler:
Though honestly i would just suggest to remove the [silk touch requirement](https://github.com/EnigmaticaModpacks/Enigmatica6/blob/master/config/apotheosis/spawner.cfg#L11) in order to harvest spawners entirely:
- cardboard is cheap & early (in both modes), so the user is already able to move spawners early on.
- cardboxed spawners only say that they're a spawner and not which mod, which makes it a pain to sort through.
- not everyone's favorite tool will have silk touch on it. (e.g. atomic disassembler, or a tinker pick which loots)

It would be far more convenient for everyone if spawners simply always drop when mined by a player, avoids the pain.
It would not even be op since you can carry them in cardboard anyways, only difference is seeing the mob & stats.

Darkere:
Quick look at the code, apotheosis overwrites a method that can only be called by real players.

MuteTiefling:
I personally like the requirement. Sure your can move them with cardboard, but using more specialized tools grants an advantage.

Same with mining in general. Sure you can fortune things as you mine, but your inventory fills faster than if you simply silk touch everything.

Quezler:
But this is just needlessly cruel towards players, and they don't (seem to) stack anyways, i see no good reason to suffer 🤔

Quezler:
> Quick look at the code, apotheosis overwrites a method that can only be called by real players.

I haven't tested it with drones or deployers yet, but i did notice that bit of code yes, and automated unboxing might only be possible with ID, people are condemned to carry around a silk touch pickaxe just for spawners if they want to avoid cardboard.

MuteTiefling:
You make it sound like such a punishment to carry a silk touch tool... Keep it in a backpack? Or just use it as your primary tool as there are more benefits to using silk all the time than fortune.

Quezler:
No, but by the time you can roughly obtain that enchantment you will already have come across tons of spawners that you either had to decide to cardboard & bring along (and get in subsequent sorting horror) or just ignore & leave em there.

NielsPilgaard:
I don't see the harm in merging this.

I also think it would be too easy to obtain spawners if the silk touch requirement wasn't there.

PR #4973, opened 2022-06-08
`tag` pneumaticcraft:pressurable

Quezler:
Somehow this does not show up in `./kubejs/exported/tags/items.txt`, so there's probably still a mistake somewhere. 🤔 

I could really use this tag for computercraft, being able to export items that can be charged in a charging station without having to hardcode this same list inside lua.

Logs for the ci run can be found [here](https://github.com/Quezler/Enigmatica6/actions/runs/2460895345), but i could not spot any kubejs errors related to why this tag might not be working.

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/storage/air.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/storage/air.js
@@ -13,5 +13,9 @@ onEvent('item.tags', (event) => {
     'pneumaticcraft:jackhammer'
   ];
 
-  event.get('pneumaticcraft:pressurable').add(items);
+  var tags = ['forge:storage', 'forge:storage/air'];
+
+  tags.forEach((tag) => {
+      event.get(tag).add(items);
+  });
 });
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/storage/air.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/storage/air.js
@@ -1,7 +1,7 @@
 onEvent('item.tags', (event) => {
   var items = [
     /pneumaticcraft:(reinforced_)?air_canister/, // 10 & 20 bar canisters
-    /pneumaticcraft:(\w*_)?drone$/,              // all 5 craftable drones
+    /pneumaticcraft:(\w+_)?drone$/,              // all 5 craftable drones
 
     /pneumaticcraft:pneumatic_(wrench|helmet|chestplate|leggings|boots)/,
 
@@ -11,7 +11,7 @@ onEvent('item.tags', (event) => {
     'pneumaticcraft:amadron_tablet',
     'pneumaticcraft:minigun',
     'pneumaticcraft:camo_applicator',
-    'pneumaticcraft:jackhammer'
+    'pneumaticcraft:jackhammer',
   ];
 
   var tags = ['forge:storage', 'forge:storage/air'];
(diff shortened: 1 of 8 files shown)


Quezler:
Fine by me, will move it there once i figure out why it isn't working as-is 🤔

Quezler:
looks like it works, but i guess i will have to fix the regexes since more stuff made it in than anticipated:
Screen Shot 2022-06-08 at 17 09 45

Quezler:
I was only looking `#pressure` in jei at the time and noticed some of the pressurized items shared the `pneumaticcraft:pressure_` prefix, didn't test it at the time since i was working on something else and didn't want to reboot that instance just yet, which is why i had set this to draft mode since it wasn't properly tested/working, a harmless oversight.

Quezler:
Decided to hijack this in order to include ci that can compare changes in tags: 
Screen Shot 2022-06-09 at 18 07 27

Clicking it will give a nice git-like diff of the changes:
Screen Shot 2022-06-09 at 18 07 39

Quezler:
when searching `#pressure` in jei:
Screen Shot 2022-06-09 at 18 21 38

when searching `$storage/air` in jei:
Screen Shot 2022-06-09 at 18 22 01

new tags under `forge:storage/air`:
Screen Shot 2022-06-09 at 18 27 42

new tags under `forge:storage`:
Screen Shot 2022-06-09 at 18 28 35

Quezler:
@NielsPilgaard ready for review 🍿

kubejs/server_scripts/enigmatica/kubejs/base/tags/items/forge/storage/air.js
```diff
@@ -0,0 +1,22 @@
...
+    'pneumaticcraft:vortex_cannon',
+    'pneumaticcraft:manometer',
+    'pneumaticcraft:logistics_configurator',
+    'pneumaticcraft:amadron_tablet',
+    'pneumaticcraft:minigun',
+    'pneumaticcraft:camo_applicator',
+    'pneumaticcraft:jackhammer',
+  ];
+
+  var tags = ['forge:storage', 'forge:storage/air'];
```
NielsPilgaard, line 17: I recommend you use `const` here.

NielsPilgaard (changes requested):
I think the tags are a fine addition, I think the CI additions are cool but overkill to be honest, I'd rather not have those.

Quezler:
> I think the tags are a fine addition, I think the CI additions are cool but overkill to be honest, I'd rather not have those.

After working on getting ci to work for pull requests + being able to compare against the pull request i did not expect this.

My other (accepted) ci pull requests were all about making computed data accessible (aka: stuff not visible in the kubejs json files alone), like: counting recipes, finding out which recipes use randomized ids, finding any startup errors, etc.

Therefore it seemed reasonable to me that showing a diff for changed tags made sense, sure they are configured inside the kubejs json files, but you won't know the result until after kubejs has had a run and exported the final tags.

For string arrays of tags the value is indeed minimal since it is 1:1, but any time regex item matches are involved it serves as a valuable check to see if all the items you want it to match are matching, as well as detecting ones that should not make it.

It also serves as a sort of test, if in the future the tag code gets refactored (with the intention of updating just the code, not changing anything) this will be able to verify if nothing has truly changed as far as kubejs is concerned.

Are you absolutely sure you want me to strip out the entire ci part? not to mention future avenues of comparing branches 🤔

NielsPilgaard (approved):

NielsPilgaard:
The thing is, I don't like to have code running in our actions that I don't fully understand. I don't have the time to learn, I barely have the time to read it. I cannot maintain this, nor do I want to.

That being said, sure, I'll merge this one, I can see the benefit of it.

Quezler:
i'll add some more comments to this tomorrow if you'd like 🤔 

also the ci is not part of the deployment process, if something breaks and for whatever reason i am not around, it can be uncommented/discarded just fine 😇

NielsPilgaard:
No thank you. I rarely have more than 20 minutes per day for Minecraft development, I do not want to spend it reading code comments for CI, no offense :P

PR #4991, opened 2022-06-12
Delete all the files flagged for deletion before 1.3.0

Quezler:
As discussed in discord, this pull request deletes the script that checks for `todo: delete in`, as well as the files it flagged.

In preparation for some future stuff this repurposes the kubejs job for all the logs parsed after both servers have ran.

The only actual addition is `- run: sleep 1m`, a temporary measure until i properly figure out how to make node kill java.

> \+ maintenance label
> \+ review from niller


NielsPilgaard:
The files shouldn't be removed until we make a breaking change, and update to version 2.0.0 :)

Quezler:
> The files shouldn't be removed until we make a breaking change, and update to version 2.0.0 :)

but those were files marked for "delete in 0.6.0" and we're @ 1.3.0 now 👀

Darkere:
0.5 -> 1.0 was not a breaking update. so 0.6 would be the same as 2.0.

NielsPilgaard:
Yeah sorry for the confusion in versioning / when a version change is breaking, I know it makes very little sense, that's why I changed it :P

Quezler:
I wrote this in js (and a previous deletion detector in php) just because of the expired versions to begin with >.<

I propose using `// todo: delete @ breaking change` instead of listing a version number 🤔

NielsPilgaard (approved):

NielsPilgaard:
Sounds good, thanks for the cleanup 👍

PR #4993, opened 2022-06-13
`ci` detect global constants and functions

Quezler:
Due to the way kubejs essentially glues all javascript files together via rhino, its hard to keep track of all global variables.

This pull request introduces a ci script that will (try to) find all the global variables added by enigmatica 6's scripts:
Screen Shot 2022-06-13 at 07 10 59

Some interesting finds so far:
- a global `i` variable originating from stonecuttables (and base `rftools.js` piggybacking off of it?)
- a global `cuttingRecipe` function in base & expert @ `~/recipetypes/farmersdelight/cutting.js`

Ideally instead of the two pre/post hooks it would modify kubejs directly in order to grab any (new) consts after each file gets loaded [here](https://github.com/KubeJS-Mods/KubeJS/blob/fa40591d5476dc8407f7bb69acadbec7aab7da65/common/src/main/java/dev/latvian/kubejs/script/ScriptManager.java#L134) & [here](https://github.com/KubeJS-Mods/KubeJS/blob/fa40591d5476dc8407f7bb69acadbec7aab7da65/common/src/main/java/dev/latvian/kubejs/script/ScriptFile.java#L31) (my simple script does not guarantee a file reference, so a few slip through the cracks)

This script has a similar functionality to the randomized recipe names locator, our constants/functions don't follow a specific naming format (some are lowercase with hyphens, others have capitalized words, etc), so conflicts can arise.

The full list of the above screenshot can be viewed here: 
https://github.com/Quezler/Enigmatica6/runs/6855484231?check_suite_focus=true#step:10:557

The two kubejs files (pre & post) are debatable, them being there causes no harm & it was easy to test that way, i could create them only during ci if that is preferred, so let me know 🙂 (also, client/startup script globals are currently ignored)


NielsPilgaard:
Very nice, good work!

> The two kubejs files (pre & post) are debatable, them being there causes no harm & it was easy to test that way, i could create them only during ci if that is preferred, so let me know 🙂 (also, client/startup script globals are currently ignored)

I'd prefer them created during ci :)

Quezler:
> @Quezler requested a review from @NielsPilgaard

NielsPilgaard (approved):

PR #4994, opened 2022-06-13
Allow waystones to be recycled by BM ARC scissors

Quezler:
Screen Shot 2022-06-13 at 10 28 53

You spawn with one, but most importantly (& like other valuables) you can find a lot of them while exploring.

This is fine for when you're playing solo, but when you play in a team you have to tpa every teammate to a freshly placed waystone when you place/move it so they can rightclick it, sharestones however use color channels.

With this pull request you can recycle all the waystone mod blocks that contain a purple warp stone.

Recycling is gated after the point where you can make warp stones natively. (tier 3 blood altar)

I have considered the sawmill (like the dimensional shards) but it felt unsafe to stick stone in there, as well as the create crushing wheels but that felt way too early.

**TLDR** makes waystones worth collecting when playing in a team ^-^

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/waystones/waystone.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/tags/items/waystones/waystone.js
@@ -1,0 +1,3 @@
+onEvent('item.tags', (event) => {
+    event.get('waystones:waystone').add(/waystones:(\w+_)?waystone$/);
+});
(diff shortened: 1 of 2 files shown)


Quezler:
ps, used `waystones:waystone` without the `s` since 1) twice the same would look weird 2) it follows the tags from the mod:

Screen Shot 2022-06-13 at 11 22 14

PR #5002, opened 2022-06-14
`ci` detect global constants and functions 2: electric boogaloo

Quezler:
This pull requests improves upon #4993 by hooking into kubejs directly, then it can sense global vars after each file loads.

For this a custom kubejs jarfile is required, instead of maintaining a separate fork (less control/overview) i have opted for a patch applied to the official kubejs repo instead. (so all the changes can be reviewed right here)

Patches have a weakness though: merge conflicts, but the 1.16 branch of kubejs has recently been renamed `eol`, so i am not that worried about any (breaking) changes at all.

Here is the new output, no more guessing by searching files for substrings, this should find everything:
Screen Shot 2022-06-14 at 19 19 47
[all output of that step](https://github.com/Quezler/Enigmatica6/runs/6885514008?check_suite_focus=true#step:10:8)


NielsPilgaard (commented):
Interesting changes, nice work. I think re-compiling the patched KubeJS every run is a bit much though, perhaps you could compile it, store it somewhere and just download?

Quezler:
it gets cached at line 159, only when the hash of the kubejs.patch file changes does it recompile 😊 (if its in the cache that job finishes before instancesync even finishes downloading)

NielsPilgaard:
Ahh okay, I missed that, looks good then 👍

PR #5005, opened 2022-06-15
[Expert] Reduce mana/blood cost of dislocation runes

MuteTiefling:
These are required to get the most out of the recent pre-WoS LP automation.

Doesn't change their gating at all, but by moving to an earlier rune and cutting the mana cost to craft, they'll be more accessible, adding value to actually setting up that automation.

Also changes BM quests to give 4 displacement runes instead of 2 speed runes. Ultimately, I think these are a more appropriate reward.


NielsPilgaard (approved):
Looks good to me!

PR #4996, opened 2022-06-13
Remove the catnip larva in favor of the kitten larva (#4937)

Quezler:
Removing the catnip_larva should have no side-effects, it crafted into a default bee anyways.

PR #5009, opened 2022-06-15
`ci` fix kubejs error detection script (`catnip_larva` slipped through the cracks)

Quezler:
Unfortunately niller beat me to it, #4996 got merged earlier in the day than expected, but here it is regardless:

KubeJS only outputs the added/modified/removed recipes while debug mode is on, unfortunately that also caused every warning to throw an err stacktrace which my detection script had trouble ignoring.

Now that patch files have gained a foothold in the ci, it can be used to get the recipe outputs without enabling debuginfo.

This means the ugly `ModLauncher running` hook can go, and allow the nodejs script to be simplified & cleaned.

Here is the output of it running on this branch, the builds for the pull request will likely succeed because of the merge:
Screen Shot 2022-06-15 at 14 09 09

### DIFF
--- a/.github/actions/ci/log-kubejs-warnings.js
+++ b/.github/actions/ci/log-kubejs-warnings.js
@@ -24,4 +24,5 @@ warnings.forEach(warning => {
     console.log(`::warning::${warning}`);
 });
 
-process.exit(warnings.length === 0 ? 0 : 1);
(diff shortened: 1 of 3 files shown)


NielsPilgaard (approved):
Good stuff, thank you kindly 👍

PR #5017, opened 2022-06-17
Modlist/octocatless

Quezler:
Simplifies automatic updates to the modlist.md by not relying on the octocat overhead, it still functions identically. ^-^


NielsPilgaard (approved):
Simplifications are always nice 👌

PR #5020, opened 2022-06-18
Generate the modlist pretty much instantly

Quezler:
Thanks to https://github.com/ModdingX/ModListCreator/issues/11 there is now a bulk projects route available.

I have sent a pr for the java based modlist generator in order to benefit from it as well, but for now we have this.

It is my intention to see if i can further improve that generator so it can perhaps include file details directly in the bulk,
so perhaps one day we won't need this javascript file anymore and the ci can circle back around to moddingx's one. ^-^


NielsPilgaard (approved):
Very nice, good work!

PR #5053, opened 2022-07-06
[Expert] Alchemy Mastery Shard Update

MuteTiefling:
First, the shard itself is streamlined to more on theme items: 
image

This is an IE brewing setup, a Stimpack full of healing supplies, craftable totems of undying (ring of undying), and the pet token to keep them from dying. 

All very on point with the message of the quest: 
image

Mixer Kit:
image

Bottling Machine
image

The Stimpack. This is the real meat of the craft, I think.
image

It's a fairly sizeable increase in the number of potions, but mainly introduces two flasks and an incense stick. Automation of the botanical brewery for the incense stick would have been done long before, but the potion flasks may require some new automation to set up the Ars Nouveau Potion Melder.

Also introduces a few options for making those bandages with varying levels of efficiency:
image


NielsPilgaard (approved):
Looks good, solid theming 👍

PR #5057, opened 2022-07-06
Automatic prettier for kubejs client/server/startup scripts

Quezler:
Screen Shot 2022-07-06 at 14 21 29

Went ahead and played around with an automatic prettier workflow, applies `.prettierrc` to all the kubejs `.js` files:

Screen Shot 2022-07-06 at 14 16 00
Screen Shot 2022-07-06 at 14 16 09

All the people who last modified any of the files updated will get credited/referenced, which for the initial run is 11,
but any future runs (merging of pull requests) are more likely to only credit one person as a co-author.

Any changes to `.prettierrc` will not cause the bot to credit itself, but instead look for the first human.

This pull request does not include the files that will get updated once its merged, but they can be reviewed headlessly:
https://github.com/Quezler/Enigmatica6/commit/217716b325c1f6cd9982cda3d4b2963a17249356

bee larva


NielsPilgaard (commented):
I like the idea a lot, but the output you linked worries me: It looks like.. pretty much all scripts? And not all of them seem to have actual changes.

Which prettier config is used for this?

Quezler:
it uses the `.prettierrc` you've last modified years ago, and the files with seemingly no chances are likely just line endings 🤔

NielsPilgaard:
Hmm alright - Do you think you can change that? I'd prefer only actual changes were committed, and hopefully not have `EOL` changes to all files :P

Quezler:
> Do you think you can change that?

which line ending format would you like to enforce? https://prettier.io/docs/en/options.html#end-of-line

Quezler:
that should do it ^

NielsPilgaard:
Thanks 😊

PR #5059, opened 2022-07-06
Cleanup - Disable unused compact crafting, remove quests about them

MuteTiefling:
Cleanup - Disable unused compact crafting, remove quests about them

Could probably remove the mod entirely, but this'll do for now.

### DIFF
--- a/config/ftbquests/quests/chapters/storage.snbt
+++ b/config/ftbquests/quests/chapters/storage.snbt
@@ -595,7 +595,7 @@
 			}]
 		}
 		{
-			x: -4.5d
+			x: -3.5d
 			y: -11.0d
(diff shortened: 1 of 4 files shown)


Gcat101:
...why was it there in the first place?

Quezler:
because that mod in it's 1.12 days had its own crafting method, so quests were prepared for it (it split up into two mods), however afaik it was unreliable or not properly finished yet so e6 made its own recipes, and that just stuck around 🤔

MuteTiefling:
Yep. Quezler has the right of it. We had plans, bugs got in the way, so we went a different route.

NielsPilgaard (approved):

PR #5060, opened 2022-07-06
`lf` vs `crlf` line endings showdown

Quezler:
93 files get converted when the mode is `lf`, compared to 738 modified files if the mode was forced on `crlf`,
therefore it seems logical to just use the prettier default of `lf` in order to keep all kubejs files the same.

"i like identical formatting even more" & "prefer only actual changes were committed" kinda clash.

i'll just leave this pr + diff as a preview, so its not mixed in with style changes like the other one,
so the line endings can be reviewed separately, feel free to either merge or close, i don't mind.


NielsPilgaard:
I'd rather keep it as-is, I have a feeling this would be a headache down the line, not sure why :P thanks for the preview though.

Quezler:
probably the only headache is git blame being a bit harder to read, no idea what other problems it could cause if merged. 😋

PR #5062, opened 2022-07-07
Prettier trailing commas

Quezler:
another preview, this time for trailing commas:

the main selling point being that any git commits that add something to an array don't need to modify the lines above,
since those already contain a comma at their end, keeping the diffs/history/blame nice and clean, vs what happens now.

in the history is also a preview of "all" mode which adds them to functions, but that just looked weird & is somewhat new.

this is of course personal preference, but now that the code style is automatically enforced kubejs-wide it seemed logical to at least start the discussion on this topic, again feel free to either close or merge, though i of course hope for the latter. ^-^


NielsPilgaard:
I prefer no trailing commas, because I dislike how they look :P I like their purpose, but.. no thanks.

PR #5064, opened 2022-07-07
Added jei description for sulfur and sulfur dust

Gcat101:
A lot of newcomers think making sulfur dust requires mekanism and other late-game mods, when really it doesn't. I added a jei description so people would stop asking the same question over and over on discord.

### DIFF
--- a/kubejs/client_scripts/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/item_modifiers/jei_descriptions.js
@@ -507,6 +507,10 @@ onEvent('jei.information', (event) => {
         {
             items: ['naturesaura:birth_spirit'],
             text: [`Obtained by manually breeding animals in high Aura areas.`]
+        },
+        {
+            items: ['emendatusenigmatica:sulfur_gem', 'emendatusenigmatica:sulfur_dust'],
+            text: [`Sulfur Ore is NOT obtainable through mining in this modpack. Check out other recipes for the Sulfur Dust and find the one suitable for your progression stage.`]
         }
     ];
 
--- a/kubejs/client_scripts/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/item_modifiers/jei_descriptions.js
@@ -510,7 +510,7 @@ onEvent('jei.information', (event) => {
         },
         {
             items: ['emendatusenigmatica:sulfur_gem', 'emendatusenigmatica:sulfur_dust'],
-            text: [`Sulfur ore isn't obtainable in this modpack. Check out the recipe for the Farmer's Delight's cooking pot.`]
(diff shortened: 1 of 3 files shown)


kubejs/client_scripts/item_modifiers/jei_descriptions.js
```diff
@@ -507,6 +507,10 @@ onEvent('jei.information', (event) => {
         {
             items: ['naturesaura:birth_spirit'],
             text: [`Obtained by manually breeding animals in high Aura areas.`]
+        },
+        {
+            items: ['emendatusenigmatica:sulfur_gem', 'emendatusenigmatica:sulfur_dust'],
+            text: [`Sulfur ore isn't obtainable in this modpack. Check out the recipe for the Farmer's Delight's cooking pot.`]
```
NielsPilgaard, line 513: I'd like `Sulfur ore` to be `Sulfur Ore`

NielsPilgaard (changes requested):
Nice addition, thank you for your contribution :D

MuteTiefling (changes requested):
Sulfur Ore is obtainable, however, just not directly from mining. So that should be clarified. 

Also, the Farmer's Delight recipe is unique to Expert. So putting this as a JEI description for both modes is confusing. It should more generically point people to look at recipes for Sulfur Dust.

Quezler:
they likely meant the `jei_decriptions.js` in the expert subdirectory so it only loads in that mode, instead of mentioning in the text for both modes that its only for expert: https://github.com/EnigmaticaModpacks/Enigmatica6/blob/develop/kubejs/client_scripts/expert/item_modifiers/jei_descriptions.js

Gcat101:
yeah I'm just dumb

MuteTiefling:
Nah, I just meant making a more generic message that works for both. There's nothing drastic about the Expert recipes. For both modes, a simple reminder to look at the recipes for Sulfur Dust is sufficient.

MuteTiefling:
We've generally only used this type of description to explain things that are otherwise not discoverable in game, like the recent descriptions added to brewing. The only way to know how to craft those is pure trial and error, or checking an external wiki.

This isn't a case like that. The recipe is plainly visible in JEI. If they aren't seeing it, they're also going to miss this description. 

A quest would make more sense, but that feels very hand-holdy for an Expert pack.

Gcat101:
whoop commit duplication

Gcat101:
Closing this since it seems there’s no progress being made

NielsPilgaard:
We were expecting you to make the changes to be honest, since you said Ok Sure, that's how pull request development usually goes :D

Gcat101:
Yeah I know but there were no comments for 15 days so I closed it

Gcat101:
Should've just made a new branch. I'm a moron.

NielsPilgaard:
Fair enough 👍 

... I don't think you're a moron though 😅

PR #5093, opened 2022-07-17
Powah - blazing blood compat

StevenDoesStuffs:


NielsPilgaard (commented):
A value of 1 million seems a bit high compared to the 10.000 of lava, was this intentional? 🙂

Quezler:
> A value of 1 million seems a bit high compared to the 10.000 of lava, was this intentional? 🙂

it does match the power output value for the liquid in the magmatic dynamo:
https://github.com/EnigmaticaModpacks/Enigmatica6/blob/develop/kubejs/server_scripts/enigmatica/kubejs/base/recipetypes/thermal/dynamo/magmatic.js#L8

NielsPilgaard:
The value in the Powah config isn't FE though, it's `heat per 100 mb`

StevenDoesStuffs:
I didn't change the lava heat value, and the ratio of the heat between lava and blazing blood is the same as the ratio of RF generated between the two fluids in the magmatic dynamo.

When checking in game it says that lava gives 10k RF/100mb and blazing blood gives 1M RF/100mb, so it seems like the heat is just equal to RF.

NielsPilgaard (approved):

NielsPilgaard:
I see, thanks for clearing that up 👍

PR #5096, opened 2022-07-19
Drop bee larva via kubejs instead of loot tables

### LINKED ISSUE #4948: Have baby bees drop larva as well
### Describe your suggestion

https://github.com/EnigmaticaModpacks/Enigmatica6/blob/develop/kubejs/server_scripts/enigmatica/kubejs/base/loot_tables/resourcefulbees/bees.js

Currently only grown-up bees drop larva, leaving a window open for bees that meet their end before they can come of age.

### LINKED ISSUE #5095: Scavenger can dupe Larvae
### Modpack Version

v1.4,1

### Describe your issue.

pretty explanatory i think. any bee that can be killed can be duped with the scavenger enchantment. i do not know if this is intentional or not, but i just thought i'd mention this!

### Crash Report

_No response_

### Latest Log

_No response_

Yes

### User Modifications

Rubidium and Rubidium Extras. I don't think these mess with scavenger but one never knows.

Singleplayer

Quezler:
as discussed in ye old bee larva thread 🐝 

- fixes #5095
- fixes #4948

- the 4 commented console.logs are left there intentionally to make it easier to work there in the future
- the `{` and `}` ensure the `bee_map` constant cannot escape into other kubejs files
- done in a generic file so all (future) entity deaths can reuse the same handler

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/entity/on_death.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/entity/on_death.js
@@ -1,0 +1,30 @@
+{
+  const bee_map = {}
+  bees.forEach((bee) => {
+    bee_map[`resourcefulbees:${bee}_bee`] = `kubejs:${bee}_larva`;
+  });
+
+  onEvent('entity.death', event => {
+    // console.log(event); // dev.latvian.kubejs.entity.LivingEntityDeathEventJS@372ec6ba [dev.latvian.kubejs.entity.LivingEntityDeathEventJS]
+    // console.log(event.entity); // Clay Bee-9f197ad6-af90-4f2e-a58d-1fbec1ceb1fc
+    // console.log(event.entity.type); // resourcefulbees:clay_bee
+    // console.log(event.entity.name); // Clay Bee
+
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/enigmatica/kubejs/base/entity/on_death.js
```diff
@@ -0,0 +1,30 @@
+{
+  const bee_map = {}
```
NielsPilgaard, line 2: I'd prefer this const inside the `onEvent` call, that's how we usually do with non-global consts

NielsPilgaard (approved):

kubejs/server_scripts/enigmatica/kubejs/base/entity/on_death.js
Quezler: but then it would get redefined **each time** any mob dies, better to initialize the map outside the death event  👀

NielsPilgaard: Very valid point 👍

PR #5102, opened 2022-07-20
[Expert] Speed increases

MuteTiefling:
Dramatically increases the speed for the natural altar's ore processing step. Also increases aura cost per tick. 

Make the IE Mixer actually worth using. Makes it aproximately 8x faster than a maxed out (256 rotation) Create Mixer

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/expert/unification/unify_materials.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/expert/unification/unify_materials.js
@@ -282,8 +282,8 @@ onEvent('recipes', (event) => {
                 output: Ingredient.of(levigated_material).toJson(),
                 catalyst: Ingredient.of('naturesaura:crushing_catalyst').toJson(),
                 aura_type: 'naturesaura:overworld',
-                aura: 1500,
-                time: 10
+                aura: 300,
+                time: 1
             })
             .id(`enigmatica:expert/magical_ore_processing/aura/${material}`);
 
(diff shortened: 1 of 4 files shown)


kubejs/server_scripts/enigmatica/kubejs/expert/unification/unify_materials.js
```diff
@@ -282,8 +282,8 @@ onEvent('recipes', (event) => {
                 output: Ingredient.of(levigated_material).toJson(),
                 catalyst: Ingredient.of('naturesaura:crushing_catalyst').toJson(),
                 aura_type: 'naturesaura:overworld',
-                aura: 1500,
-                time: 10
+                aura: 300,
```
NielsPilgaard, line 285: I believe your intention was to increase this?

MuteTiefling: yep, it's doubled. was 150 per tick, now it's 300 per tick

MuteTiefling: er, wait where'd the time change go >.>

MuteTiefling: ok, no time is still 1, so 300 per tick :D

NielsPilgaard: The old value was 1500 per tick though, according to the diff above?

MuteTiefling: 1500 over ten. It's the total cost, not per tick cost

NielsPilgaard: Ahh okay, thanks 😅

PR #5135, opened 2022-08-02
Make the brass bee output brass dust from emendatus instead of crushed brass from create

Gcat101:
Title.

### DIFF
--- a/config/resourcefulbees/bees/alloy/Brass.json
+++ b/config/resourcefulbees/bees/alloy/Brass.json
@@ -39,7 +39,7 @@
     "CentrifugeData": {
         "hasCentrifugeOutput": true,
 
-        "mainOutput": "create:crushed_brass",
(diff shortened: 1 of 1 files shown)


MuteTiefling:
Why?

Gcat101:
because every other alloy bee outputs dusts from emendatus?

MuteTiefling:
@TemperedFool can correct me if I'm wrong, but I seem to recall that we did this intentionally to keep the processing of the bee outputs somewhat related to Create processing, since brass is from Create.

TemperedFool:
Yeah, when possible we stuck with products that required more processing. Tbh, dusts are too easy.

Gcat101:
Ah ok. Just seemed weird that it's the only bee that's like that.

NielsPilgaard:
Thanks but no thanks :P

PR #5141, opened 2022-08-03
Remove "This item is disabled" descriptions for docks from Dank Storage and Tank Null

Gcat101:
Title.

Since these items have recently been re-enabled, the jei description doesn't make sense.


MuteTiefling:
Ah, good catch, thanks.

PR #5143, opened 2022-08-03
[Expert] Fireproof Tree of Life

### LINKED ISSUE #5124: The Tree of Life occasionally breaks
### Modpack Version

1.4.1

### Describe your issue.

The Tree of Life occasionally gives me the message "failed to construct the structure" after having been functional for a while, with no action by me which could have reasonably had that result. Rebuilding it appears to replace some blocks on the ground near the base of the tree, possibly decoration. I suspect snowfall destroyed some of the decoration. Also, the pattern of replaced blocks appears to mirrors the dried kelp bocks - this is only an impression though. 

### Crash Report

_No response_

### Latest Log

_No response_

No

### User Modifications

_No response_

Singleplayer

### Discord Username

Ieldra@9172

MuteTiefling:
Should resolve #5124

Introduces 3 new fireproof replacement blocks to prevent external leaves/wood from burning. 
image
image

Existing Trees will continue to work, as the pattern is looking for new tags instead of the new blocks.
Building Gadget and Schematicanon schematics are updated to use these new blocks by default. 

This also changes the pathblocks inside the structure to accept `#rsgauges:soils` which is a pre-existing block tag comprised of the following:

- minecraft:dirt
- minecraft:coarse_dirt
- minecraft:podzol
- minecraft:farmland
- minecraft:grass_path

This means if anyone accidentally converts their paths to dirt, the structure won't fail.


MuteTiefling:
Item Fluid Transform script also re-written. Was going to use fluid transform for the recipes, but opted for lightning instead. Left the corrections in place as they follow current standards

PR #5154, opened 2022-08-06
Making Cloggrum Bars meltable

g0urra:


Kanzaji:
Looks good for me 👍

Kanzaji:
Okay actually after checking, it seems its actually a bit loosy. I know 48 is a bit better number but to make no lose conversion, one bar should give 54mb :D (48mb gives ~5,3 ingots from 16 bars, were spent were 6 ingots*)

NielsPilgaard (approved):

NielsPilgaard:
Thank you kindly :D

PR #5159, opened 2022-08-08
[Expert] Will crystal update

MuteTiefling:
With Blood Magic now having a net positive crystal growth mechanic, I think it's time to say goodbye to the PNC recipes. 

The basic will crystal growth recipe will remain: 
image

But the four flavors are removed and the new Will Catalysts are updated to the following:

image


MuteTiefling:
Demon Pylon:
image

Basic Routing Logic Upgrade:
image
Brings it earlier to match the rest of the routing node stuff.

Synthetic Upgrade Point
image

NielsPilgaard (approved):

PR #5162, opened 2022-08-11
[Expert] Add Recipe for Seared Bricks to Enrichment Chamber (#5142)

### LINKED ISSUE #5142: Melting grout through mekanism
### Describe your suggestion

Suggestion: add a recipe to [energized smelter] for melting grout into seared bricks. 

In order to avoid this recipe
image

Kanzaji:
Added grout -> Seared bricks recipe in Enrichment Chamber for expert. Resolves #5142

obraz


MuteTiefling (commented):
Looks good to me as is. I think we could be more generous with this recipe, however, given how late it is. Maybe 2x output instead?

Kanzaji:
Tbh I thought about that as well 😄 Just wasn't sure if you would like that idea

NielsPilgaard (approved):

PR #5170, opened 2022-08-16
Blacklist anvils from PNC and IF enchant extraction

### LINKED ISSUE #5167: Major Infinite Fortune exploit

MuteTiefling:
Partially resolves #5167. Leaves tome of scrapping as I don't see a way to blacklist items from that...


MuteTiefling:
Checking with Shadows to see if there's any way to deal with this.

MuteTiefling:
Looks like Shadows can fix this more cleanly from their end. They'll be releasing a fix that'll remove invalid enchants from anvils on placement.

PR #5174, opened 2022-08-16
Few changes to IF quest line

Kanzaji:
Again while working on E8 quests, it did bother me that generators descriptions are empty, and even tho you can craft all 4 tiers of machine frames, it still says you can only 3.
so:

- Changed first Frame quest to say there are 4 not 3 frames
- Added short descriptions to all generators used for Mycielial Reactor


MuteTiefling:
Looking good, thanks again :D

PR #5181, opened 2022-08-24
Fix a typo in Nature's Aura quest line.

Kanzaji:


config/ftbquests/quests/chapters/natures_aura.snbt
```diff
@@ -932,9 +932,9 @@
 			y: -7.0d
 			subtitle: "Abaddon's Gate"
 			description: [
-				"Ender Chests with a twist. Ender Crates can be assigned to a channel in an Anvil by simultaneously renaming them and combining them with an Eye of Ender."
+				"Ender Chests with a twist. Ender Crates can be assigned to a channel in an Anvil by simultaneously renaming them and combining with an Eye of Ender."
```
NielsPilgaard, line 935: This isn't actually a typo though, I don't think, the original sounds correct.

NielsPilgaard (changes requested):

config/ftbquests/quests/chapters/natures_aura.snbt
Kanzaji: To be honest, for me it does sound a bit weird. But it is your choice which one is better :D

NielsPilgaard (approved):

Issue #5261, opened 2022-11-15
Immersive Engineering: Gamestage not updated so multiblock cannot be formed


Lommelun:
### Modpack Version

1.6.1

### Describe your issue.

Gamestage prevents multiblock (tested with Fermenter and Squeezer) from forming.

Workaround:
Adding the medium machinery gamestage `/gamestages add ...` resolves the issue (allowing you to form the multiblock by whacking it with a hammer).

Possible relevant history:
When forming the Alloy Kiln (in version 1.6.0, about 2 weeks ago), my game crashed. Related: #4232 Could this impact the gamestage update?

### Latest Log

https://gist.github.com/Lommelun/ade65a645f6e5ff7aef98ac9ec78ba17

### Have you modified the modpack?

Yes

### User Modifications

Oculus
Rubidium
Magnesium Extras
Dynamic Lights Reforged
Abscent By Design

### Did the issue happen in singleplayer or on a server?

Singleplayer

### Discord Username

Lommelun#2653

theboo:
Likely this is caused by not right clicking the correct block in the multiblock structure with the hammer.. if you click the wrong block it will send the message about needing the proper schematics because IE checks through every single multiblock when you right click a block used in their multiblock structures with the hammer so when you right click the wrong block you end up getting that message talking about needing heavy schematics to form this multiblock even though you're attempting to form a medium schematic multiblock

In the IE book I believe it highlights the block you need to interact with to get the multiblock to form.

Lommelun:
I right clicked every block, and also rebuilt the structure three times. Both trying the copy-paste tool and the projector.

Also, it requires medium schematics not heavy. I only had the `starting_items` gamestage prior to adding `medium_machinery_schematics` manually.

MuteTiefling:
There have been a number of very random issues that have lead to these gamestages either being lost or never gained. I suspect your crash may have happened at just the wrong time leading to the data not being saved.

Next update, you'll be able to right click the various game stage items to gain the game stage again if it is lost for any reason.

JonasDoe:
One question though: Are gamestage something supposed to be in E6E? Tbh. I've no clue whether I need to reach another game stage first (and how, probably by finishing quest categories) or it's just a bug.

MuteTiefling:
Yes, the pack uses gamestages intentionally.

JonasDoe:
> Yes, the pack uses gamestages intentionally.

[...]

PR #5264, opened 2022-11-16
[Expert] Enable Right-Click on GameStage items to resolve accidental loss of GameStage

MuteTiefling:
Really should have thought twice about the names here, but it's too much work to fix them now, so more if statements!


MuteTiefling:
Resolves #5261

PR #5286, opened 2022-12-17
Fix a typo in Modded for Dummies book!

Kanzaji:
While playing with a friend, we found this small typo 😋
(I think it's a typo ^^", it doesn't sound natural anyways)


MuteTiefling:
good catch!

PR #5287, opened 2022-12-17
Fix Broken Language Entry for Tetra

Kanzaji:
Another hotfix after I found something broken 😋
obraz
obraz


Kanzaji:
PS: Not sure if this is a good way to fix lang files issues, but at least we know this is a thing right 😅

NielsPilgaard:
> PS: Not sure if this is a good way to fix lang files issues, but at least we know this is a thing right 😅

It is 👍 thanks!

NielsPilgaard (approved):

PR #5318, opened 2023-01-23
[Bountiful] Disable certain high-end rewards

Discomanco:
Since the entire thing was balanced with weights and unitWorths actually working, it became very powerful. It was sort of fine when it wasn't well known, but information spreads, and it might be time to disable some of the better rewards. All the disabled rewards has been put into a separate .txt file and can easily be put back in their respective files (or renamed back to .json in some cases)


NielsPilgaard (approved):
Thanks for the fix 👍

PR #5364, opened 2023-03-21
[Expert] More complete fix for scorching heat

MuteTiefling:
Since the previous fix wasn't as complete as I thought it was...

This removes ore smelting from the furnace entirely, moving it to the blast furnace and other modded furnaces. It also removes the metal requirements from crafting a blast furnace so as to make it possible to smelt early game still prior to any form of ore processing.


NielsPilgaard (approved):
Looks good to me, nice work 👍

PR #5391, opened 2023-05-16
Expert - Final Goal

MuteTiefling:
The Creative Worldshaper never sat right with me. I think it's time for it to go. 

The real end of the pack will be obtaining the tree of life trophy inspired by the recipe for the tree of life controller.
image

Ultimately, this doesn't change much. It just gets rid of the worldshaper recipes and introduces a new one for the trophy itself.


ZZZank:
I personally don't think this to be a necessary change, because the trophy recipe is a little duplicated compared with the Tree Of Life Controller recipes, while the Worldshaper recipe is more ‘varied'. And the player can played around with the Worldshaper after getting it, but a trophy cannot be do so, losing some fun. 

And I'm curious about what `sat right with me` means, can you explained it in more detail?

MuteTiefling:
> And I'm curious about what `sat right with me` means, can you explained it in more detail?

Creative items make bad 'rewards'.

NielsPilgaard:
To be honest I like the Worldshaper as the end goal. I think creative items make for bad rewards in most cases, but I think it's on-theme and appropriate here :)

Issue #5400, opened 2023-06-05
Add Refined Crafter Proxy


StevenDoesStuffs:
### URL

https://www.curseforge.com/minecraft/mc-mods/refined-crafter-proxy

### Why would you like the mod added?

Disclaimer: I am the author of this mod.

This mod makes it easier to distribute the same recipes over multiple machines by adding a "crafter proxy" which mirrors another crafter's patterns. RS will then automatically load balance the recipes across all the machines without forcing the player to create multiple of the same pattern for multiple crafters. It's a very small mod purely for QOL, and its only dependency is RS. 

I've been playing with this in my E6E world for a week or so now and have fixed all the bugs I could find in that time, and everything seems to be working now.

NielsPilgaard:
Ooh, nice mod idea! I vote yes 😄

StevenDoesStuffs:
Ahhh shoot I forgot that kotlinforforge is also a dependency, but y'all already have that.

MuteTiefling:
No issue here. Just curious what the power draw is and if it is configurable.

Darkere:
Appears to be 4x the regular crafter powerdraw. Which in our case... would be an issue.

NielsPilgaard:
Is 4x energy cost worth it in our case?

StevenDoesStuffs:
Ah yeah that was more of a placeholder value. I can make it configurable.

theboo:
My only other question would be does it work with the upgraded crafters we have in e6e? Cuz upgrading for energy efficiency in e6e is a big deal lol

StevenDoesStuffs:
Yeah I was really struggling to figure out how do deal with the extrastorage crafters. 

On one hand, I'm thinking I could just make it so that the proxy takes the properties (speed + energy draw) of whatever crafter you craft it with (meaning you could craft it with any of the upgraded crafters as well), but that's kinda difficult.

I'd have to either pull in extrastorage as a dependency or try to generically wrap a crafter node and turn it into a proxy (which might not even be possible, since I'm not sure forge would let me go from an item to a RS node that easily). 

The other idea is to store properties as part of the NBT tag, meaning they could be customized as a part of the crafting recipe, and then modpack creators like y'all could create the recipes for the upgraded crafters on your side. 

I'm thinking of going for the second one, since it's much easier on my end. Thoughts?

MuteTiefling:
[...]

PR #5403, opened 2023-06-12
Add refinedcrafterproxy

StevenDoesStuffs:
#5400 

Please let me know if I've missed anything (I'm not sure if I have to actually add the mod anywhere or if that's all on your side).

### DIFF
--- a/config/configswapper/expert/config/refinedcrafterproxy-common.toml
+++ b/config/configswapper/expert/config/refinedcrafterproxy-common.toml
@@ -1,0 +1,62 @@
+
+#Note: for technical reasons, if anything in the tiers section is invalid, it will reset the entire section.
+#Luckily, forge backs up invalid configs in .toml.bak files, so look there if everything got reset.
+[tiers]
+
+	#Special tier for crafters without a tier tag.
+	#For values other than `default`, this is the value that appears within the tier tag.
+	[tiers.default]
+		#The display name of the tier.
+		#If nonempty, the name of the block will be formatted as `{name} Crafter Proxy`.
+		displayName = ""
+		#The energy used for the crafter.
+		#Must be an integer >= 0.
+		crafterEnergyUsage = 64
+		#The energy used for every pattern in the crafter.
+		#Must be an integer >= 0.
+		patternsEnergyUsage = 16
+		#The multiplier for the energy usage of the upgrades in the crafter.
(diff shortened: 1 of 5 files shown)


config/configswapper/expert/config/refinedcrafterproxy-common.toml
```diff
@@ -0,0 +1,62 @@
...
+		#Must be a list of integers of length 5 with elements > 0.
+		updateIntervals = [10, 8, 6, 4, 2]
+		#The maximum number of successful crafting updates per crafter update,
+		#where the (zero-indexed) index in the list is the number of speed upgrades.
+		#Must be a list of integers of length 5 with elements > 0.
+		maximumSuccessfulCraftingUpdates = [1, 2, 3, 4, 5]
+
+	[tiers.extrastorage_iron]
+		displayName = "Iron"
+		crafterEnergyUsage = 15
```
NielsPilgaard, line 34: I'm thinking this should be at least 64, I'm unsure how much the Iron Crafter actually uses though. Same goes for the other tiers, at least 64, unsure how much 😄

NielsPilgaard (changes requested):
This is such a boss move :P Very nice work, you've followed our coding guidelines to a T!

There's just that one comment I'd like addressed, and then I'll happily merge this. There's no release planned for E6 in the immediate future, but I'm sure we'll make one eventually.

config/configswapper/expert/config/refinedcrafterproxy-common.toml
StevenDoesStuffs: The upgraded crafters use `15 * tier` for their base energy (for 1.16.5 this is actually hardcoded into extrastorage, see [here](https://github.com/Edivad99/ExtraStorage/blob/a628fb87c8084a118c9cc3b064bd6d836c32a184/src/main/java/edivad/extrastorage/nodes/AdvancedCrafterNetworkNode.java#LL165C54-L165C54)). Currently all the settings match their respective crafters. Do you still want me to change this?

NielsPilgaard: Ahh okay, I wasn't aware - Then this is good to go 👌

NielsPilgaard (approved):
Thank you for the mod and integration :D

StevenDoesStuffs:
Done :smile:

PR #5423, opened 2023-11-13
makes tooltips/descriptions localizable

ZZZank:
This PR makes tooltips/descriptions defined in scripts localizable, no longer hard-coded. 
It includes:
- expert tooltips
- expert JEI descriptions
- base JEI descriptions

This PR does not thoroughly makes every texts localizable, because moving and renaming keys is generally tedious but time consuming. So if there be some form of continuation of this PR, it will be in another seperate PR. Sorry for incompleteness. 

I've test this change on my own instances, it works pretty well. 
In fact it looks exactly the same as before, while also enabling localization:
屏幕截图 2023-11-13 000644
屏幕截图 2023-11-13 000637


kubejs/client_scripts/item_modifiers/tooltips.js
Darkere, line 1: Does this still work like this? Looks like you forgot to replace the text.

ZZZank: Yes, it still works, translate() function can handle raw text. I did not move text here just because it's hard to move text when wrapped by various function calls.

NielsPilgaard (approved):
Phew this must've been a ton of work, thank you for your cotribution 😊
I'd like MuteTiefling to have a quick look at this as well, if they're okay with it we'll merge it in 👍

MuteTiefling:
Looks good to me! Nicely done <3

MuteTiefling (approved):

PR #5426, opened 2023-11-17
Complete infrastructures for localizing tooltips/descriptions

ZZZank:
This PR makes every tooltips/descritions defined in scripts localizable. It's a continuation of #5423 . 
## Summary
- base::tooltips
- normal::tooltips
- normal::jei_descriptions

and
- move dynamic added text to the front of text adding action, to enable localization
- build infrastructures to allow handling texts with `%s` placeholder. It's first used in Simple Farming Beverages descriptions

## New Infrastructure `with`
The new infrastructures added a `with` , which is in the same level of `items` and `text` , to allow handling `%s` in `text` . `with` should be a nested array. An real world example may look like this:
```
{
  items: ['minecraft:stick'],
  text: ['example1: %s', '%s-example2-%s'],
  with: [
     ['insert1'],
     ['insert2', 'insert3'],
  ]
}
```
This will be interpreted as: 
```
example1: insert1
insert2-example2-insert3
```

## Test
I don't have time to do a throughout test recently, but I do made a quick test, and no problem found.


.gitignore
```diff
@@ -224,3 +224,8 @@ config/configswapper.json
 config/betterendforge/client.json
 config/Mekanism/tiers.toml
 pregen/supporters.json
+schematics/uploaded/
+modernfix/structureCacheV1/
+lightspeed-cache/1.16.5/
+config/appliedenergistics2-client.toml
+config/rubidium-options.json
```
NielsPilgaard, line 231: Please remove these additions to the .gitignore :)

kubejs/assets/enigmatica/lang/en_us.json
```diff
@@ -236,7 +239,64 @@
...
+    "descriptions.enigmatica.base.beverages.cider.effect": "Luck",
+    "descriptions.enigmatica.base.beverages.mead.ingredient": "Honeycomb",
+    "descriptions.enigmatica.base.beverages.mead.effect": "Haste",
+    "descriptions.enigmatica.base.beverages.sake.ingredient": "Rice",
+    "descriptions.enigmatica.base.beverages.sake.effect": "Speed",
+    "descriptions.enigmatica.base.beverages.tiswin.ingredient": "Cactus Fruit",
+    "descriptions.enigmatica.base.beverages.tiswin.effect": "a Health Boost",
+    "descriptions.enigmatica.base.beverages.vodka.ingredient": "Potatoes",
+    "descriptions.enigmatica.base.beverages.vodka.effect": "Resistance",
+    "descriptions.enigmatica.base.beverages.wiskey.ingredient": "Wheat",
```
NielsPilgaard, line 262:
```suggestion
    "descriptions.enigmatica.base.beverages.whiskey.ingredient": "Wheat",
```

kubejs/assets/enigmatica/lang/en_us.json
```diff
@@ -236,7 +239,64 @@
...
+    "descriptions.enigmatica.base.beverages.wiskey.effect": "Absorption",
```
NielsPilgaard, line 263:
```suggestion
    "descriptions.enigmatica.base.beverages.whiskey.effect": "Absorption",
```

kubejs/client_scripts/item_modifiers/jei_descriptions.js
```diff
@@ -429,82 +429,61 @@ onEvent('jei.information', (event) => {
...
 
     recipes.forEach((recipe) => {
         for (let i = 0; i < recipe.text.length; i++) {
-            recipe.text[i] = Text.translate(recipe.text[i]);
+            if (recipe.with[i]) {
+                recipe.with[i] = recipe.with[i].map((str) => Text.translate(str));
+                recipe.text[i] = Text.translate(recipe.text[i], ...recipe.with[i]);
+            } else {
+                recipe.text[i] = Text.translate(recipe.text[i]);
+            }
```
NielsPilgaard, line 483: Could you please explain what's going on here? :D

NielsPilgaard (commented):
Nice work again!

.gitignore
ZZZank: Oh, I forgot this. Fixed in new commits

kubejs/client_scripts/item_modifiers/jei_descriptions.js
ZZZank: Sure. Line479 and line480 are two major additions, so I will explain more on these two. 
It will first try to determine if there is a `recipe.with[i]` , which should provides texts for `recipe.text[i]` to replace %s . If not, it will simply use old handlers(line482) .
If there is a `recipe.with[i]` , it will first try to localize texts inside `recipe.with[i]` using `.map((str) => Text.translate(str))` , raw texts will be retained as-is. 
Then those texts will be "unpacked" and passed to translate() function. The `...` in `...recipe.with[i]` is used for unpacking elements. Let's take an example:
Assume an array `arr = ["str1", 2]` and a function `func()` . Calling `func(arr)` equals to `func(["str1", 2])` , while `func(...arr)` equals to `func("str1", 2)` , note that there is no "[" and "]" when using "..." .

NielsPilgaard: I see, thank you!
What happens if %s is first in the string? Does the ordering of parameters in `Text.translate(recipe.text[i], ...recipe.with[i]);` matter?

Cause it seems to me like this would produce original string + any occurrences of %s, afterwards?

ZZZank: Yes, the ordering of parameters does matter, because translate() function works like this: 
```
function translate(text, ...args){
    t = langFile[text] ? langFile[text] : text.copy()
    args.foreach(arg => {
        t.replaceOnce("%s", arg)
    })
    return t
}
```
(These codes are simplified, not actual source code.)
It will takes the first parameter it received as the base, and other parameters are used for replacing %s . 
That's why recipe.text[i] should be the first parameter, we wants it to be the base. 
And, about `...args` , here it means every parameters that are not the first param, will be packed into an array called "args" . So when passing `…recipe.with[i]` , we needs to unpack first, to avoid nested array which cannot be handled by `t.replaceOnce("%s", arg)` .

NielsPilgaard: Ahh okay, thanks for bearing with me and explaining 😃

NielsPilgaard (approved):

PR #5427, opened 2023-11-20
Add crafting recipe for Hanging Vines from Valhelsia Structures

Kanzaji:
Found this really nice vine, however currently it's somewhat rare to find and is not craftable, so decided to make it craftable 😅 

The vine in question is the one from Valhelsia structures, that can be placed on the ceilings.
obraz

Crafting recipe:
obraz
Additional conversion recipe at a loss:
obraz


MuteTiefling:
There's already stonecutting for other vine conversions. Maybe it'd be better to stick this there?

image

Kanzaji:
> There's already stonecutting for other vine conversions. Maybe it'd be better to stick this there?
Good point 😄

NielsPilgaard (approved):

PR #5428, opened 2023-11-25
cutting recipe for turning Root into Root(item)

ZZZank:
Ported from [here](https://github.com/ZZZank/Enlightened6/commit/de9667dba1cee46a83ac30eb614abc4771e5e1b2)
In expert mode, Redstone Root (from Botania) requires Root item (Quark) to craft, which should not be a problem because Root (block) is craftable. 
But players (like me) sometimes didn't realize this, and will takes redundant efforts in caves just to get a Root (item).

So this PR adds a simple recipe, to tell players that you can turn Root into Root(item)
image


NielsPilgaard (approved):

NielsPilgaard:
Nice addition, thank you!
@MuteTiefling would you mind also giving this a once over? 😊

MuteTiefling (approved):
Looks good to me

PR #5429, opened 2023-11-28
use length check to handle `with` in JEI Description

ZZZank:
JavaScript's variable accessing is trickier than my expection. Errors will be reported by KubeJS at seemly completely random time. 
So, instead of painstakingly debugging(I've done so, really painstaking), let's just use `try-catch` .

Ported from https://github.com/ZZZank/Enlightened6/commit/21170382c9f805a5359eb4f3d4718ac0a5aa98a3

### DIFF
--- a/kubejs/client_scripts/item_modifiers/jei_descriptions.js
+++ b/kubejs/client_scripts/item_modifiers/jei_descriptions.js
@@ -491,9 +491,9 @@ onEvent('jei.information', (event) => {
 
     recipes.forEach((recipe) => {
         for (let i = 0; i < recipe.text.length; i++) {
-            try {
+            if (recipe.with && recipe.with.length > i) {
                 recipe.text[i] = Text.translate(recipe.text[i], recipe.with[i]);
-            } catch (e) {
+            } else {
                 recipe.text[i] = Text.translate(recipe.text[i]);
             }
         }


NielsPilgaard:
Hmm, what seems to throw an exception, do you have any examples?

ZZZank:
The error happens within `text.with[i]` .
When the `with` is not defined, accessing it will gets an `undefined`, so `text.with[i]` becomes accessing the i-th element of undefined, which will causes an Exception to be thrown. 
Originally I tried `text.with && text.with[i]` to fix this. But on my side the problem is not fixed by doing this, at least not completely. So finally I switched to `try-catch` .

NielsPilgaard:
What about `text?.with[i]`? that'll return undefined instead of throwing, if with[i] isn't there

NielsPilgaard:
Aw damn. You could do a check on the length of text.with to see if the entry i is there.

kubejs/client_scripts/item_modifiers/jei_descriptions.js
```diff
@@ -491,9 +491,9 @@ onEvent('jei.information', (event) => {
 
     recipes.forEach((recipe) => {
         for (let i = 0; i < recipe.text.length; i++) {
-            if (recipe.with[i]) {
+            try {
```
Kanzaji, line 494: How about doing something like this:
```js
if (Object.keys(recipe).includes("with")) {
    recipe.text[i] = Text.translate(recipe.text[i], recipe.with[i]);
} else {
    recipe.text[i] = Text.translate(recipe.text[i]);
}
```

NielsPilgaard: The main issue is that the recipe.text and recipe.with arrays can be different lengths, checking whether recipe.with exists is simply `if (recipe.with){...}`

NielsPilgaard: exactly 👍

NielsPilgaard, line 494:
```suggestion
            if (recipe.with && recipe.with.length > i) {
```

kubejs/client_scripts/item_modifiers/jei_descriptions.js
```diff
@@ -491,9 +491,9 @@ onEvent('jei.information', (event) => {
...
                 recipe.text[i] = Text.translate(recipe.text[i], recipe.with[i]);
-            } else {
+            } catch (e) {
```
NielsPilgaard, line 496:
```suggestion
            } else {
```

NielsPilgaard (changes requested):

ZZZank:
`recipe.with && recipe.with.length > i` does works, switched! 

But I'm still confused. On theory, `text.with && text.with[i]` should also be effective, but in fact it will causes errors on my side. Strange.

NielsPilgaard (approved):

PR #5433, opened 2023-12-07
port vanilla textures for ingot/nugget/block in E9E

ZZZank:
Hopefully this won't cause me any copyright issuess.

This PR ports textures for vanilla ingots/nuggets/blocks in E9 back to E6, to makes them looks more consistent when placed together with EE items.


NielsPilgaard:
I'd rather leave the textures as is to be honest, to not mess too much with existing worlds/builds.

ZZZank:
Got it!

PR #5435, opened 2023-12-30
Fluid registry rework

ZZZank:
Cherry-picked from https://github.com/ZZZank/Enlightened6/commit/9e9b9a977c30acd2fb83fd57cdf8fbdad0133195

There's nothing destructive in this PR, just refactor scripts for fluid registry. 

The arg `display` is now deprecated, in favor of using lang file(en_us.json) , so that localization can be easier. 
But, if someone uses `display` , it will still properly set the displayName of targeted fluid as before, but a warning will be printed into the log. 
There's also some tips to help devs move from using `display` to using `en_us.json` .


NielsPilgaard (approved):
LGTM, thank you 👍

PR #5444, opened 2024-03-01
Function cleanup

ZZZank:
This PR:
- Adds type comments for most of functions in `functions.js`. These comments are basically copied from [my `functions.js`](https://github.com/ZZZank/Enlightened6/blob/enlightened_master/kubejs/server_scripts/enigmatica/kubejs/functions.js)
- Rewrite about half of all functions in `functions.js`, to make them clearer and possibly more performant. E.g. `getPreferredItemInTag` is rewritten to use `maxOf` instead of `sort`, this can reduce time complexity from O(n*lg(n)) to O(n), so should be faster
- fixed a [horribly wrong format](https://github.com/EnigmaticaModpacks/Enigmatica6/commit/ffda42490e8a6ac0b904a21b4148f8008d11c336) in unify_sawables::create_cutting(), which will prevent recipes from actually getting added
- move functions for recipe-gen from global scope to inside event callback, to prevent access
ing by accident. Also removing an unused param `event`. These functions are located mostly at `server_script/xxx/unification/`

These changes have been applied to my fork of E6/E6E for a long time, and they seems working pretty well

### DIFF
--- a/kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_dyes.js
+++ b/kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_dyes.js
@@ -3,286 +3,286 @@ onEvent('recipes', (event) => {
     const id_prefix = 'enigmatica:base/unification/unify_dyes/';
 
     dyeSources.forEach((recipe) => {
-        botania_dye_pestle_mortar(event, recipe, id_prefix);
-        create_dye_milling(event, recipe, id_prefix);
-        ars_nouveau_dye_crushing(event, recipe, id_prefix);
-        immersiveengineering_dye_crusher(event, recipe, id_prefix);
-        //integrateddynamics_dye_squeezing(event, recipe, id_prefix);
-        mekanism_dye_enriching(event, recipe, id_prefix);
-        mekanism_pigment_extracting(event, recipe, id_prefix);
-        pedestals_dye_crushing(event, recipe, id_prefix);
-        thermal_dye_centrifuge(event, recipe, id_prefix);
-        atum_quern_milling(event, recipe, id_prefix);
-        shapeless_dye_crafting(event, recipe, id_prefix);
-        occultism_dye_crushing(event, recipe, id_prefix);
+        botania_dye_pestle_mortar(recipe, id_prefix);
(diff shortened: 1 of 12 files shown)


NielsPilgaard:
Thanks for the PR, I'll go over it once I find some time 😅

kubejs/server_scripts/enigmatica/kubejs/base/unification/unify_dyes.js
```diff
@@ -3,286 +3,286 @@ onEvent('recipes', (event) => {
...
-        //integrateddynamics_dye_squeezing(event, recipe, id_prefix);
-        mekanism_dye_enriching(event, recipe, id_prefix);
-        mekanism_pigment_extracting(event, recipe, id_prefix);
-        pedestals_dye_crushing(event, recipe, id_prefix);
-        thermal_dye_centrifuge(event, recipe, id_prefix);
-        atum_quern_milling(event, recipe, id_prefix);
-        shapeless_dye_crafting(event, recipe, id_prefix);
-        occultism_dye_crushing(event, recipe, id_prefix);
+        botania_dye_pestle_mortar(recipe, id_prefix);
+        create_dye_milling(recipe, id_prefix);
```
NielsPilgaard, line 7: Sorry for the suuuper late reply to this 😬 
Why remove event? I like that it's an explicit requirement for the function to work, rather than an implicitly defined variable.

NielsPilgaard: Sigh ignore me, I hadn't read the PR description properly. I get it now.

NielsPilgaard:
Do you have any idea whether this is still safe to merge? 😬

ZZZank:
I would say surely yes, but such large amount of line changes did make it a bit hard to keep track of codes. I think I will soon make another PR that only focuses on fixing `unify_sawables.js` and expanding `functions.js`

NielsPilgaard:
> I would say surely yes, but such large amount of line changes did make it a bit hard to keep track of codes. I think I will soon make another PR that only focuses on fixing `unify_sawables.js` and expanding `functions.js`

I would really appreciate that 😃

ZZZank:
replaced by #5483

PR #5483, opened 2025-02-18
Functions doc

ZZZank:
This is a "distilled" version of #5444 , basically guaranteed to be safe, because changes here will only be one of:
1. function body is not changed, only JSDoc added
2. function body changed, but logic is the same
3. function moved from one place to another
4. function rewritten, but tested by myself ingame

Note that most JSDoc needs ProbeJS Legacy to actually work


NielsPilgaard (approved):
Thanks! 😊

PR #5484, opened 2025-02-18
fix create sawables

ZZZank:
changes extracted from #5444


NielsPilgaard (approved):
Big improvement, thanks 👍

PR #5485, opened 2025-02-19
remove `const` in for loop

ZZZank:
There was a bug in Lat's Rhino (KubeJS's dependency) that `const` in any kind of for loop will break codes at runtime. It's partially resolved in 1.18+, but not 1.16, so we can't use `const` here :(


NielsPilgaard (approved):

NielsPilgaard:
Ah alright, thanks for the fix 👍 const doesn't really make sense in a loop like this anyway.

PR #5488, opened 2025-03-02
Restore CI

ZZZank:
The difference between this new CI and old deleted CI:
- CI can only be ran manually, PR or pushing commits will not trigger CI
- KubeJS compiling in CI is now using a special fork of KubeJS:
```
      # using a special fork of KubeJS with several debug features enabled by default
      - run: git clone --branch enigmatica https://github.com/ZZZank/KubeJS.git
```
It should fix the problem that patch applying will fail
- updated actions to the latest available (usually v4) to fix this:
```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```


NielsPilgaard (approved):
Sure, why not, I don't think I'll be using it though 😅 thank you nonetheless!

